PR #4, opened 2022-07-19
CADL of DPG as Resiliency tests

lmazuel:


packages/cadl-ranch-specs/http/resiliency/customization.ts
```diff
@@ -0,0 +1,65 @@
+import { app, json } from "../../api";
```
lmazuel, line 1: @timotheeguerin probably this won't work now

packages/cadl-ranch-specs/http/resiliency/cadl-project.yaml
```diff
@@ -0,0 +1,2 @@
+emitters:
```
lmazuel, line 1: Should I commit this file?

packages/cadl-ranch-specs/http/resiliency/customization.ts
```diff
@@ -0,0 +1,65 @@
...
+
+app.category("dpg", () => {
```
lmazuel, line 3: I will change the category to "azure"

packages/cadl-ranch-specs/http/resiliency/cadl-project.yaml
timotheeguerin: I think we should only need one `cadl-project.yaml` at the root if we want to configure the default emitter. SHould it be `cadl-autorest` though for now at least?

packages/cadl-ranch-specs/http/resiliency/package.json
```diff
@@ -0,0 +1,17 @@
+{
```
timotheeguerin, line 1: shouldn't need that file, the one at the root of `cadl-ranch-specs` should cover all the scenarios

timotheeguerin (approved):
Looks good, I think there is some inconsitentcy with the scenario name in the mock apis but might need to resolve that when the generated type is here

PR #28, opened 2022-07-29
Initial test server tests for models with Cadl readonly properties 

annelo-msft:


timotheeguerin:
@annelo-msft  there is code to start adding the mockapis and should be validation to complain if it isn't added but I excluded the model folder for now as there is the existing spec missing the mock apis but think should add them now with new specs
Laurent added a few here as an example https://github.com/Azure/cadl-ranch/pull/4

annelo-msft:
Hi @timotheeguerin, thanks!  Can you tell me more about the validation, and how I can test my implementations to make sure they're correct?

PR #42, opened 2022-08-11
Add extensible enum parameters, both optional and required

AlexanderSher:


packages/cadl-ranch-specs/http/clients/interfaces/main.cadl
```diff
@@ -24,18 +24,29 @@ model Cat {
   name : string;
 }
 
+enum DogBreedValues {
```
annelo-msft, line 27: Would it be possible to add this in a separate Cadl file?  This one is specifically for interface scenarios: https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http#clients

I would put it under models/enums, and update the README, or perhaps you could use the one that Java recently added? https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/from-testserver/non-string-enum/main.cadl

PR #49, opened 2022-08-15
Remove message property in message model

pshao25:
message property is not valid in .net model

### DIFF
--- a/packages/cadl-ranch-specs/http/resiliency/srv-driven-1/main.cadl
+++ b/packages/cadl-ranch-specs/http/resiliency/srv-driven-1/main.cadl
@@ -15,7 +15,6 @@ model PostInput {
 }
 
 model Message {
-  message: string;
 }
 
 @doc("Answer from service")
--- a/packages/cadl-ranch-specs/http/resiliency/srv-driven-1/main.cadl
+++ b/packages/cadl-ranch-specs/http/resiliency/srv-driven-1/main.cadl
@@ -14,13 +14,14 @@ model PostInput {
   url: string;
 }
 
-model Message {
+model MessageBody {
+  message: string;
 }
 
 @doc("Answer from service")
 model MessageResponse {
   @statusCode statusCode: 200;
-  @body body: Message;
+  @body body: MessageBody;
 }
 
 @route("/serviceDriven")
--- a/packages/cadl-ranch-specs/http/resiliency/srv-driven-2/main.cadl
+++ b/packages/cadl-ranch-specs/http/resiliency/srv-driven-2/main.cadl
@@ -23,13 +23,14 @@ model PostInput {
   url: string;
 }
 
-model Message {
+model MessageBody {
+  message: string;
 }
 
 @doc("Answer from service")
 model MessageResponse {
   @statusCode statusCode: 200;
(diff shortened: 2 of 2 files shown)


timotheeguerin:
what does this mean `message` property is not valid in `.net` model? You cannot use a property named `message` in the csharp language? That seems like something that should be handled by the generator with an alternate name. (This could potentially be handled at the cadl level with sidecar files)

pshao25:
> what does this mean `message` property is not valid in `.net` model? You cannot use a property named `message` in the csharp language? That seems like something that should be handled by the generator with an alternate name. (This could potentially be handled at the cadl level with sidecar files)

If we generate model like
```C#
public partial class Message
{
    public string Message { get; set; }
}
```
It will throw exception as 

> error CS0542: 'Message': member names cannot be the same as their enclosing type

Yes, we could pick an alternative name. However, the parity swagger [dpg-initial.json](https://github.com/Azure/autorest.testserver/blob/da9260e442bf3736cac5017bf34af6b34f9ceeb5/swagger/dpg-initial.json#L22) has response body object (which means it does not have message property).

+ @lmazuel @annelo-msft

lmazuel:
I agree with @timotheeguerin as well, we can't just a constraint on CADL writers that says "don't name a property and a class the same", this has to be handled at the generator level.

lmazuel (changes requested):
I disagree with the intent of the PR. At the worst case, we may discuss renaming it (not removing), as removing makes no sense as this would not fit the mockapi behavior anymore.
This PR would also need anyway to change the mockapi to adapt to the behavior.
Sorry but this PR is wrong in many aspect.

pshao25:
> I disagree with the intent of the PR. At the worst case, we may discuss renaming it (not removing), as removing makes no sense as this would not fit the mockapi behavior anymore. This PR would also need anyway to change the mockapi to adapt to the behavior. Sorry but this PR is wrong in many aspect.

Just curious how we handle this in swagger. So we allow this behavior in swagger but we don't have real cases so that we didn't find this before?

But I hope to update the name to unblock August MVP.

@annelo-msft I create this to track: https://github.com/Azure/autorest.csharp/issues/2626

iscai-msft:
@pshao25 @annelo-msft  I would recommend adding a step where you "pad" these names that are reserved keywords in C#. For example, if you see a property named `message`, and you know this is invalid in your C# generator, then you can "pad" it to be something like `MessageProperty`.

This is something Python has been doing since the beginning (so we haven't had to worry about breaking changes), and it can be done in a way that introduces granularity (for example, maybe `Message` is a valid parameter name, but not a valid property name). You can look through our [preprocessing](https://github.com/Azure/autorest.python/blob/autorestv3/packages/autorest.python/autorest/preprocess/python_mappings.py#L54) step if you're interested in the Python-specific implementation

lmazuel:
@pshao25 I don't know if the information went through, but I was discussing with @annelo-msft that I was fine to rename the class MessageClass or something like that, as the intent of the test is not to test C# ability to solve name conflict (but their should be another test about that).

Closing this PR in the meantime, if you want to rename the class, please tell me the new PR. Thanks!

PR #52, opened 2022-08-17
Add mockapi for basic input

haolingdong-msft:


packages/cadl-ranch-specs/http/models/input-basic/main.cadl
```diff
@@ -24,6 +24,6 @@ model InputModel {
 
 @scenario
 @scenarioDoc("Generate and send an input-only model with required reference and value type properties.")
-@route("/models")
+@route("/models/basic")
 @get
 op getModel(@body input: InputModel): Response<200> | ErrorResponse ;
```
weidongxu-microsoft, line 29: is `Response<200>` same as `OkResponse`?

packages/cadl-ranch-specs/http/models/input-basic/mockapi.ts
```diff
@@ -0,0 +1,19 @@
+import { passOnSuccess, ScenarioMockApi, mockapi, json } from "@azure-tools/cadl-ranch-api";
+
+export const Scenarios: Record<string, ScenarioMockApi> = {};
+
+Scenarios.Basic_Input = passOnSuccess(
+    mockapi.get("/models/basic", (req) => {
+      if (req.body && req.body["requiredString"] && req.body["requiredInt"]) {
+          return {
+            status: 200,
+            body: json({ message: `InputModel was successfully returned` }),
```
weidongxu-microsoft, line 10: Where is the cadl model for this `{ message: `InputModel was successfully returned` }`?

haolingdong-msft: I will modified as 'InputModel was successfully received'.

weidongxu-microsoft: What I mean is that if you have `Response<200>` in Cadl without a body, you should not return the body in mock api?

packages/cadl-ranch-specs/http/models/input-basic/main.cadl
haolingdong-msft: Yes, https://github.com/microsoft/cadl/blob/main/packages/rest/lib/http.cadl#L35

packages/cadl-ranch-specs/http/models/input-basic/mockapi.ts
haolingdong-msft: oh, yes, previously I thought adding body with message can help to validate the response, but validating on status code should be good enough. We should make mockapi.ts consistent with main.cadl file. It is better to remove the the returned body.

weidongxu-microsoft: Yeah. If you want to add the message, modify the Cadl file to return proper model.

Same for the error case, follow the model in core.

packages/cadl-ranch-specs/http/models/input-basic/mockapi.ts
```diff
@@ -0,0 +1,18 @@
...
+Scenarios.Basic_Input = passOnSuccess(
+    mockapi.get("/models/basic", (req) => {
+      if (req.body && req.body["requiredString"] && req.body["requiredInt"]) {
+          return {
+            status: 200
+          };
+        } else {
+          return {
+            status: 400,
+            body: json({ code: 400, message: `Expected InputModel parameter with required fields` }),
```
weidongxu-microsoft, line 14: this is still not correct! read the file, it should have `error` first.

you might need open a new PR.

Issue #39, opened 2022-08-10
Mock api implementation [Models]


timotheeguerin:
Creating an issue for the missing implementations now that the CI enforce it 

Update here when added the test https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch/src/actions/validate-mock-apis.ts#L90

[Docs for writing mock apis](https://github.com/Azure/cadl-ranch/blob/main/docs/writing-mock-apis.md)

Use `npm run validate-mock-apis` to check everything is good or the errors(Remove the exluded line first)

  - [ ] `models/collections-basic`         @msyyc 
  - [ ] `models/collections-models`     @msyyc 
  - [ ] `models/enum-properties`        @msyyc 
  - [ ] `models/inheritance`                  @weidongxu-microsoft 
  - [ ] `models/input-basic`                 @weidongxu-microsoft 
  - [ ] `models/readonly-properties`   @weidongxu-microsoft 
  - [ ] `models/roundtrip-basic`          @qiaozha 
  - [ ] `models/nested-models`           @qiaozha
  - [ ] `models/primitive-properties`    @qiaozha
  - [ ] `models/optional-properties`    @pshao25 
  - [ ] `models/output-basic`               @pshao25 
  - [ ] `clients/interfaces`                      @pshao25

lirenhe:
Discussed with the team and each SDK team will pick up a couple of the time for Mock API

annelo-msft:
We will want to also reconcile these test cases with [the efforts to port autorest.testserver tests directly](https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http#from-testserver) - we don't want to duplicate tests for given features/feature groups, so we can stay clear about which languages support which features.  If we have duplicate test cases, and different languages use different tests for the same feature, it will be hard to communicate about feature parity across languages going forward.  @lmazuel FYI.

annelo-msft:
Per @lmazuel, please merge "smart Salmon" into inheritance/main.cadl as follows:

Add:
```
model DerivedFromBaseClassWithDiscriminatorC extends DerivedFromBaseClassWithDiscriminatorB {
  discriminatorProperty: "C";
}
```

PR #55, opened 2022-08-18
Add mockapi

msyyc:
https://github.com/Azure/cadl-ranch/issues/39
add mockapi for:
`models/collections-basic`    
 `models/collections-models`    
`models/enum-properties`


packages/cadl-ranch-specs/http/models/collections-basic/mockapi.ts
```diff
@@ -0,0 +1,24 @@
+import { passOnSuccess, mockapi, json } from "@azure-tools/cadl-ranch-api";
+import { ScenarioMockApi } from "@azure-tools/cadl-ranch-api";
+
+export const Scenarios: Record<string, ScenarioMockApi> = {};
+
+Scenarios.CollectionPropertiesBasic_sendCollectionModel = passOnSuccess(
+  mockapi.post("/models/collections/basic", (req) => {
+    const expectedBody = { requiredStringList: ["post"], requiredIntList: [100] };
```
pshao25, line 8: Is it better we just check whether the field exists instead of a hardcode value?

packages/cadl-ranch-specs/http/models/collections-models/mockapi.ts
```diff
@@ -0,0 +1,32 @@
+import { passOnSuccess, mockapi, json } from "@azure-tools/cadl-ranch-api";
+import { ScenarioMockApi } from "@azure-tools/cadl-ranch-api";
+
+export const Scenarios: Record<string, ScenarioMockApi> = {};
+
+Scenarios.ModelCollectionProperties_sendCollectionModel = passOnSuccess(
+  mockapi.post("/models/collections/models", (req) => {
+    const expectedBody = {
+      requiredModelCollection: [{ requiredStringList: ["post_required"], requiredIntList: [100] }],
```
pshao25, line 9: Isn't the field called `optionalModelCollection`?

packages/cadl-ranch-specs/http/models/collections-basic/mockapi.ts
msyyc: I think we could discuss about the check rules in next LLC scrum meeting

packages/cadl-ranch-specs/http/models/collections-basic/mockapi.ts
```diff
@@ -0,0 +1,24 @@
+import { passOnSuccess, mockapi, json } from "@azure-tools/cadl-ranch-api";
+import { ScenarioMockApi } from "@azure-tools/cadl-ranch-api";
+
+export const Scenarios: Record<string, ScenarioMockApi> = {};
+
+Scenarios.CollectionPropertiesBasic_sendCollectionModel = passOnSuccess(
+  mockapi.post("/collection-properties-basic/models", (req) => {
+    const expectedBody = { requiredStringList: ["post"], requiredIntList: [100] };
+    req.expect.coercedBodyEquals(expectedBody);
```
timotheeguerin, line 9:
```suggestion
    req.expect.bodyEquals(expectedBody);
```

timotheeguerin: coercedBodyEquals only used to give this for date manipulation

timotheeguerin: same for all the other uses

packages/cadl-ranch-specs/http/models/enum-properties/main.cadl
```diff
@@ -10,7 +10,7 @@ using Azure.Core.Foundations;
 @server("http://localhost:3000", "TestServer endpoint")
 @doc("Illustrates input, output, and round-trip models with required enum type properties.")
 @route("/enum-properties-basic")
-namespace Cadl.TestServer.EnumPropertiesBasic;
+namespace EnumPropertiesBasic;
```
timotheeguerin, line 13: Wonder if all those scenarios that are related to models. Should maybe be `Model.XYZ` so `Model.EnumProperties`. This would help keeping a good structure in the coverage dashboard where we can group by area well.

 Does basic here also add any value? We are just testing a model with enum properties works fine.

iscai-msft: Second the `Model` namespace, esp for the dashboard. With that we can also feel more comfortable flattening out the model test folder

msyyc: I think `model.xxx` is good idea for a clear coverage dashboard and flatting test folder. 
About `Does basic here also add any value? We are just testing a model with enum properties works fine.` `extensible-enums` mainly test `@knownValues` and `models/enum-properties` test both basic enum and `@knownValues`

packages/cadl-ranch-specs/http/models/collections-basic/main.cadl
```diff
@@ -10,7 +10,7 @@ using Azure.Core.Foundations;
 @server("http://localhost:3000", "TestServer endpoint")
 @doc("Illustrates input, output, and round-trip models with required collection type properties.")
 @route("/collection-properties-basic")
-namespace Cadl.TestServer.CollectionPropertiesBasic;
+namespace Model.CollectionPropertiesBasic;
```
iscai-msft, line 13: Can just be `Models.CollectionProperties`

packages/cadl-ranch-specs/http/models/enum-properties/main.cadl
```diff
@@ -10,7 +10,7 @@ using Azure.Core.Foundations;
 @server("http://localhost:3000", "TestServer endpoint")
 @doc("Illustrates input, output, and round-trip models with required enum type properties.")
 @route("/enum-properties-basic")
-namespace Cadl.TestServer.EnumPropertiesBasic;
+namespace Model.EnumPropertiesBasic;
```
iscai-msft, line 13: `Models.EnumProperties`

packages/cadl-ranch-specs/http/models/enum-properties/main.cadl
iscai-msft: I think the difference between those tests is more:
- `extensible-enums` tests enums as a standalone entity
- `models.enumproperties` tests enums as a property of models

I think generally I would recommend tests like these be flattened, but we are also placing a special emphasis on models (because that's what elevates cadl from dpg, especially for our first sprint). So I think it's still valid to keep these separate (at least for now), and make clear the purpose of each

timotheeguerin: So this test test that a model can have property of type enum or extensible enum. It shouldn't actually be testing teh enum and extensible enum too much, just that they work with model. That job should be left to an Enum and `ExtensibleEnum` set of scenarios that are not tied to model.

I think this is the same for a lot of the other model tests. If the goal is really just to check that you can use type X or that it works with optional properties, etc. on model property then it should limit the test to that. If the goal is to test the type X in depth then it should have its own dedicated spec outside of the model scope.

iscai-msft: Yeah agreed, and that has me thinking: Do we want to have dedicated tests (i.e. a dedicated test for enums, and a dedicated test for dictionaries), and just one model cadl that tests all of these as properties (i.e. this model has an enum and a dictionary property). I'm just worried we're going to balloon with all of these tiny cadl files that could easily be one cadl file

msyyc: too many tiny cadl files will make test troublesome while a big cadl that contains many kinds of property may make test too complicated. I think we could use eclectic way: one cadl shall test for 3~5 kinds of property. It takes both flexibility and complexity into account.

timotheeguerin: talked with @iscai-msft last night we believe we'll structure things this way:
- one spec for model property types. Each type that we want to test gets a `get` and a `put`. This WILL NOT be testing the types in depth. Just make sure they work fine as model properties.
- one spec for optional model properties
- one spec for nullable properties
- one spec for model visibility(read only write only)
- one spec per different entity type(enum, union, array, etc.) where we test that entity in depth but not with models

iscai-msft:
@msyyc I've merged these tests into my `property-types` test (the one combining all of the diff property types) https://github.com/Azure/cadl-ranch/pull/55#discussion_r956069939

Are you ok if I close this PR, because the mock tests are now merged, just in another folder? Thanks for your contribution!

PR #56, opened 2022-08-18
add clients/authentication in specs

weidongxu-microsoft:
mock API will be added later (in this PR), if the spec is OK.

mock API likely will only validate the existence of the x-ms-api-key header.

get a request that no code if no doc. will also add the docstring if spec is good.


packages/cadl-ranch-specs/http/clients/authentication/main.cadl
```diff
@@ -0,0 +1,26 @@
...
+using Cadl.Http;
+using Azure.Core.Foundations;
+
+@serviceTitle("ClientAuthentication")
+@serviceVersion("1.0.0")
+@server("http://localhost:3000", "TestServer endpoint")
+@doc("Illustrates clients generated with OAuth2 and ApiKey authentication.")
+
+@useAuth(OAuth2Auth<[OAuth2AuthFlow]> | ApiKeyAuth<ApiKeyLocation.header, "x-ms-api-key">)
+namespace Cadl.TestServer.ClientAuthentication;
```
timotheeguerin, line 13:
```suggestion
namespace OAuth2Authentication;
```
this bad pattern of having the `Cadl.TestServer` namespace was started, I think we can just have the namespace be the scenario group we are testing.

packages/cadl-ranch-specs/http/clients/authentication/main.cadl
```diff
@@ -0,0 +1,26 @@
...
+    tokenUrl: "https://api.example.com/oauth2/token";
+    refreshUrl: "https://api.example.com/oauth2/refresh";
+    scopes: [
+        "https://api.example.com/.default"
+    ]
+}
+
+@route("/clients/authentication/get")
+@get
+op get(): OkResponse | ErrorResponse;
```
timotheeguerin, line 26: Would need to add the `@scenario` decorator to mark the endpoint as a scenario.
I expect at least of have something like
- `authenticatedCall`: Expect the test to send a valid oauth2 token. 

and then not sure if we need that:
- `invalidAuthenticationCall` : Expect the test to send a invalid oauth2 token and fail the request and the client should handle fine.

packages/cadl-ranch-specs/http/clients/authentication/main.cadl
```diff
@@ -0,0 +1,26 @@
+import "@cadl-lang/rest";
```
timotheeguerin, line 1: Because authentication has to be at the service level(for now at least) I would structure the files more or less like that:
- [clients/]authentication/
  - oauth2/
  - apiKey/
  - basic/
  - multiAuth/
  - etc...

weidongxu-microsoft:
Thanks @timotheeguerin for the feedback.

I will restructure this scenario.
1. namespace "OAuth2Authentication"
2. `authenticatedCall`, and if possible `invalidAuthenticationCall` with 401
3. update folder structure, let's first start with `apiKey`
4. add mock API implementation

packages/cadl-ranch-specs/http/authentication/apiKey/mockapi.ts
```diff
@@ -0,0 +1,24 @@
...
+Scenarios.ApiKeyAuthentication_ApiKey_authenticated = passOnSuccess(
+  mockapi.get("/authentication/api-key/authenticated", (req) => {
+    req.expect.containsHeader("x-ms-api-key", "valid-key");
+    return { status: 204 };
+  }),
+);
+
+Scenarios.ApiKeyAuthentication_ApiKey_invalidAuthentication = passOnSuccess(
+  mockapi.get("/authentication/api-key/invalid-authentication", (req) => {
+    return { status: 403,
```
weidongxu-microsoft, line 15: Since api-key is usually not via "Authentication" header, return 403 (instead of 401 which would require an 'WWW-Authenticate` header for challenge.

Checked storage and search, both uses 403 (though storage actually send the SharedKey via Authentication header).

---

Java azure-core actually forbid sending key on HTTP.
https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/AzureKeyCredentialPolicy.java#L33

Hence testing is done via Postman.

timotheeguerin: We could make the mock api support https for completeness, probably would have to just do one of the following:
- Install the mock api certificate before calling. A bit anoying but would be the most accurate testing.
- Disable cert validation on that test. Usually much easier but I guess that depends on the language.

weidongxu-microsoft: Maybe we can delay the decision of whether to have HTTPS, depends on whether other language get to the problem.

If the problem is Java only, we probably do a less accurate testing by somehow disable that check in Cadl Java tests.

packages/cadl-ranch-specs/http/authentication/apiKey/main.cadl
```diff
@@ -0,0 +1,30 @@
...
+
+@serviceTitle("ApiKeyAuthentication")
+@serviceVersion("1.0.0")
+@server("http://localhost:3000", "TestServer endpoint")
+@route("/authentication/api-key")
+@doc("Illustrates clients generated with ApiKey authentication.")
+@useAuth(ApiKeyAuth<ApiKeyLocation.header, "x-ms-api-key">)
+namespace ApiKeyAuthentication;
+
+interface ApiKey {
```
timotheeguerin, line 16: Do we need this interface? I think just having the operation at the root would make the scenario name cleaner

weidongxu-microsoft: Sure, I will remove them (was following the "extensible-enums" file which having a String interface).

timotheeguerin: Yeah I did that there because I think we could have other extensible enum types in the future. But here I think this is just redundant.

timotheeguerin: I think in general all specs  don't need to have the exact same hierachy. Some are going to need more or less nesting

packages/cadl-ranch-specs/http/authentication/apiKey/main.cadl
```diff
@@ -0,0 +1,28 @@
+import "@cadl-lang/rest";
+import "@azure-tools/cadl-azure-core";
+import "@azure-tools/cadl-ranch-expect";
+
+using Cadl.Http;
+using Azure.Core.Foundations;
+
+@serviceTitle("ApiKeyAuthentication")
```
timotheeguerin, line 8: Just a tought not for your PR. Wonder if we could provide a new decorator in cadl-ranch-expect that does reduce a bunch of the service level boilerplate. Every spec needs pretty much the same 
- serviceVersion 
- server
- serviceTitle(take the namespace)

timotheeguerin (approved):
Perfect!

packages/cadl-ranch-specs/http/authentication/apiKey/main.cadl
weidongxu-microsoft: Agree. Should be handy.

There is also a requirement that SDK emitter should fail on missing "doc" decorator (to enforce best-practice on Cadl file). https://github.com/Azure/autorest.csharp/issues/2621

Hence we might like any new Cadl spec here having full doc. I could ping dev that they would take best efforts to have their new/modified Cadl file having full doc.

timotheeguerin: I did already say that on the chat but that requirement should absolutely not be there. The csharp generator should be generating a SDK if there is no doc, that's not the job of csharp generator to decide that. There is/will be linter enforcing doc for Azure services but if someone wants to generate CSharp without doc, they should be able to.

timotheeguerin: Guess, need to talk with Laurent about this as seems like he signed up on this. It is ridiculous that if we are lacking doc the generator will just blow up, this would make generator pretty useless outside of Azure.

weidongxu-microsoft: Got it. I see.

It does feel like a lint/validation rule that should happen to Cadl file itself (with a tool maybe, but SDK emitter might not be the best one in this case).

timotheeguerin: Yes exactly, right now with this behavior this mean someone will have a totally valid Cadl file with no error and then in the csharp(and seems like other generators) it will blow up giving no editor support for where the error is as well.

timotheeguerin: Created an issue for the decorator https://github.com/Azure/cadl-ranch/issues/72

packages/cadl-ranch-specs/http/authentication/apiKey/mockapi.ts
```diff
@@ -0,0 +1,25 @@
...
+  mockapi.get("/authentication/api-key/authenticated", (req) => {
+    req.expect.containsHeader("x-ms-api-key", "valid-key");
+    return { status: 204 };
+  }),
+);
+
+Scenarios.ApiKeyAuthentication_invalidAuthentication = passOnSuccess(
+  mockapi.get("/authentication/api-key/invalid-authentication", (req) => {
+    return {
+      status: 403,
```
srnagar, line 16: Should we have a scenario for 401 too if the header name is incorrect or missing?

srnagar (approved):

packages/cadl-ranch-specs/http/authentication/apiKey/mockapi.ts
timotheeguerin: Would generator handle those differently?

weidongxu-microsoft: copy one comment [above](https://github.com/Azure/cadl-ranch/pull/56#discussion_r951378553)

> Since api-key is usually not via "Authentication" header, return 403 (instead of 401 which would require a WWW-Authenticate header in response for challenge).
>
> Checked storage and search, both uses 403 (though storage actually send the SharedKey via Authentication header).

`WWW-Authenticate` is required for 401, per [rfc7235](https://www.rfc-editor.org/rfc/rfc7235#section-4.1).
But so far I didn't find a case in our backend in ApiKey case (admit I am not an expert on auth, so maybe we need change this later).

E.g., if I send "https://management.azure.com/subscriptions/ec0aa5f7-9e78-40c9-85cd-535c6305b380/resourceGroups/rg-weidxu?api-version=2021-01-01" which require Bearer auth (the OAuth2 case), I will get

```
401

WWW-Authenticate: Bearer authorization_uri="https://login.windows.net/72f988bf-86f1-41af-91ab-2d7cd011db47", error="invalid_token", error_description="The authentication failed because of missing 'Authorization' header."

{
    "error": {
        "code": "AuthenticationFailed",
        "message": "Authentication failed. The 'Authorization' header is missing."
    }
}
```

---

Code generator does not handle it differently (as this is runtime behavior, not in swagger or cadl). Usually azure-core or azure-identity (on each language) would handle the ApiKey and OAuth2 flow.

The difference of 401 and 403 in runtime would likely get SDK client throw a different error/exception.

timotheeguerin: don't really have a strong opinion on this, if you feel like having 401 and 403 is valuable I'm down to add it otherwise I think this PR Is good! We can always add the 401 scenario later if we find more value in it.

weidongxu-microsoft: So far, no opinion from me. 401 will likely happen to OAuth2 case.

@qiaozha @msyyc @pshao25 for opinion from SDK.

pshao25: In my understanding, we only need to test that expected key are set to header `x-ms-api-key`. We even don't the need the `ApiKeyAuthentication_invalidAuthentication` scenario, because it is not what a service need to handle.

When a service receives a request, it just redirects to an auth provider with the auth headers. If auth provider says it fails, service would send back the suth result to the client. 401/403 is what auth provider provides, we clients should not do anything about this.

timotheeguerin: well however a service implement the authentication logic is irrelevant. The result from a client point of view is that it can succeed or fail. And it is definitely valuable to have a scenario where you check the SDK doesn't blow up with unhelpful information if the service return authentication failure.

weidongxu-microsoft: > In my understanding, we only need to test that expected key are set to header `x-ms-api-key`. We even don't need the `ApiKeyAuthentication_invalidAuthentication` scenario, because it is not what a service need to handle.
> 
> When a service receives a request, it just redirects to an auth provider with the auth headers. If auth provider says it fails, service would send back the auth result to the client. 401/403 is what auth provider provides, we clients should not do anything about this.
> 
> So my opinion is to delete this scenario.

No, 401/403 is what service respond. You can try with any of your existing SDK.

So far none of them is about auth server/provider.

PR #57, opened 2022-08-18
Add mock apis

pshao25:
Add below mock APIs.
 
models/optional-properties
 models/output-basic
 clients/interfaces


packages/cadl-ranch-specs/http/clients/interfaces/main.cadl
```diff
@@ -32,7 +32,6 @@ interface dogs {
 
   @scenario
   @scenarioDoc("Illustrate grouping operations on subclient.")
-  @route("/models")
```
haolingdong-msft, line 35: why delete @route?

pshao25: It looks like added wrong, because other operations don't have this.

packages/cadl-ranch-specs/http/models/optional-properties/mockapi.ts
```diff
@@ -0,0 +1,53 @@
+import { passOnSuccess, mockapi, ValidationError, json } from "@azure-tools/cadl-ranch-api";
+import { ScenarioMockApi } from "@azure-tools/cadl-ranch-api";
+
+export const Scenarios: Record<string, ScenarioMockApi> = {};
+
+Scenarios.OptionalProperties_sendOptionalPropertyModel = passOnSuccess(
```
timotheeguerin, line 6: For example here I would expect the scenarios for a model with optional properties to be:
- Send only required properties(Mock api can just call `req.expectBodyEquals()`) which should make sure the optional properrties are not send with `null`
- Receive only required properties
- Send with optional properties
- Receive with optional properties

pshao25: > Send only required properties(Mock api can just call req.expectBodyEquals()) which should make sure the optional properrties are not send with null

There is no required properties in the models. And I already checked the properties in below `checkPropertyValid`. What do you mean by `required properties`.

packages/cadl-ranch-specs/http/clients/interfaces/mockapi.ts
```diff
@@ -0,0 +1,54 @@
+import { passOnSuccess, mockapi, json } from "@azure-tools/cadl-ranch-api";
+import { ScenarioMockApi } from "@azure-tools/cadl-ranch-api";
+
+export const Scenarios: Record<string, ScenarioMockApi> = {};
+
+Scenarios.MultiInterfaceClient_dogs_getDogs = passOnSuccess(
```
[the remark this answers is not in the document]
pshao25: What scenario names do you expect to have? I think `clients\interfaces` is to validate different interfaces can have different routes.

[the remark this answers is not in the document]
timotheeguerin: then the scenarios should be describing that. If someone looks at the coverage dashboard and see that a generator support or doesn't `getDogs` that means absolutely nothing.

Scenario don't need to be on operation they can be on Namespace too so that you can have a scenario where you call multiple operations. But I guess I can also allow putting in on interface. So we could have a scenario `OperationGroup_group1` and `OperationGroup_group2`

Also note that I am not sure if the goal is to test operation group that this spec is up to date with the latest decision on what operation group should be in cadl.

packages/cadl-ranch-specs/http/models/optional-properties/mockapi.ts
timotheeguerin: I see, so the goal of this is to test a model with different optional property types. Maybe then should be named `OptionalPropertyTypes` and then test should describe which type they are checking work fine as am optional property

iscai-msft:
@pshao25 I added your optional property tests in my refactoring PR for the optional property tests! Thanks for your help with optional properties!

PR #59, opened 2022-08-18
Infra cleaning

timotheeguerin:
- Use `workspace:~` for internal package versions. Makes it easier to bump.
- Fix some specs
- Add owners
- Make pnpm minimum version `7.6.0` as this is where `auto-install-peers` was addded


iscai-msft (approved):
:shipit:

PR #62, opened 2022-08-19
add mocktest for readonly-properties

XiaofeiCao:
fix https://github.com/Azure/autorest.java/issues/1630

### DIFF
--- a/packages/cadl-ranch-specs/http/models/readonly-properties/main.cadl
+++ b/packages/cadl-ranch-specs/http/models/readonly-properties/main.cadl
@@ -79,4 +79,4 @@ op getOptionalPropertyModel(): OutputModel | ErrorResponse;
 @scenarioDoc("Generate, send, and receive round-trip model with readonly properties.")
 @route("/models")
 @put
-op setOptionalPropertyModel(@body input: RoundTripModel): RoundTripModel | ErrorResponse;
+op setOptionalPropertyModel(@body input: ReadonlyModel): RoundTripModel | ErrorResponse;
--- a/packages/cadl-ranch-specs/http/models/readonly-properties/mockapi.ts
+++ b/packages/cadl-ranch-specs/http/models/readonly-properties/mockapi.ts
@@ -1,0 +1,34 @@
+import { passOnSuccess, mockapi, json, ValidationError } from "@azure-tools/cadl-ranch-api";
+import { ScenarioMockApi } from "@azure-tools/cadl-ranch-api";
+
+export const Scenarios: Record<string, ScenarioMockApi> = {};
+
+let mockRoundTripModel = {
+  requiredReadonlyString: "requiredReadonlyStringValue",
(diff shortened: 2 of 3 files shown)


packages/cadl-ranch-specs/http/models/readonly-properties/main.cadl
```diff
@@ -79,4 +79,4 @@ op getOptionalPropertyModel(): OutputModel | ErrorResponse;
 @scenarioDoc("Generate, send, and receive round-trip model with readonly properties.")
 @route("/models")
 @put
-op setOptionalPropertyModel(@body input: RoundTripModel): RoundTripModel | ErrorResponse;
```
XiaofeiCao, line 82: All of `RoundTripModel`'s properties are read-only, thus not suitable for request body.

packages/cadl-ranch-specs/http/models/readonly-properties/main.cadl
```diff
@@ -64,19 +64,52 @@ model OutputModel {
   optionalReadonlyIntList?: int32[];
 }
 
-@doc("Round-trip model with readonly properties.")
+@doc("Input model for setting values.")
+model InputModel {
```
iscai-msft, line 68: I don't fully see the point of the `InputModel`, because all it is is flattened into the `RoundTripModel`. I think it's valid to have a test for a model with only `readonly` properties, and a model that has a mix of both normal properties and readonly properties, but I don't see the point of a model with only normal properties for the `readonly-properties` test

XiaofeiCao: Hi, the purpose here is to mimic the scenario when a user puts a resource with non-readonly properties(those from `InputModel` and get a response from server with both the readonly properties and the non readonly ones(like create a resource with `name` properties and get a response containing `id`, `name`, etc).

You are right, maybe I'll remove the `InputModel` and flatten the properties to `RoundTripModel` instead.

packages/cadl-ranch-specs/http/models/readonly-properties/main.cadl
```diff
@@ -64,19 +64,52 @@ model OutputModel {
...
+    requiredString: "requiredStringValue",
+  },
+  requiredReadonlyStringList: ["value1", "value2"],
+  requiredReadonlyIntList: [1, 2, 3, 4, 5],
+}
+""")
 @route("/models")
 @get
-op getOptionalPropertyModel(): OutputModel | ErrorResponse;
+op getReadOnlyProperties(): RoundTripModel | ErrorResponse;
```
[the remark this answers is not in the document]
XiaofeiCao: It's for consistency with the other op below. But if we want to return a model with only readonly properties, we can mix all properties in RoundTripModel, and filter the readonly properties in `OutputModel` by using `WithVisibility("read")`.

PR #63, opened 2022-08-19
add mockapi for models/input-basic

haolingdong-msft:
Link https://github.com/Azure/autorest.java/issues/1631

- add mockapi.ts. validate if the required body exists, if not, throws `ValidationError`.
- update namespace according to this comment: https://github.com/Azure/cadl-ranch/pull/55#discussion_r951570585


packages/cadl-ranch-specs/http/models/input-basic/mockapi.ts
```diff
@@ -0,0 +1,15 @@
...
+export const Scenarios: Record<string, ScenarioMockApi> = {};
+
+Scenarios.InputBasic_getModel = passOnSuccess(
+  mockapi.get("/input-basic/models", (req) => {
+    if (req.body && req.body["requiredString"] && req.body["requiredInt"]) {
+      return {
+        status: 200,
+      };
+    } else {
+      throw new ValidationError(`Expected InputModel with required properties`, null, null);
```
haolingdong-msft, line 12: Hi @timotheeguerin  
`ValidationError` (https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-api/src/validation-error.ts) has different format with `ErrorResponse`  defined in `.cadl` file (https://github.com/Azure/cadl-azure/blob/main/packages/cadl-azure-core/lib/foundations.cadl#L71). Is it better to change `ValidationError` format to make it consistent with `ErrorResponse`?

packages/cadl-ranch-specs/http/models/input-basic/main.cadl
```diff
@@ -10,7 +10,7 @@ using Azure.Core.Foundations;
 @server("http://localhost:3000", "TestServer endpoint")
 @doc("Illustrates an input-only model with required reference and value type properties.")
 @route("/input-basic")
-namespace Cadl.TestServer.InputBasic;
+namespace Models.InputBasic;
```
iscai-msft, line 13: I think we should go with just `Models.Input`. I don't htink `basic` adds anything

Also I'm going to be restructuring the model tests so there aren't a bunch of different cadl folders for little scenarios. Here's the plan: https://github.com/Azure/cadl-ranch/pull/55#discussion_r956069939, it will mainly be a re-org of the existing tests

PR #74, opened 2022-08-24
migrate bodycomplex in autorest.testserver (for complex polymorphism and dictionary)

changlong-liu:
This PR is for migrating [body-complex](https://github.com/Azure/autorest.testserver/blob/main/swagger/body-complex.json) in autorest.testserver into cadl.
Since there are already projects for primitives, collections and readonly in cadl-ranch, I don't add related tests in body-complex to here.
I split the remains into two packages as `dictionary` and `inheritance-complex`:
- dictionary: for cadl datatype `Record`
- inheritance-complex: provide testcases for complex polymorphic scenario

### DIFF
--- a/packages/cadl-ranch-specs/http/README.md
+++ b/packages/cadl-ranch-specs/http/README.md
@@ -25,3 +25,5 @@ The tests in the http/models directory describe scenarios for generating models 
 - [nested-models](https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/models/nested-models) - Generate, send, and receive input, output, and round-trip models with required nested model properties.
 - [readonly-properties](https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/models/readonly-properties) - Generate, send, and receive input, output, and round-trip models with readonly properties.
 - [inheritance](https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/models/inheritance) - Generate, send, and receive round-trip models polymorphic models.
+- [inheritance-complex](https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/models/inheritance-complex) - Generate, send, and receive complex polymorphic models.
+- [dictionary](https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/models/dictionary) - Generate, send, and receive dictionary properties.
--- a/packages/cadl-ranch-specs/http/models/dictionary/main.cadl
+++ b/packages/cadl-ranch-specs/http/models/dictionary/main.cadl
@@ -1,0 +1,57 @@
+import "@azure-tools/cadl-azure-core";
+import "@azure-tools/cadl-ranch-expect";
+
+using Cadl.Http;
+using Azure.Core.Foundations;
+
+@serviceTitle("Dictionary in Complex Test Service")
+@serviceVersion("2016-02-29")
+@server("http://localhost:3000", "Testserver endpoint")
+@doc("Illustrates complex model generation, as well as serialization and deserialization.")
+namespace Cadl.Testserver.BodyComplexDictionary;
+
+model dictionary_wrapper {
+  defaultProgram?: Record<string>;
+}
+
+@scenario
+@route("/complex/dictionary/typed")
+namespace dictionary {
+  @route("/valid")
+  @scenarioDoc("Get complex types with dictionary properties")
+  @get
+  op getValid(): dictionary_wrapper | ErrorResponse;
+
+  @route("/valid")
+  @scenarioDoc("Put complex types with dictionary properties")
+  @put
+  op putValid(
+    @doc("Please put a dictionary with 5 key-value pairs: \"txt\":\"notepad\", \"bmp\":\"mspaint\", \"xls\":\"excel\", \"exe\":\"\", \"\":null")
+    @body
+    complexBody: dictionary_wrapper
+  ): {
+    @statusCode statusCode: 200;
+  } | ErrorResponse;
+
+  @route("/empty")
+  @scenarioDoc("Get complex types with dictionary property which is empty")
+  @get
+  op getEmpty(): dictionary_wrapper | ErrorResponse;
+
+  @route("/empty")
+  @scenarioDoc("Put complex types with dictionary properties which is empty")
+  @put
+  op putEmpty(@doc("Please put an empty dictionary") @body complexBody: dictionary_wrapper): {
+    @statusCode statusCode: 200;
+  } | ErrorResponse;
+
+  @route("/null")
+  @scenarioDoc("Get complex types with dictionary property which is null")
+  @get
+  op getNull(): dictionary_wrapper | ErrorResponse;
+
+  @route("/notprovided")
+  @scenarioDoc("Get complex types with dictionary property while server doesn't provide a response payload")
+  @get
+  op getNotProvided(): dictionary_wrapper | ErrorResponse;
+}
--- a/packages/cadl-ranch-specs/http/models/inheritance-complex/main.cadl
+++ b/packages/cadl-ranch-specs/http/models/inheritance-complex/main.cadl
@@ -1,0 +1,256 @@
+import "@cadl-lang/rest";
+import "@azure-tools/cadl-azure-core";
+import "@azure-tools/cadl-ranch-expect";
+
+using Cadl.Rest;
+using Cadl.Http;
+using Azure.Core.Foundations;
+
+@serviceTitle("Polymorphism in Complex Test Service")
+@serviceVersion("2016-02-29")
+@server("http://localhost:3000", "Testserver endpoint")
+@doc("Illustrates complex model generation, as well as serialization and deserialization.")
+namespace Cadl.Testserver.BodyComplexPolymorphism;
+
+@knownValues(CMYKColorsValues)
+model CMYKColors is string {}
+
+enum CMYKColorsValues {
+  cyan,
+  Magenta,
+  YELLOW,
+  blacK,
+}
+
+model basic {
+  @doc("Basic Id")
+  id?: int32;
+
+  @doc("Name property with a very long description that does not fit on a single line and a line break.")
+  name?: string;
+  color?: CMYKColors;
+}
+
+model pet {
+  id?: int32;
+  name?: string;
+}
+
+model cat extends pet {
+  color?: string;
+  hates?: dog[];
+}
+
+model siamese extends cat {
+  breed?: string;
+}
+
+model dog extends pet {
+  food?: string;
+}
+
+@discriminator("fish_type")
+model DotFish {
+  species?: string;
+}
+
+model DotSalmon extends DotFish {
+  fish_type: "DotSalmon";
+  location?: string;
+  iswild?: boolean;
+}
+
+model DotFishMarket {
+  sampleSalmon?: DotSalmon;
+  salmons?: DotSalmon[];
+  sampleFish?: DotFish;
+  fishes: DotFish[];
+}
+
+@discriminator("fishtype")
+model Fish {
+  species?: string;
+  length: float32;
+  siblings?: Fish[];
+}
+
+model salmon extends Fish {
+  fishtype: "salmon";
+  location?: string;
+  iswild?: boolean;
+}
+
+model smart_salmon extends Fish {
+  fishtype: "smart_salmon";
+  college_degree?: string;
+}
+
+@discriminator("sharktype")
+model shark extends Fish {
+  fishtype: "shark";
+  age?: int32;
+  birthday: zonedDateTime;
+}
+
+model sawshark extends shark {
+  sharktype: "sawshark";
+  picture?: bytes;
+}
+
+@doc("Colors possible")
+@knownValues(GoblinSharkColorValues)
+model GoblinSharkColor is string {}
+
+enum GoblinSharkColorValues {
+  pink,
+  gray,
+  brown,
+
+  @doc("Uppercase RED")
+  upperRed: "RED",
+
+  @doc("Lowercase RED")
+  lowerRed: "red",
+}
+
+model goblinshark extends shark {
+  sharktype: "goblin";
+  jawsize?: int32;
+  color?: GoblinSharkColor = "gray";
+}
+
+model cookiecuttershark extends shark {
+  sharktype: "cookiecuttershark";
+}
+
+@scenario
+@route("/complex/basic")
+namespace basicOps {
+  @route("/valid")
+  @scenarioDoc("Get complex type {id: 2, name: 'abc', color: 'YELLOW'}")
+  @get
+  op getValid(): basic | ErrorResponse;
+
+  @route("/valid")
+  @scenarioDoc("Please put {id: 2, name: 'abc', color: 'Magenta'}")
+  @put
+  op putValid(@header "api-version": string, @body complexBody: basic): {
+    @statusCode statusCode: 200;
+  } | ErrorResponse;
+
+  @route("/invalid")
+  @scenarioDoc("Get a basic complex type that is invalid for the local strong type")
+  @get
+  op getInvalid(): basic | ErrorResponse;
+
+  @route("/empty")
+  @scenarioDoc("Get a basic complex type that is empty")
+  @get
+  op getEmpty(): basic | ErrorResponse;
+
+  @route("/null")
+  @scenarioDoc("Get a basic complex type whose properties are null")
+  @get
+  op getNull(): basic | ErrorResponse;
+
+  @route("/notprovided")
+  @scenarioDoc("Get a basic complex type whose properties are null")
+  @get
+  op getNotProvided(): basic | ErrorResponse;
+}
+
+@route("/complex/inheritance/valid")
+namespace inheritance {
+  @scenarioDoc("Get complex types that extend others")
+  @get
+  op getValid(): siamese | ErrorResponse;
+
+  @scenarioDoc("Put complex types that extend others")
+  @put
+  op putValid(
+    @doc("Please put a siamese with id=2, name=\"Siameee\", color=green, breed=persion, which hates 2 dogs, the 1st one named \"Potato\" with id=1 and food=\"tomato\", and the 2nd one named \"Tomato\" with id=-1 and food=\"french fries\".")
+    @body
+    complexBody: siamese
+  ): {
+    @statusCode statusCode: 200;
+  } | ErrorResponse;
+}
+
+@route("/complex/polymorphism")
+namespace polymorphism {
+  @route("/valid")
+  @scenarioDoc("Get complex types that are polymorphic")
+  @get
+  op getValid(): Fish | ErrorResponse;
+
+  @route("/valid")
+  @scenarioDoc("Put complex types that are polymorphic")
+  @put
+  op putValid(
+    @doc("Please put a salmon that looks like this:\n{\n        'fishtype':'Salmon',\n        'location':'alaska',\n        'iswild':true,\n        'species':'king',\n        'length':1.0,\n        'siblings':[\n          {\n            'fishtype':'Shark',\n            'age':6,\n            'birthday': '2012-01-05T01:00:00Z',\n            'length':20.0,\n            'species':'predator',\n          },\n          {\n            'fishtype':'Sawshark',\n            'age':105,\n            'birthday': '1900-01-05T01:00:00Z',\n            'length':10.0,\n            'picture': new Buffer([255, 255, 255, 255, 254]).toString('base64'),\n            'species':'dangerous',\n          },\n          {\n            'fishtype': 'goblin',\n            'age': 1,\n            'birthday': '2015-08-08T00:00:00Z',\n            'length': 30.0,\n            'species': 'scary',\n            'jawsize': 5\n          }\n        ]\n      };")
+    @body
+    complexBody: Fish
+  ): {
+    @statusCode statusCode: 200;
+  } | ErrorResponse;
+
+  @route("/dotsyntax")
+  @scenarioDoc("Get complex types that are polymorphic, JSON key contains a dot")
+  @get
+  op getDotSyntax(): DotFish | ErrorResponse;
+
+  @route("/composedWithDiscriminator")
+  @scenarioDoc("Get complex object composing a polymorphic scalar property and array property with polymorphic element type, with discriminator specified. Deserialization must NOT fail and use the discriminator type specified on the wire.")
+  @get
+  op getComposedWithDiscriminator(): DotFishMarket | ErrorResponse;
+
+  @route("/composedWithoutDiscriminator")
+  @scenarioDoc("Get complex object composing a polymorphic scalar property and array property with polymorphic element type, without discriminator specified on wire. Deserialization must NOT fail and use the explicit type of the property.")
+  @get
+  op getComposedWithoutDiscriminator(): DotFishMarket | ErrorResponse;
+
+  @route("/complicated")
+  @scenarioDoc("Get complex types that are polymorphic, but not at the root of the hierarchy; also have additional properties")
+  @get
+  op getComplicated(): Fish | ErrorResponse;
+
+  @route("/complicated")
+  @scenarioDoc("Put complex types that are polymorphic, but not at the root of the hierarchy; also have additional properties")
+  @put
+  op putComplicated(@body complexBody: Fish): {
+    @statusCode statusCode: 200;
+  } | ErrorResponse;
+
+  @route("/missingdiscriminator")
+  @scenarioDoc("Put complex types that are polymorphic, omitting the discriminator")
+  @put
+  op putMissingDiscriminator(@body complexBody: salmon): salmon | ErrorResponse;
+
+  @route("/missingrequired/invalid")
+  @scenarioDoc("Put complex types that are polymorphic, attempting to omit required 'birthday' field - the request should not be allowed from the client")
+  @put
+  op putValidMissingRequired(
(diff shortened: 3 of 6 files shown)


packages/cadl-ranch-specs/http/models/dictionary/main.cadl
```diff
@@ -0,0 +1,57 @@
...
+using Cadl.Http;
+using Azure.Core.Foundations;
+
+@serviceTitle("Dictionary in Complex Test Service")
+@serviceVersion("2016-02-29")
+@server("http://localhost:3000", "Testserver endpoint")
+@doc("Illustrates complex model generation, as well as serialization and deserialization.")
+namespace Cadl.Testserver.BodyComplexDictionary;
+
+model dictionary_wrapper {
```
msyyc, line 13: Model name shall be camel-case

changlong-liu: changed to `dictionaryWrapper`

packages/cadl-ranch-specs/http/models/dictionary/main.cadl
```diff
@@ -0,0 +1,53 @@
+import "@azure-tools/cadl-azure-core";
+import "@azure-tools/cadl-ranch-expect";
+
+using Cadl.Http;
+using Azure.Core.Foundations;
+
+@serviceTitle("Dictionary in Complex Test Service")
```
iscai-msft, line 7: For cadl-ranch we want to be sure that every scenario is valid in itself. You can use `body complex` as inspiration to find missing scenarios in the testserver, but the test shouldn't be tied to body complex in any way

packages/cadl-ranch-specs/http/models/dictionary/main.cadl
```diff
@@ -0,0 +1,53 @@
...
+@serviceVersion("2016-02-29")
+@server("http://localhost:3000", "Testserver endpoint")
+@doc("Illustrates dictionary generation, as well as serialization and deserialization.")
+namespace Cadl.Testserver.BodyComplexDictionary;
```
iscai-msft, line 11: `Models.Dictionary`

packages/cadl-ranch-specs/http/models/dictionary/main.cadl
```diff
@@ -0,0 +1,53 @@
...
+
+model dictionaryWrapper {
+  defaultProgram?: Record<string>;
```
iscai-msft, line 14: `defaultProgram` isn't a good property name, we can do `optionalDictionaryProperty` (and maybe even add a `requiredDictionaryProperty`.

packages/cadl-ranch-specs/http/models/dictionary/main.cadl
```diff
@@ -0,0 +1,53 @@
...
+}
+
+@scenario
+@route("/complex/dictionary/typed")
+namespace dictionary {
+  @route("/valid")
+  @scenarioDoc("Put complex types with dictionary properties")
+  @put
+  op putValid(
+    @doc("Please put a dictionary with 5 key-value pairs: \"txt\":\"notepad\", \"bmp\":\"mspaint\", \"xls\":\"excel\", \"exe\":\"\", \"\":null")
```
iscai-msft, line 24: The `@doc` property is more for describing the property, i.e. what we would generate if this was a client library. The description of what the mock api tests should expect should be in the `@scenarioDoc` header. See the docs for more info https://github.com/Azure/cadl-ranch/blob/main/docs/writing-scenario-spec.md

iscai-msft (commented):
@changlong-liu can you read over the [docs](https://github.com/Azure/cadl-ranch/tree/main/docs) and align your tests so they aren't tied to `BodyComplex`, and every scenario is clearly documented and necessary? Thanks!

changlong-liu:
> @changlong-liu can you read over the [docs](https://github.com/Azure/cadl-ranch/tree/main/docs) and align your tests so they aren't tied to `BodyComplex`, and every scenario is clearly documented and necessary? Thanks!

Hi @iscai-msft ,  thanks for the information!
I refactored the scenarios and mockapi to align the naming and doc style, and make each scenario for single purpose. The mock apis in PR can get work in my local Python tests.

packages/cadl-ranch-specs/http/models/dictionary-basic/main.cadl
```diff
@@ -0,0 +1,48 @@
+import "@azure-tools/cadl-azure-core";
+import "@azure-tools/cadl-ranch-expect";
+
+using Cadl.Http;
+using Azure.Core.Foundations;
+
+@serviceTitle("DictionaryPropertiesBasicClient")
```
iscai-msft, line 7: we're against using qualifiers like "basic" and "complex". I think the "models/dictionary-basic" code shouldn't be its own test (since we're testing dictionary properties). Instead, it should be added to https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/models/property-types.

iscai-msft: And I just want to clarify that just because it's in the testserver means we have to have it in cadl ranch. Looking at these tests, we don't need separate tests for a dictionary property with values and without, because this testing will be captured by a combination of `optional-properties` and `property-types` tests

packages/cadl-ranch-specs/http/models/inheritance-complex/main.cadl
```diff
@@ -0,0 +1,114 @@
...
+}
+
+@doc("The second level model in three levels inheritance with discriminator and defined a new discriminator.")
+@discriminator("discriminator2")
+model SubModelWithDiscriminator extends BaseModelWithDiscriminator {
+  discriminator1: "SubModel";
+}
+
+@doc("The second level model in three levels inheritance with discriminator.")
+model DerivedFromBaseModelWithDiscriminator extends BaseModelWithDiscriminator {
```
iscai-msft, line 43: I understand that you're following the naming schema of the existing `inheritance` models, but I've been meaning to change that one as well, since it's not clear what the models are. @timotheeguerin I honestly think in situations like these, it's more clear to name these models things like `Animal`, `Pet`, `Dog`, `Puppy` for example, since you can easily visualize the relationship between these models. The scenario names shouldn't include references to dogs etc, but I think it makes sense for the models to have a metaphorical relationship. What are your thoughts?

packages/cadl-ranch-specs/http/models/special-characters/main.cadl
```diff
@@ -0,0 +1,33 @@
...
+using Azure.Core.Foundations;
+
+@serviceTitle("SpecialCharacterClient")
+@serviceVersion("1.0.0")
+@server("http://localhost:3000", "TestServer endpoint")
+@doc("Illustrates models with special characters or keyword in property name")
+@route("/special-chars")
+namespace Model.SpecialCharacter;
+
+@doc("Base type with a discriminator containing dot.")
```
iscai-msft, line 15: I don't think `special-characters` should be tied to models. We can have a separate issue to add `special-characters` tests directly under `http`. This way we can test special characters for mdoels, operations, etc

iscai-msft (commented):
looking better!

packages/cadl-ranch-specs/http/models/inheritance-complex/main.cadl
timotheeguerin: yeah the models can have those names, just the scenarios names should be clear about whwat they do. The implementation details of a scenario(like model names) should be tailored to make the life easy for generator looking to implement the scenario

packages/cadl-ranch-specs/http/README.md
```diff
@@ -25,3 +25,5 @@ The tests in the http/models directory describe scenarios for generating models
 - [nested-models](https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/models/nested-models) - Generate, send, and receive input, output, and round-trip models with required nested model properties.
 - [readonly-properties](https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/models/readonly-properties) - Generate, send, and receive input, output, and round-trip models with readonly properties.
 - [inheritance](https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/models/inheritance) - Generate, send, and receive round-trip models polymorphic models.
+- [inheritance-complex](https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/models/inheritance-complex) - Generate, send, and receive complex polymorphic models.
```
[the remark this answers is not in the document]
changlong-liu: I moved them to the `inheritance`.
Since the tests in original `inheritance-complex` can fully cover the existing `inheritance`, I think it's fine to remove the existing model&operations in the heritance?

packages/cadl-ranch-specs/http/models/dictionary-basic/main.cadl
changlong-liu: moved dictionary test into `property-types`

packages/cadl-ranch-specs/http/models/special-characters/main.cadl
changlong-liu: removed `special-characters` from this PR

packages/cadl-ranch-specs/http/models/inheritance-complex/main.cadl
changlong-liu: renamed model&property names to make the model more easy to understand

packages/cadl-ranch-specs/http/models/inheritance/main.cadl
```diff
@@ -1,60 +1,112 @@
...
 import "@azure-tools/cadl-ranch-expect";
 
 using Cadl.Rest;
 using Cadl.Http;
 using Azure.Core.Foundations;
 
-@doc("Illustrates round-trip polymorphic models.")
-@scenarioService("/polymorphic")
-namespace BasicPolymorphicModels;
+@serviceTitle("ComplexInheritanceClient")
```
iscai-msft, line 8: You can do `@scenarioService("/models/inheritance")` and just have doc, scenarioService, and namespace

packages/cadl-ranch-specs/http/models/inheritance/main.cadl
```diff
@@ -1,60 +1,112 @@
...
+@route("/inheritance")
+namespace Model.Inheritance;
 
-@doc("Example base type.")
-model BaseClass {
-  @doc("An example property.")
-  baseClassProperty: string;
+@doc("Base model for non-discriminator three levels inheritance.")
+model Pet {
+  name?: string;
```
iscai-msft, line 17: I think `name` can be a required property

packages/cadl-ranch-specs/http/models/inheritance/main.cadl
```diff
@@ -1,60 +1,112 @@
...
 }
 
-model DerivedFromBaseClassA extends BaseClass {
-  @doc("An example property on a derived type")
-  derivedClassAProperty: string;
+@doc("The second level model A in the non-discriminator three levels inheritance.")
```
iscai-msft, line 20: Update doc, and I think `age` can just be required. To be honest most of these properties can be required unless you're specifically trying to test optionality

packages/cadl-ranch-specs/http/models/inheritance/main.cadl
```diff
@@ -1,60 +1,112 @@
...
+  sharktype: "goblin";
 }
 
 @scenario
-@scenarioDoc("Generate, send, and receive round-trip inherited model.")
-@route("/model")
+@scenarioDoc("Generate and send model. The valid input value is {name: \"abc\", age: 32, smart: true}")
+@route("/valid")
+@post
+op sendInheritanceValid(@body input: Siamese): OkResponse | ErrorResponse;
```
iscai-msft, line 64: nit: you don't need to repeat `inheritance` in the name, you can just say `putValid` and `getValid` etc

packages/cadl-ranch-specs/http/models/inheritance/main.cadl
iscai-msft: I'd also specify in the doc that this is normal multiple level inheritance, then specify in the doc for the polymorphic model that its for polymorphic inheritance

packages/cadl-ranch-specs/http/models/inheritance/main.cadl
```diff
@@ -1,60 +1,112 @@
...
-@scenarioDoc("Generate, send, and receive round-trip model with a polymorphic property.")
-@route("/property")
+@route("/invalid")
+@scenarioDoc("Get a basic model that is invalid for the local strong kind.")
+@get
+op getInheritanceInvalid(): Pet | ErrorResponse;
+
+@scenario
+@scenarioDoc("Generate, send, and receive round-trip model that optional property is not presented.")
+@route("/empty")
```
iscai-msft, line 86: I'm not convinced we need multiple dedicated tests for invalid and empty, @timotheeguerin thoughts?

packages/cadl-ranch-specs/http/models/inheritance/main.cadl
```diff
@@ -1,60 +1,112 @@
...
-op setValueWithPolymorphicProperty(
-  @body input: ModelWithPolymorphicProperty
-): ModelWithPolymorphicProperty | ErrorResponse;
+op setInheritanceEmpty(@body input: Siamese): Pet | ErrorResponse;
+
+@scenario
+@scenarioDoc("Get a basic model that optional property is sent as null.")
+@route("/null")
+@get
+op getInheritanceWithNullProperty(): Pet | ErrorResponse;
```
iscai-msft, line 94: what are we trying to test here? Sorry I'm just trying to figure out what scenario we're trying to test and if it's a good scenario to test

packages/cadl-ranch-specs/http/models/inheritance/main.cadl
```diff
@@ -1,60 +1,112 @@
...
+
+@scenario
+@route("/polymorphism/valid")
+@scenarioDoc("Generate, send, and receive round-trip models in three levels inheritance with 2 discriminators. The valid input is  { age: 1, kind: \"shark\",  sharktype: \"goblin\"}")
+@put
+op setPolymorphism(@body input: Fish): Fish | ErrorResponse;
```
iscai-msft, line 100: I think we can do `getPolymorphicModel` and `putPolymorphicModel`, one just returning the model and one putting it

packages/cadl-ranch-specs/http/models/inheritance/main.cadl
```diff
@@ -1,60 +1,112 @@
...
+
+@scenario
+@route("/polymorphism/recursive")
+@scenarioDoc("Generate, send, and receive round-trip models has collection and dictionary properties referring to models in three levels inheritance with 2 discriminator")
+@put
+op setPolymorphismRecursively(@body input: Salmon): Salmon | ErrorResponse;
```
iscai-msft, line 106: I think for polymorphic models we always use the base class, is that correct @timotheeguerin

packages/cadl-ranch-specs/http/models/inheritance/main.cadl
changlong-liu: modified

packages/cadl-ranch-specs/http/models/inheritance/main.cadl
changlong-liu: done

packages/cadl-ranch-specs/http/models/inheritance/main.cadl
changlong-liu: These abnormal cases are found in bodycomplex :), below are my understanding:
- The `invalid` is for testing what happen when service return an incorrect type of property value, for example: return an integer value for a string property.
- The `null` is for testing when service return a null property explicitly.
- The `empty` is for testing optional property is not sent. (The service returns null property implicitly)

Since they do not be related to inheritance and I find the `empty` has just been covered in `property-optional`, I remove them from this PR. And trying to create new PR to add `null` test in `property-optional`.

packages/cadl-ranch-specs/http/models/inheritance/main.cadl
changlong-liu: It's for testing what happen when service return an optional property as null.
Moved this test case to `property-optional``in this PR https://github.com/Azure/cadl-ranch/pull/92

packages/cadl-ranch-specs/http/models/inheritance/main.cadl
changlong-liu: done

packages/cadl-ranch-specs/http/models/inheritance/main.cadl
changlong-liu: Originally I use `Salmon` because the mockapi input/output an Salmon. But yes, base class `Fish` make more sense as an operation body.
Have changed to `Fish`.

packages/cadl-ranch-specs/http/models/property-types/main.cadl
```diff
@@ -146,3 +146,22 @@ model CollectionsModelProperty is ModelTemplate<InnerModel[]>;
 @route("/collections/model")
 interface CollectionsModel
   extends ModelOperations<CollectionsModelProperty, "[{'property': 'hello'}, {'property': 'world'}]"> {}
+
+// Test a model with a string dictionary property
```
timotheeguerin, line 150: I don't think this should be added here. Point of this spec is to make sure all the different types works as the type of a property.(check an `int` work, check a `dictionary` work, etc.) It shouldn't go in details. For that we should have a dedicated spec testing dictionaries

packages/cadl-ranch-specs/http/models/inheritance/main.cadl
```diff
@@ -1,60 +1,253 @@
...
 using Cadl.Rest;
 using Cadl.Http;
 using Azure.Core.Foundations;
 
-@doc("Illustrates round-trip polymorphic models.")
-@scenarioService("/polymorphic")
-namespace BasicPolymorphicModels;
+@doc("Illustrates inheritance and polymorphic model.")
+@scenarioService("/models/inheritance")
+namespace Models.Inheritance;
```
timotheeguerin, line 10: This is great name! specially as there is also discriminated union in cadl that we'll need to support which will difer from this but is also polymorhism

packages/cadl-ranch-specs/http/models/inheritance/main.cadl
```diff
@@ -1,60 +1,253 @@
...
+@scenarioDoc("""
+Generate and receive model.
+Expected response body:
+```json
+{"name": "abc", "age": 32, "smart": true}
+```
+""")
+@route("/valid")
+@get
+op getValid(): Siamese | ErrorResponse;
```
timotheeguerin, line 79: I would remove `| ErrorResponse` from all the operations. We should not care about errors in most specs.( if there is an error, it is unexpected and the test fail)

packages/cadl-ranch-specs/http/models/inheritance/main.cadl
```diff
@@ -1,60 +1,253 @@
...
+@route("/polymorphism/valid")
+@scenarioDoc("""
+Generate and receive polymorphic model in multiple levels inheritance with 2 discriminators.
+Expected response body:
+```json
+{"age": 1, "kind": "shark",  "sharktype": "goblin"}
+```
+""")
+@get
+op getPolymorphicModel(): Fish | ErrorResponse;
```
timotheeguerin, line 97: Should the `polymorphism` operations be moved inside a subnamespace called `Polymorphism` or maybe `Discriminated` and change the route as well

packages/cadl-ranch-specs/http/models/inheritance/main.cadl
```diff
@@ -1,60 +1,253 @@
...
-): ModelWithPolymorphicProperty | ErrorResponse;
+op putRecursivePolymorphicModel(@body input: Fish): OkResponse | ErrorResponse;
+
+@scenario
+@route("/polymorphism/missingdiscriminator")
+@scenarioDoc("""
+Get a model omitting the discriminator.
+""")
+@get
+op getPolymorphismMissingDiscriminator(): Fish | ErrorResponse;
```
timotheeguerin, line 253: Should we have one where the discriminator is unknown as well?

timotheeguerin (commented):
I think the inheritance spec looks pretty good overall, just a few organizing issues. 

For the null case, I think we should maybe have a dedicated tests for passing null somewhere else because it isn't tied to optional properties.

packages/cadl-ranch-specs/http/models/inheritance/main.cadl
iscai-msft: yeah i like the idea of a `polymorphism` subnamespace

packages/cadl-ranch-specs/http/models/inheritance/main.cadl
iscai-msft: oh good idea!

packages/cadl-ranch-specs/http/models/inheritance/main.cadl
timotheeguerin: I think `Discriminated` make more sense as technically with inheritance in nominal language it is always polymorphism

packages/cadl-ranch-specs/http/models/inheritance/main.cadl
changlong-liu: removed all `| ErrorResponse` since there is no error tests

packages/cadl-ranch-specs/http/models/inheritance/main.cadl
changlong-liu: moved polymorphic tests into interface `Discriminated` and routes are updated as well

packages/cadl-ranch-specs/http/models/inheritance/main.cadl
changlong-liu: yes, added a `getWrongDiscriminator` :)

packages/cadl-ranch-specs/http/models/property-types/main.cadl
changlong-liu: Hi @timotheeguerin , cadl `Record<>` type is not covered in existing `property-types/main.cadl`, so I'm trying to add a basic test for it. 

Just want to check `dictionary` can work.
Is it valuable to keep at least one `Record<>` test in this `property-types`?

timotheeguerin: Yeah I think is is valuable to add a single test here to make sure properties and dictionaries work together.

timotheeguerin: and then in `dictionaries/main.cadl` we can have the more in depth test for dictionaries whatever they are

packages/cadl-ranch-specs/http/dictionary/main.cadl
```diff
@@ -0,0 +1,99 @@
...
+    {
+      doc: TDoc,
+    }
+  )
+  @put
+  put(@body body: TModel): OkResponse;
+}
+
+// Test a model with an int dictionary property
+@doc("Model with dictionary int properties")
```
iscai-msft, line 50: @timotheeguerin correct me if I'm wrong, but I think for the dictionary tests, we really want to be focusing on testing dictionaries themselves, not dictionaries as properties. We can add dictionary property tests to the `property-types` test, but for the dictionary test we don't want them all to be tied to models

packages/cadl-ranch-specs/http/models/inheritance/main.cadl
```diff
@@ -4,53 +4,268 @@ import "@azure-tools/cadl-ranch-expect";
...
+        ]
+      }
+    }
+  }
+  ```
+  """)
+  @put
+  putRecursiveModel(@body input: Fish): OkResponse;
+
+  @scenario
```
iscai-msft, line 248: inheritance looks good to me!

packages/cadl-ranch-specs/http/dictionary/main.cadl
```diff
@@ -0,0 +1,99 @@
+import "@cadl-lang/rest";
+import "@azure-tools/cadl-ranch-expect";
+
+@doc("Illustrates various of dictionary.")
+@scenarioService("/dictionary")
+namespace Dictionary;
+using Cadl.Http;
```
timotheeguerin, line 7: nit: would move using above the file namespace.

packages/cadl-ranch-specs/http/dictionary/main.cadl
```diff
@@ -0,0 +1,99 @@
...
+@scenarioDoc("""
+Dictionary as output in body.
+Expected output body:
+```json
+{ "k1": "hello", k2: "world" }
+```
+""")
+@route("/body")
+@get
+op getBody(): {
```
timotheeguerin, line 84: The other test are great but I think we just don't need to test this as a property but for every test we can just get or put the Record directly. (No need for this explicit get/putBody test as every test should already be using the record at the root of the body)

packages/cadl-ranch-specs/http/dictionary/main.cadl
```diff
@@ -0,0 +1,99 @@
...
+
+// TEMPLATES
+@doc("Template type for testing dictionary with specific value type. Pass in the type of the value you are looking for")
+model ModelValueTemplate<TProperty> {
+  @doc("Property")
+  property: Record<TProperty>;
+}
+
+@doc("Template to have models operations")
+interface ModelOperations<TModel, TDoc extends string> {
```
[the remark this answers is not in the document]
changlong-liu: yes, changed the naming in new PR

changlong-liu:
Hi @timotheeguerin , @iscai-msft , I moved the dictionary changes into PR https://github.com/Azure/cadl-ranch/pull/104 and leave this PR only for inheritance :)

packages/cadl-ranch-specs/http/dictionary/main.cadl
changlong-liu: yes, I put dictionary as root type in new PR

packages/cadl-ranch-specs/http/dictionary/main.cadl
```diff
@@ -0,0 +1,99 @@
...
+using Cadl.Http;
+
+// TEMPLATES
+@doc("Template type for testing dictionary with specific value type. Pass in the type of the value you are looking for")
+model ModelValueTemplate<TProperty> {
+  @doc("Property")
+  property: Record<TProperty>;
+}
+
+@doc("Template to have models operations")
```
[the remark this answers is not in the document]
changlong-liu: learnt from the `property-types` :)

timotheeguerin (approved):
sweet! looks great!

PR #78, opened 2022-08-24
fix extensible enums mock api tests

iscai-msft:


timotheeguerin (approved):
My bad on writting those... not a great job :)
Wondering if the validate-mockapi should also validate the right endpoints are defined

iscai-msft:
@timotheeguerin yeah I was wondering too if it should've caught it. Is there a way to catch verb and endpoint once the scenario names are linked?

PR #79, opened 2022-08-25
New @scenarioService to reduce dup

timotheeguerin:


iscai-msft (approved):
thanks @timotheeguerin this looks so good!

Issue #51, opened 2022-08-16
Serve should fail when different scenarios have same uri


pshao25:
We could start `cadl-ranch serve ./http` when different scenrios have same uri now. e.g., 
https://github.com/Azure/cadl-ranch/blob/c4f41f483eac812527f7b6dc837bd22d255a18ed/packages/cadl-ranch-specs/http/resiliency/srv-driven-1/mockapi.ts#L16
https://github.com/Azure/cadl-ranch/blob/c4f41f483eac812527f7b6dc837bd22d255a18ed/packages/cadl-ranch-specs/http/resiliency/srv-driven-2/mockapi.ts#L16

I think we should fail this.

haolingdong-msft:
There is also the same problem with tests under /models folder,all operations under models/ folder defined route as "/models", e.g. https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/models/collections-basic/main.cadl#L47, https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/models/collections-models/main.cadl#L35,  this can also lead to all the operations have the same url and have problem in finding the corresponding operation.

weidongxu-microsoft:
For cadl under "models" folder, should we just use different routes in different test scenarios?

timotheeguerin:
Yeah I agree, there shouldn't be any duplicated endpoint. The rule right now is 1 scenario to 1 or more endpoints(for things like paging lro). But either way each endpoint should be completely unique. I'll add some validation to the mock server.

pshao25:
What's more, we also need to fail when two scenarios have the same name.

timotheeguerin:
Agree on scenario name duplicate as well. Have a PR adding validation and doing a rough fix to make it pass https://github.com/Azure/cadl-ranch/pull/58

PR #87, opened 2022-08-30
Add validation for duplicate Uris

pshao25:
This is what I want to fail as mentioned in https://github.com/Azure/cadl-ranch/issues/51
image

### DIFF
--- a/packages/cadl-ranch/src/actions/validate-mock-apis.ts
+++ b/packages/cadl-ranch/src/actions/validate-mock-apis.ts
@@ -15,6 +15,19 @@ export async function validateMockApis({ scenariosPath }: ValidateMockApisConfig
   const cadlCompiler = await importCadl(scenariosPath);
   const cadlRanchExpect = await importCadlRanchExpect(scenariosPath);
   const diagnostics = createDiagnosticReporter();
+
+  const foundUris = new Set<string>();
+  mockApis.forEach(mockApi =>
+    Object.values(mockApi.scenarios).forEach(scenario =>
+      scenario.apis.forEach(api => {
+        const key = `${api.method} ${api.uri}`;
+        foundUris.has(key)
+          ? diagnostics.reportDiagnostic({
+            message: `Uri ${api.uri} with HTTP verb ${api.method} is being defined twice.`,
+          })
+          : foundUris.add(key);
+      })));
+
   for (const { name, cadlFilePath } of scenarioFiles) {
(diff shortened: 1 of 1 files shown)


packages/cadl-ranch/src/actions/validate-mock-apis.ts
```diff
@@ -15,6 +15,19 @@ export async function validateMockApis({ scenariosPath }: ValidateMockApisConfig
...
   const diagnostics = createDiagnosticReporter();
+
+  const foundUris = new Set<string>();
+  mockApis.forEach(mockApi =>
+    Object.values(mockApi.scenarios).forEach(scenario =>
+      scenario.apis.forEach(api => {
+        const key = `${api.method} ${api.uri}`;
+        foundUris.has(key)
+          ? diagnostics.reportDiagnostic({
+            message: `Uri ${api.uri} with HTTP verb ${api.method} is being defined twice.`,
```
timotheeguerin, line 26: So there is already validation to make sure 2 scenario don't define the same URL here https://github.com/Azure/cadl-ranch/blob/7435655e82b732b3573379f76e0e613423d244e9/packages/cadl-ranch/src/scenarios-resolver.ts#L83

Making sure 2 endpoints are not the same could be useful too but then I would update this message to be 
`... is being defined twice in Mock Apis.`

pshao25: If I understand correctly, this snippet is to validate the cadl file, not the mockapi file. That is why validate-scenarios could stilll pass even there exists duplicate endpoint.

I hope the mockapi file, which is used when we start cadl-ranch server, could fail when it defines duplicate endpoints.

It would be even better if we could check the consistency between cadl file and mockapi file.

timotheeguerin: consistency check between the scenario and implemention would  be nice but it is a little tricky as the mockapi endpoints might not exactly match the cadl (could be a parameter in the cadl but implemented as multiple endpoints in the mockapi)

pshao25: Understand. So we still might need this PR to validate the mockapi, so that next time others and I won't take so much time figure out why the ranch server cannot recognize the request.

lmazuel:
@pshao25 closing to clean open PRs, feel free to re-open and conclude if you still want it

PR #91, opened 2022-09-05
Fix issues in property-types

pshao25:
`datetime` conparison changes to `toISOString` comparison.


packages/cadl-ranch-specs/http/models/property-types/mockapi.ts
```diff
@@ -63,7 +69,7 @@ const enumMock = createMockApis("enum", "ValueOne");
 Scenarios.Models_Property_Types_Enum_get = passOnSuccess(enumMock.get);
 Scenarios.Models_Property_Types_Enum_put = passOnSuccess(enumMock.put);
 
-const extensibleEnumMock = createMockApis("extensible-enum", "UnknownValue");
+const extensibleEnumMock = createMockApis("extensible-enum", "ValueOne");
```
weidongxu-microsoft, line 72: extensible-enum means client/service can use any value (I assume the "UnknownValue" is for this case).

packages/cadl-ranch-specs/http/models/property-types/main.cadl
```diff
@@ -6,6 +6,8 @@ using Cadl.Rest;
 using Cadl.Http;
 using Azure.Core.Foundations;
 
+@Cadl.Rest.produces("application/json")
+@Cadl.Rest.consumes("application/json")
```
weidongxu-microsoft, line 10: I assume they are going to be deprecated anyway?

pshao25: It would be a long way I think. Currently, we .net still use these decorators.

packages/cadl-ranch-specs/http/models/property-types/mockapi.ts
pshao25: [ExtensibleEnumProperty](https://github.com/Azure/cadl-ranch/blob/9e5ddbeda8413655341e92392ea288ccc67f2064/packages/cadl-ranch-specs/http/models/property-types/main.cadl#L116) has [InnerExtensibleEnum](https://github.com/Azure/cadl-ranch/blob/9e5ddbeda8413655341e92392ea288ccc67f2064/packages/cadl-ranch-specs/http/models/property-types/main.cadl#L112-L113) inside whose allowed values are only ValueOne and ValueTwo. I'm not aware we could accept any values. @archerzz do you know?

packages/cadl-ranch-specs/http/models/property-types/mockapi.ts
```diff
@@ -23,6 +23,12 @@ function createMockApis(route: string, value: any): MockApiGetPut {
       };
     }),
     put: mockapi.put(url, (req) => {
+      if (route == "datetime" && new Date(body.property).toISOString() === new Date(req.body.property).toISOString()) {
```
lirenhe, line 26: use ===

packages/cadl-ranch-specs/http/models/property-types/main.cadl
timotheeguerin: We shouldn’t add this. As @weidongxu-microsoft  it will most likely get deprecated on the next design meeting and even if it wasn’t the default is assumed to be json. So either way the csharp generator shouldn’t be needing it to function.

pshao25: We really need this. Currenly we need these to be the default value of content-type and accept, otherwise no headers will be generated. 

cc @lirenhe the information seems not aligned with ours. We cannot generate workable SDKs without these two lines.

timotheeguerin: this is a problem with the csharp generator... 

This is a spec for what the generator should do. **It is valid.** It is the entire point of this repo, defining the behavior the generators should be respecting.

The generator should be fixed.

packages/cadl-ranch-specs/http/models/property-types/mockapi.ts
```diff
@@ -23,6 +23,12 @@ function createMockApis(route: string, value: any): MockApiGetPut {
       };
     }),
     put: mockapi.put(url, (req) => {
+      if (route === "datetime" && new Date(body.property).toISOString() === new Date(req.body.property).toISOString()) {
+        return {
+          status: 204,
+        };
+      }
+
       req.expect.bodyEquals(body);
```
haolingdong-msft, line 32: If we do the comparison in using ISOString, do we still need this `req.expect.bodyEquals(body)`?

timotheeguerin: I think in that case it would already have returned before(line 28). Think for a better design we could have an additional argument to `createMockApis` called `convertBody` and it would call that when comparing the body instead

packages/cadl-ranch-specs/http/models/property-types/main.cadl
pshao25: Removed.

timotheeguerin (approved):

PR #92, opened 2022-09-08
add getNull testcase for optional property

changlong-liu:

### DIFF
--- a/packages/cadl-ranch-specs/http/models/property-optional/mockapi.ts
+++ b/packages/cadl-ranch-specs/http/models/property-optional/mockapi.ts
@@ -5,6 +5,7 @@ export const Scenarios: Record<string, ScenarioMockApi> = {};
 interface MockApiGetPut {
   getAll: MockApi;
   getDefault: MockApi;
+  getNull: MockApi;
   putAll: MockApi;
   putDefault: MockApi;
 }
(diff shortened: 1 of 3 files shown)


packages/cadl-ranch-specs/http/models/property-optional/mockapi.ts
```diff
@@ -55,36 +65,42 @@ function createMockApis(route: string, value: any): MockApiGetPut {
 const stringMock = createMockApis("string", "hello");
 Scenarios.Models_Property_Optional_String_getAll = passOnSuccess(stringMock.getAll);
 Scenarios.Models_Property_Optional_String_getDefault = passOnSuccess(stringMock.getDefault);
+Scenarios.Models_Property_Optional_String_getNull = passOnSuccess(stringMock.getNull);
```
timotheeguerin, line 68: `null` is not the same as optional optional, this is technically an unexpected value. what are we trying to test here. 

This is how you would represent nullable in cadl(which will also be discouraged as optional is really what you want).
```cadl
model Foo {
  isNullable: string | null;
}
```

changlong-liu: Thanks @timotheeguerin for the detail knowledge!  I see so we don't need this for `property-optional`.

Though the `string|null` is discouraged, shall we add a case for nullable test in `property-types`?

PR #105, opened 2022-09-15
Basic LRO polling scenario

msyyc:


packages/cadl-ranch-specs/http/lro/lro-basic/main.cadl
```diff
@@ -0,0 +1,27 @@
...
+import "@azure-tools/cadl-azure-core";
+
+using Cadl.Http;
+using Azure.Core;
+
+@scenarioService("/lro/basic")
+@doc("Basic test for lro")
+namespace Lro.Basic;
+
+#suppress "@azure-tools/cadl-azure-core/use-standard-operations" "For testing"
```
msyyc, line 12: Hi @timotheeguerin I can pass the [python test](https://github.com/Azure/autorest.python/pull/1442/files#diff-ac511dcfc320ffecb1bc05f7b575da71c4b5b87f78c75ea9a7ae2b7284a0d848) in local environment with the PR but CI shows that I don't implement it: 
image
I don't find root cause. Is it related with the `suppress`?

timotheeguerin: shouldn't be the suppress just ignore the cadl warning

timotheeguerin: the names are different though. Here it is `Lro_Basic_create` and `Lro_Basic_read` but for the mock scenario you have 3 `Lro_Basic_create_init`, `Lro_Basic_create_polling`, `Lro_Basic_create_final_result`

What is the goal of this test, this would a great case in my opinion where a scenario doesn't match to an operation but to an interface or namespace as what you want it to have multilple endpoint being called

msyyc: This test is to simulate the LRO interaction:
image

msyyc: Hi @timotheeguerin The CI passes. Could you help merge the PR? We need a new release version for https://github.com/Azure/autorest.python/pull/1442

timotheeguerin: this should be a single scenario. The scenario is the end-to-end LRO. 
If you put `@scenario` on an interface or namespace it means all oeration under belong to that scenario. Then `passOnSuccess` can take a list of endpoints that all needs to be called.

packages/cadl-ranch-specs/http/lro/lro-basic/main.cadl
```diff
@@ -0,0 +1,27 @@
+import "@cadl-lang/rest";
+import "@azure-tools/cadl-ranch-expect";
+import "@azure-tools/cadl-azure-core";
+
+using Cadl.Http;
+using Azure.Core;
+
+@scenarioService("/lro/basic")
+@doc("Basic test for lro")
+namespace Lro.Basic;
```
timotheeguerin, line 10: @iscai-msft this being an Azure.Core functionaltiy do you think should we scope those scenarios under azure?

iscai-msft: I'm not actually sure. For `azure` are we referring to management plane or just azure specific tests. I think its the latter, in which case that makes sense to me

timotheeguerin: yeah AZure.core is technically the base for all azure(ARM is yet another layer on top of it)

packages/cadl-ranch-specs/http/lro/lro-basic/main.cadl
timotheeguerin: is the read not part of the same scenario? Isn't the scenario for LRO
- create
- poll
- read

```cadl
@scenario // this makes it that the `Basic` namespace is the scenario instead of having 1 per operation.
namespace Basic {
 op create() ...
op read()...
}
```

msyyc: @timotheeguerin  Yeah, create/poll/read shall be in the same scenario. This also reminds me: now that LRO contains three part, there is no need to split into read(). I update the code with only one operation and the test all the three endpoint in `Scenarios.Lro_Basic_create`.

timotheeguerin: well I think it is kinda cheating and not really representative, you won't poll on a put (what are you even going to be sending)
I think it should still have the create, polling and reading endpoint but just add a namespace grouping them together to make it a single scenario.

Not the most familiar with LRO but I think we should also have the linking between the different operations like it is done in the `Azure.Core` library
https://github.com/Azure/cadl-azure/blob/8f23398e8c04131b609c7cdd78b465234e08c91b/packages/cadl-azure-core/lib/foundations.cadl#L51-L63
Wondering if maybe we shouldn't just use the resource operation themself https://github.com/Azure/cadl-azure/blob/c0add931f6cc26dadcedb33d9c87c4319bb9b2e7/packages/cadl-azure-core/README.md#define-resource-operations
(But maybe not as this is just a layer on top)

timotheeguerin: Thinking more about it we should have the following scenario for Lro:
- `Azure.Lro.Polling` : Test a Lro operation that will need polling.
- `Azure.Lro.Immediate`: Test a Lro operation that returns the response immediately(Doesn't need for polling).

msyyc: (1)`I think it should still have the create, polling and reading endpoint but just add a namespace grouping them together to make it a single scenario.`
- That makes sense and will do it.

(2) `Not the most familiar with LRO but I think we should also have the linking between the different operations like it is done in the Azure.Core library
https://github.com/Azure/cadl-azure/blob/8f23398e8c04131b609c7cdd78b465234e08c91b/packages/cadl-azure-core/lib/foundations.cadl#L51-L63
Wondering if maybe we shouldn't just use the resource operation themself https://github.com/Azure/cadl-azure/blob/c0add931f6cc26dadcedb33d9c87c4319bb9b2e7/packages/cadl-azure-core/README.md#define-resource-operations
(But maybe not as this is just a layer on top)`
- I just want to test the scenario equals to `x-ms-long-running: true` which is asked in https://github.com/Azure/autorest.python/issues/1435. My intention is not test all the scenario for LRO. That's why I call it "lro.basic"
(3) 

> Thinking more about it we should have the following scenario for Lro:
> 
> * `Azure.Lro.Polling` : Test a Lro operation that will need polling.
> * `Azure.Lro.Immediate`: Test a Lro operation that returns the response immediately(Doesn't need for polling).
> 
> Both those scenarios probably have the exact same definition but the implementation is where it differ
- That makes sense and I will do it.

Will update the code soon

timotheeguerin: Ok yeah that make sense, if you want to just do the basic case for now just do the namepsace containing the operations and make sure the scenario name is `Azure.Lro.Polling.Success`
and leave a comment about future potential scenarios that should be here and we can do that later

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -311,6 +311,15 @@ This test is testing this payload is returned from the server
...
 ```
 
+### Azure_Lro_PollingSuccess
+
+- Endpoints:
+  - `put /lro/basic/put`
+  - `put /lro/basic/put/polling`
+  - `put /lro/basic/put`
+
+undefined
```
msyyc, line 321: @timotheeguerin  it seems that `pnpm generate-scenarios-summary` does not work well when we group several endpoints into one scenario

packages/cadl-ranch-specs/http/lro/lro-basic/main.cadl
msyyc: @timotheeguerin updated and please review

packages/cadl-ranch-specs/cadl-ranch-summary.md
timotheeguerin: hhm, I'll look into it, lets not block merging on this unless I fix it before

timotheeguerin (approved):
Looks good!

PR #125, opened 2022-09-28
Redresign usage scenarios

timotheeguerin:


iscai-msft (approved):
amazing!

packages/cadl-ranch-specs/http/models/input-basic/main.cadl
```diff
@@ -1,23 +0,0 @@
-import "@cadl-lang/rest";
-import "@azure-tools/cadl-ranch-expect";
-
```
MaryGao, line 3: @timotheeguerin I'd like to confirm with you you delete test cases including input-basic, nested-models, output-basic and roundtrip-basic. Could you share some reasons here?

timotheeguerin: Those scenarios had no mock api and didn’t make sense to have 3 different file for testing the same thing so I consolidated that in models usage

PR #132, opened 2022-10-14
fix dictionary mockapi

changlong-liu:
1. fix response body
2. fix response status code

### DIFF
--- a/packages/cadl-ranch-specs/http/dictionary/mockapi.ts
+++ b/packages/cadl-ranch-specs/http/dictionary/mockapi.ts
@@ -24,7 +24,7 @@ function createModelMockApis(route: string, value: any): MockApiGetPut {
     put: mockapi.put(url, (req) => {
       req.expect.bodyEquals(value);
       return {
-        status: 204,
+        status: 200,
       };
     }),
   };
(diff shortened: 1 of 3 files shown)


packages/cadl-ranch-specs/http/dictionary/mockapi.ts
```diff
@@ -14,18 +14,17 @@ interface MockApiGetPut {
...
-        body: json(body),
+        body: json(value),
       };
     }),
     put: mockapi.put(url, (req) => {
-      req.expect.bodyEquals(body);
+      req.expect.bodyEquals(value);
       return {
-        status: 204,
+        status: 200,
```
timotheeguerin, line 27: Should we change the spec to return `void ` instead of `OkResponse` would be more consistent with the others

timotheeguerin (approved):

packages/cadl-ranch-specs/http/dictionary/mockapi.ts
changlong-liu: changed to use `void` and resumed to `status: 204`

timotheeguerin (approved):

PR #150, opened 2022-10-26
Rename namespace array to arrays

pshao25:
We .net don't allow namespace contains "array", which will lead to code compilation failure.


changlong-liu:
Hi @iscai-msft , @timotheeguerin , would you help review this?

Python has similar problem with .Net that the `array` is an invalid namespace.

iscai-msft (approved):

PR #153, opened 2022-10-28
fix mockapi of getting datetime

changlong-liu:
### For fixing below problem:
Currently, the result of `get` operation on datetime in packages `property-optional` and `property-types` varies depending on the sequence of the executing of testcases:
+ If the `put datetime` is executed before `get datetime`, then the result of the `get` is "2022-08-26T18:38:00.000Z";
+ If the `get datetime` invoked ahead of `put datetime`, then the result of the `get` is "2022-08-26T18:38:00Z".

### Fix
This PR make the result of the `get` solid to be "2022-08-26T18:38:00.000Z".

### DIFF
--- a/packages/cadl-ranch-specs/http/models/property-optional/mockapi.ts
+++ b/packages/cadl-ranch-specs/http/models/property-optional/mockapi.ts
@@ -20,6 +20,9 @@ function createMockApis(route: string, value: any, convertBodyProperty?: (_: any
   const defaultUrl = `${url}/default`;
   const allBody = { property: value };
   const defaultBody = {};
+  if (convertBodyProperty && allBody.property) {
+    allBody.property = convertBodyProperty(allBody.property);
+  }
   const getAll = mockapi.get(allUrl, (req) => {
     return {
       status: 200,
@@ -33,9 +36,8 @@ function createMockApis(route: string, value: any, convertBodyProperty?: (_: any
     };
   });
   const putAll = mockapi.put(allUrl, (req) => {
-    if (convertBodyProperty && req.originalRequest.body?.property && allBody?.property) {
+    if (convertBodyProperty && req.originalRequest.body?.propertyy) {
       req.originalRequest.body.property = convertBodyProperty(req.originalRequest.body.property);
-      allBody.property = convertBodyProperty(allBody.property);
     }
 
     req.expect.bodyEquals(allBody);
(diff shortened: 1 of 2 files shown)


packages/cadl-ranch-specs/http/models/property-optional/mockapi.ts
```diff
@@ -33,9 +36,8 @@ function createMockApis(route: string, value: any, convertBodyProperty?: (_: any
     };
   });
   const putAll = mockapi.put(allUrl, (req) => {
-    if (convertBodyProperty && req.originalRequest.body?.property && allBody?.property) {
+    if (convertBodyProperty && req.originalRequest.body?.propertyy) {
       req.originalRequest.body.property = convertBodyProperty(req.originalRequest.body.property);
-      allBody.property = convertBodyProperty(allBody.property);
     }
 
     req.expect.bodyEquals(allBody);
```
pshao25, line 43: I think we should
- If there is a convertBodyProperty, create a new local allBody with converted body and `req.expect.bodyEquals(localAllBody);`
- Otherwise still `req.expect.bodyEquals(allBody);`

MaryGao:
Thanks changlong for the fix we also met the issue.

packages/cadl-ranch-specs/http/models/property-optional/mockapi.ts
changlong-liu: Hi @pshao25, 

I added code here: https://github.com/Azure/cadl-ranch/pull/153/files#diff-7cedfaa56d319065d6385f01969648a56e8d91556bfa0b9018a99ad9b0e4a2bdR23 ,
It can achieve same goal as your description?

pshao25: No, I hope the expected value `allBody` not impacted, so that e.g. when we call `get`, we could still get the unmodified value.

changlong-liu: I'm not sure this is the right direction.
It will make the data value in wire of `put` and `get` be different for `datetime` while other datetypes have exactly the same values between these two operations.

changlong-liu: Hi @pshao25 , I updated as offline discussion.

iscai-msft: I’m going to go ahead and merge thanks changlong

PR #159, opened 2022-11-07
Remove put operation in arrays

pshao25:
Array as put body is invalid, we will add such linter https://github.com/Azure/cadl-azure/issues/2194. So remove it from cadl-ranch.


timotheeguerin (changes requested):
This is again a case of something that is specific to azure. This is testing the language basics of cadl.
If `put` is really a bad choice we can change to `get` but apart from having a linter rule validating that it should still work.

lmazuel:
Agree with Tim, conceptually this is not wrong, I wouldn't remove it

PR #163, opened 2022-11-11
Automatic Visibility scenarios(read, create, update, etc.)

changlong-liu:


packages/cadl-ranch-specs/http/models/readonly-properties/main.cadl
```diff
@@ -4,7 +4,7 @@ import "@azure-tools/cadl-ranch-expect";
 using Cadl.Http;
 
 @doc("Illustrates models with readonly properties.")
-@scenarioService("/readonly-properties")
+@scenarioService("/models/readonly-properties")
 namespace ReadonlyProperties;
```
timotheeguerin, line 8:
```suggestion
namespace Models.ReadonlyProperties;
```

packages/cadl-ranch-specs/http/models/readonly-properties/main.cadl
```diff
@@ -65,13 +65,60 @@ model RoundTripModel {
...
+    requiredString: "**",
+  },
+  requiredReadonlyStringList: ["foo", "bar"],
+  requiredReadonlyIntList: [1, 2],
+  optionalReadonlyStringList: ["fox", "fish"],
+  optionalReadonlyIntList: [3, 4],
+}
+```
+""")
 @route("/models")
```
timotheeguerin, line 90: Do we need this extra route?

timotheeguerin (changes requested):
Wondering if we should not step back from this spec. There is a few issues:
- scenario names don't make sense right now. what are they really testing
- this spec is specific to visibility read. We should really have a visibility spec instead that cover(read, create, etc.). Readonly not being that special of a case(It is in openapi but not in cadl)

changlong-liu:
> Wondering if we should not step back from this spec. There is a few issues:
> 
> * scenario names don't make sense right now. what are they really testing
> * this spec is specific to visibility read. We should really have a visibility spec instead that cover(read, create, etc.). Readonly not being that special of a case(It is in openapi but not in cadl)

Hi @timotheeguerin , thanks for taking review!
Agree that we can have a test spec for visibility which includes `visibility("read")` as part of it.  I updated to change the spec to be a test of visibility of read/write/create/update.

I'm not sure my understanding of the create/update visibility are right, would you take a look at the mockapi implementation?
https://github.com/Azure/cadl-ranch/pull/163/files#diff-f1f73eea37f49652a997a7fa21f7d524b9a83a5f4f324fe94b3bc7cf2a2be2b5R31-R50
(here I assume visibility("create") property should be used in the request of PUT&POST)

packages/cadl-ranch-specs/http/models/readonly-properties/main.cadl
changlong-liu: removed extra route

packages/cadl-ranch-specs/http/models/readonly-properties/main.cadl
changlong-liu: named to "Models.Visibility"

packages/cadl-ranch-specs/http/models/visibility/main.cadl
```diff
@@ -0,0 +1,83 @@
+import "@cadl-lang/rest";
```
timotheeguerin, line 1: Think it looks good. A good way to check you have the visibility setup correctly is to write the code with the openapi3 emitter(or in the cadl-playground)

https://cadlplayground.z22.web.core.windows.net/?c=aW1wb3J0ICJAY2FkbC1sYW5nL3Jlc3QiOwoKdXNpbmcgQ2FkbC5IdHRwOwoKCkBkb2MoIk91dHB1dCBtb2RlbCB3aXRoIHZpc2liaWxpdHkgcHJvcGVydGllcy4iKQrGJFbJH03FEHsKICDGS1JlcXVpcmVkIHN0cmluZywgaWxsdXN0cmF0xHxhIHJlYWRvbmx5IHZhbHVlIHR5cGXIYnnEYCAgQMp8KCLEL8UWcsdYUsdAU8VfOsdnOwryAIFpbnTRfndyaXRl33%2FMf8Uw7QCAV8hCSW50OsRkMzLUfWNyZWHHauYAoiBjb2xsZWN0aW9u02jGL81pQ8lC5gDrTGlzdOgA71td1HR1cGTIdGludN5xxizNcVXJP0ludMZu5QDfxG19CgpAZ2V0Cm9wIGdldOUB9CgpOvACBzsKCkBwdcUmcHXHJkBib2TkAJBwdXTRMCk6IHZvaWTFN2F0Y2jFOcQJ3zvPO29zxnNvc990zTk%3D

packages/cadl-ranch-specs/http/models/visibility/main.cadl
```diff
@@ -0,0 +1,83 @@
...
+namespace Models.Visibility;
+
+@doc("Output model with visibility properties.")
+model VisibilityModel {
+  @doc("Required string, illustrating a readonly value type property.")
+  @visibility("read")
+  requiredReadonlyString: string;
+
+  @doc("Required int, illustrating a writeonly value type property.")
+  @visibility("write")
```
timotheeguerin, line 17: Think we removed `write` from visiblity so would remove that property

packages/cadl-ranch-specs/http/models/visibility/main.cadl
timotheeguerin: There is some docs on the automatic visibility https://microsoft.github.io/cadl/standard-library/rest/operations#automatic-visibility

Would also maybe name the namespace `Models.Visibility.Automatic`

timotheeguerin: data looks great now, think just can play a bit with the names to have the scenarios be better

packages/cadl-ranch-specs/http/models/visibility/main.cadl
```diff
@@ -0,0 +1,120 @@
+
+import "@cadl-lang/rest";
+import "@azure-tools/cadl-ranch-expect";
+
+
+using Cadl.Http;
+
+@doc("Illustrates models with visibility properties.")
+@scenarioService("/models/visibility")
+namespace Models.Visibility;
```
timotheeguerin, line 10:
```suggestion
namespace Models.Visibility.Automatic;
```

timotheeguerin (approved):
@changlong-liu thanks for this, think it is looks great apart from maybe a few naming changes

packages/cadl-ranch-specs/http/models/visibility/main.cadl
```diff
@@ -0,0 +1,120 @@
...
+
+@doc("Output model with visibility properties.")
+model VisibilityModel {
+  @doc("Required string, illustrating a readonly property.")
+  @visibility("read")
+  requiredReadonlyString: string;
```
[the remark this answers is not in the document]
changlong-liu: yes @timotheeguerin , the new names make sense. I updated the PR.

packages/cadl-ranch-specs/http/models/visibility/mockapi.ts
```diff
@@ -0,0 +1,66 @@
...
+  };
+  for (const k of keys) {
+    if (k in fullData) {
+      ret[k] = fullData[k];
+    }
+  }
+  return ret;
+}
+
+Scenarios.Models_Visibility_getModel = passOnSuccess(
```
timotheeguerin, line 24: Need to update the mock apis as well to include `_Automatic` in scenario name

changlong-liu: my bad :( for even missing the pipeline errors. 
I updated to get fixed.

weidongxu-microsoft:
Does python generate all these VisibilityModelCreate / VisibilityModelCreateOrUpdate / VisibilityModelDelete / VisibilityModelQuery ?
What about round-trip model?

https://cadlplayground.z22.web.core.windows.net/cadl-azure/?c=aW1wb3J0ICJAY2FkbC1sYW5nL3Jlc3QiOwoKdXNpbmcgQ2FkbC5IdHRwOwoKQGRvYygiSWxsdXN0cmF0ZXMgbW9kZWxzIHdpdGggdmlzaWJpbGl0eSBwcm9wZXJ0aWVzLiIpCkBzZXJ2aWNlCm5hbWVzcGFjZSBNxTguVskzLkF1dG9tYXRpY8loT3V0cHV0xmPfYsYkyk7FXyB7CiAgxktSZXF1aXJlZCBzdHJpbmcsIGnoAMTkAONhIHJlYWRvbmzJV3nEVSAgQMpxKCLEJMUWxAlQcm9wOsdOOwrSaGludDMy0WdxdWVy3WTFIcUXxQrGZsVM1GXGf1td0WhjcmVhdGXcacYixRjGC%2BwA0Vtd1G7lAIfTbXVwZN9txiLFGMYL6wDY1mxib29s0WlkZWxl3mnGIsUYxgvGacROZWFuOwp9CgpAZ2V0Cm9wIGdldOUCJShAYm9keSBpbnB1dDrwAkIp0RLkAp1oZWFkxEPECN9ExkRPa1Jlc3BvbnNlxD9wdeUAgXB1%2FwCBxz12b2lkxTdhdGNoxTnECd87zztvc8Zzb3PfdM855gFFxDvGCt89yz0%3D

Issue #158, opened 2022-11-03
Update to tests for extensible enums required


tjprescott:
At a recent Cadl design meeting, we approved a design to simplify the way extensible enums work. This will require changing the tests in cadl-rach to force DPG emitters to update their behavior:

Here's the underlying Cadl issue:
https://github.com/Azure/cadl-azure/issues/2134

```

// Old gross behavior that we are currently enforcing
enum ThingKV{
  a;
  b;
}
@knownValues(ThingKV)
model Thing is string {};

// New hotness
enum Thing {
  a;
  b;
}
```

In the new behavior, all regular Cadl enums should be translated as extensible by emitters. The exception, where an enum is NOT extensible, will be denoted by the `@fixed` decorator:

```
@fixed
enum YesNo {
 yes;
 no;
}
```

weidongxu-microsoft:
SDK will temporary disable tests on https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/extensible-enums/main.cadl#L9-L17 (which now being fixed enum, before this behavior change)

tjprescott:
> SDK will temporary disable tests on https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/extensible-enums/main.cadl#L9-L17 (which now being fixed enum), before this behavior change.

@weidongxu-microsoft I don't understand this comment. The whole point of making the change to this test is so that emitters will begin failing the revised checks to ensure that all emitters are updated to generate extensible enums correctly.

weidongxu-microsoft:
@tjprescott 
I didn't found the `@fixed` in microsoft/cadl. Just figured out it is in `Azure.Core`. Sure, will switch the behavior in emitter.

PR #164, opened 2022-11-11
Update for new pattern of extensible enum

tjprescott:
Please see: https://github.com/Azure/cadl-azure/pull/2229

Fixes: #158.

This does not add any new scenarios. It updates an existing one.


packages/cadl-ranch-specs/http/extensible-enums/main.cadl
```diff
@@ -6,7 +6,7 @@ using Cadl.Http;
 @scenarioService("/extensible-enums")
 namespace ExtensibleEnums;
```
timotheeguerin, line 7: Think we should just change completely the structure of this test.

```cadl
namespace Enums
```

and then have `Extensible` and `Fixed` subnamespace showing the use case for both

tjprescott: I made that change (I think) but I don't know if I did it right. The CI won't pass anyhow until cadl-azure-core is released. I wanted to get the extensible test in quicker so emitters could be updated. 

I also don't know if the fixed test makes any sense, but let me know what you think.

packages/cadl-ranch-specs/http/enums/fixed/main.cadl
```diff
@@ -0,0 +1,40 @@
+import "@cadl-lang/rest";
+import "@azure-tools/cadl-azure-core";
+import "@azure-tools/cadl-ranch-expect";
+
+using Cadl.Http;
+
+@scenarioService("/fixed-enums")
+namespace FixedEnums;
```
timotheeguerin, line 8: I would put both those scenarios into the same file so with the following structure


```
namespace Enums;

namespace Extensible {
 ...
}
namespace Fixed {}
```

or at least have the namespace called like that so in the dashboard it is collapsed under `Enums`

packages/cadl-ranch-specs/http/enums/fixed/main.cadl
```diff
@@ -0,0 +1,40 @@
...
+  @scenarioDoc("Expect to send a known value. Mock api expect to receive 'Monday'")
+  @put
+  @route("/known-value")
+  putKnownValue(@body body: DaysOfWeekEnum): void;
+
+  @scenario
+  @scenarioDoc("Expect to handle an unknown value. Mock api expect to receive 'Weekend'")
+  @put
+  @route("/unknown-value")
+  putUnknownValue(@body body: DaysOfWeekEnum): void;
```
timotheeguerin, line 39: Is that a scenario that is possible with fixed enums? I coudl understad the reverse, a service actually returning a new value but sending an unknown value should that work?

tjprescott: Well, in that case I think it would be a client failure... it there a way to model that in cadl-ranch?

timotheeguerin: I mean you could just model it as the service returning this unknown value and then the generator test needs to do whatever it wants to handle that(try catch the exception for example)

Maybe we  could have another decorator that says that this scenario is really to handle a case that is bad and fail nicely

tjprescott:
This is blocked until cadl-azure-core 0.10.0 is released.

packages/cadl-ranch-specs/http/enums/fixed/main.cadl
```diff
@@ -0,0 +1,48 @@
+import "@cadl-lang/rest";
+import "@azure-tools/cadl-azure-core";
+import "@azure-tools/cadl-ranch-expect";
+
+using Azure.Core;
+using Cadl.Http;
+
+@scenarioService("/fixed-enums")
+namespace FixedEnums;
```
timotheeguerin, line 9:
```suggestion
@scenarioService("/enums/fixed")
namespace Enums.Fixed;
```

So we can get this structure in the ui
Enums:
   Fixed:
      ...
   Extensible:
      ...

packages/cadl-ranch-specs/http/enums/fixed/main.cadl
tjprescott: @iscai-msft is there something you recommend here?

lmazuel (approved):

PR #167, opened 2022-11-17
add tests for client path parameters

iscai-msft:


packages/cadl-ranch-specs/http/models/property-types/main.cadl
```diff
@@ -172,4 +172,4 @@ interface DictionaryString extends ModelOperations<DictionaryStringProperty, "{'
 model NeverProperty is ModelTemplate<never>;
 @route("/never")
 @operationGroup
-interface Never extends ModelOperations<NeverProperty, "<dont include this property>"> {}
+interface Never extends ModelOperations<NeverProperty, "<don't include this property>"> {}
```
iscai-msft, line 175: fixed bc it failed cspell

packages/cadl-ranch-specs/http/clients/parameters/path/main.cadl
```diff
@@ -0,0 +1,56 @@
...
+
+@scenario
+@scenarioDoc("""
+Expected response body:
+  ```json
+  {"id": 1}
+  ```
+""")
+@doc("Get a resource")
+op get is ResourceRead<Product>;
```
timotheeguerin, line 56: I would just remove the resource part no? If we are testing server params maybe just 
```
op noOperationParams(): NoContentResponse;

// And maybe adding this one to test it can still add path param for operations
op withOperationPathParam(@path id: string): NoContentResponse;
```

like we have for authentication scenarios

iscai-msft: I can split this up more, I kept it with templating bc we have api version both in the path, and the template will return api version as aquery param (this is what's happening in anomaly detector). Maybe the two operations you've listed, and then another one with a template to test the duplicate api versions?

packages/cadl-ranch-specs/http/models/property-types/main.cadl
timotheeguerin: or did I merge without the ci passing...

tadelesh:
@iscai-msft could you help to review?

timotheeguerin (approved):

iscai-msft:
@tadelesh talked with @timotheeguerin , we don't feel the need to have a core test right now (in the future all core tests should also be in the azure core folder). We're then combining the existing parameterized test with the test you added with api version as a path aparameter and they'll both be under the `Server.Path` route. Thanks!

PR #181, opened 2022-12-22
Update url of visibility mock api

timotheeguerin:


lmazuel (approved):
I can confirm it's fixing a real problem :p

PR #186, opened 2022-12-28
Add projection tests

lmazuel:

### DIFF
--- a/packages/cadl-ranch-specs/http/projection/main.cadl
+++ b/packages/cadl-ranch-specs/http/projection/main.cadl
@@ -55,3 +55,5 @@ CADL name is JSON, Projection is client name per language override. Send:
 @route("/language")
 @post
 op languageProjection(...Project): void;
+
+// May add query/header project after https://github.com/microsoft/cadl/issues/1487
--- a/packages/cadl-ranch-specs/http/projection/main.cadl
+++ b/packages/cadl-ranch-specs/http/projection/main.cadl
@@ -1,0 +1,57 @@
+import "@cadl-lang/rest";
+import "@azure-tools/cadl-ranch-expect";
+
+using Cadl.Http;
+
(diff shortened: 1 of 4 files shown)


packages/cadl-ranch-specs/http/projection/main.cadl
```diff
@@ -0,0 +1,59 @@
+import "@cadl-lang/rest";
+import "@azure-tools/cadl-ranch-expect";
+
+using Cadl.Http;
+
+@doc("Projection")
+@supportedBy("dpg")
+@scenarioService("/projection")
+namespace Projection;
```
timotheeguerin, line 9: Think this should either be `ProjectedName` or `TargetProjection` to describe it is testing just that one `target` projection

packages/cadl-ranch-specs/http/projection/main.cadl
```diff
@@ -0,0 +1,59 @@
...
+
+model Project {
+  @projectedName("json", "codegen")
+  producedBy?: string;
```
timotheeguerin, line 13: Can we add `/**` comments to describe for each on of those properties what is expected name to be sent via json and generated in clients

packages/cadl-ranch-specs/http/projection/main.cadl
```diff
@@ -0,0 +1,59 @@
...
+
+  @projectedName("client", "createdBy")
+  builtfrom?: string;
+
+  @projectedName("client", "DontUseMeAsName") // Should be ignored
+  @projectedName("csharp", "MadeFor")
```
timotheeguerin, line 19: Should we have a different name for each languages just to make sure it is respected. Like `madeForCS`. At that point would also respect the naming convention of each language

timotheeguerin (approved):

PR #190, opened 2023-01-17
testcase for type of url

changlong-liu:


packages/cadl-ranch-specs/http/models/property-types/main.cadl
```diff
@@ -184,3 +192,10 @@ model NeverProperty is ModelTemplate<never>;
 @route("/never")
 @operationGroup
 interface Never extends ModelOperations<NeverProperty, "<don't include this property>"> {}
+
+// Test a model with a url property
+@doc("Model with a property type of url.")
```
iscai-msft, line 197: i think this should be a whole new test. We can do a parameterized server test where we just have a simple head call, or something like that. When I think of testing urls, I don't think of testing properties, so for clarity I think it's better to have a separate test case. What are your thoughts?

changlong-liu: Yes @iscai-msft , I updated to add a brand new test folder for `server`, and this can avoid breaking changes comparing to adding to existing folder.

iscai-msft (approved):

PR #198, opened 2023-01-30
fix @convenientAPI usage

changlong-liu:
Fix below compile error when upgrading `@azure-tools/cadl-dpg` from "0.25.0" to "0.26.0-dev.5" in emitter:
image

### DIFF
--- a/.changeset/smooth-ducks-hang.md
+++ b/.changeset/smooth-ducks-hang.md
@@ -1,0 +1,5 @@
+---
+"@azure-tools/cadl-ranch-specs": patch
+---
+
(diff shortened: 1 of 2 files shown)


pshao25 (approved):

iscai-msft:
my understanding is that these decorators should just work without having to specify `True`

iscai-msft:
ok talked with the cadl team and we want to force users to specify `true` or `false` in the decorator, so this PR is correct. GOing to merge, thanks @changlong-liu !

iscai-msft (approved):

PR #200, opened 2023-02-01
Update azure core case for ResourceOperations

weidongxu-microsoft:
Fix a bug of route URL. Path should be `/azure/core/users/{id}` 

Migrate to `ResourceOperations` (service trait), as `Azure.Core.ResourceCreateOrUpdate` is [`@removed`](https://github.com/Azure/cadl-azure/blob/main/packages/cadl-azure-core/lib/operations.cadl#L324-L328)

It does not directly use `StandardResourceOperations`, as David told me `StandardResourceOperations` might get removed as well (prefer service to explicitly define the `ResourceOperations`).
relate to https://github.com/Azure/cadl-azure/issues/2468

### DIFF
--- a/.changeset/dirty-guests-film.md
+++ b/.changeset/dirty-guests-film.md
@@ -1,0 +1,5 @@
+---
+"@azure-tools/cadl-ranch-specs": minor
+---
+
+Use ResourceOperations for Azure.Core Preview2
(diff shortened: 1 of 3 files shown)


weidongxu-microsoft:
@iscai-msft
Pending on follow-up. I plan to add other simple CRUD + ResourceList and ResourceAction (as part of verification that Azure.Core operations is good for all emitters). Let me know whether it already covered by plan from Redmond.
Also welcome suggestion whether we add them in this file, or use separate Cadl sources.

Futher pending would be LRO operations, operations with simple traits.

haolingdong-msft (approved):

iscai-msft (approved):

PR #216, opened 2023-02-28
update collection format test

haolingdong-msft:
1. update response body from object to string

[Cadl definition](https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/collectionFormat/main.cadl#L43) defines the return model as string, but mockapi.ts returns an object. It will cause strong type languages like java and .net can't convert return type to string in convenient method .

```typescript
op testMulti(...MultiCollectionFormatParameter): string;
```

2. update csv case

Reasons:

- https://www.rfc-editor.org/rfc/rfc3986#section-2.4 RFC does not have restriction for whether we do encoding. 

- I've also verified in Java, our existing DPG behavior (non-cadl one) will encode comma. The url would be: 
```
info: GET /collectionFormat/csv?colors=blue%2Cred%2Cgreen
```

- I saw the autorest.testserver [case](https://github.com/Azure/autorest.testserver/blob/79b23e5ebff1caca9dbda93a9220485331537e5f/swagger/url.json#L740) on csv format for header param will do the encoding. I think it would be better to make it consistent in both header and query param in csv collection format. 

So I make both encoding comma and not encoding be successful case. 

### DIFF
--- a/packages/cadl-ranch-specs/http/collectionFormat/main.cadl
+++ b/packages/cadl-ranch-specs/http/collectionFormat/main.cadl
@@ -35,7 +35,7 @@ alias DefaultCollectionFormatParameter = {
   colors: string[];
 };
 
-model Message {
+model MessageResponse {
   message: string;
 }
 
@@ -44,14 +44,14 @@ model Message {
 This test is testing sending a multi collection format array query parameters
 """)
 @route("/multi")
-op testMulti(...MultiCollectionFormatParameter): Message;
+op testMulti(...MultiCollectionFormatParameter): MessageResponse;
 
 @scenario
 @scenarioDoc("""
 This test is testing sending a csv collection format array query parameters
 """)
 @route("/csv")
-op testCsv(...CsvCollectionFormatParameter): Message;
+op testCsv(...CsvCollectionFormatParameter): MessageResponse;
 
 // @scenario
 // @scenarioDoc("""
--- a/packages/cadl-ranch-specs/http/collectionFormat/main.cadl
+++ b/packages/cadl-ranch-specs/http/collectionFormat/main.cadl
@@ -35,19 +35,23 @@ alias DefaultCollectionFormatParameter = {
   colors: string[];
 };
 
+model Message {
+  message: string;
+}
+
 @scenario
 @scenarioDoc("""
 This test is testing sending a multi collection format array query parameters
 """)
 @route("/multi")
-op testMulti(...MultiCollectionFormatParameter): string;
+op testMulti(...MultiCollectionFormatParameter): Message;
 
 @scenario
 @scenarioDoc("""
 This test is testing sending a csv collection format array query parameters
 """)
 @route("/csv")
-op testCsv(...CsvCollectionFormatParameter): string;
+op testCsv(...CsvCollectionFormatParameter): Message;
 
 // @scenario
 // @scenarioDoc("""
--- a/packages/cadl-ranch-specs/http/collectionFormat/mockapi.ts
+++ b/packages/cadl-ranch-specs/http/collectionFormat/mockapi.ts
@@ -24,7 +24,7 @@ Scenarios.CollectionFormat_testCsv = passOnSuccess(
     req.expect.containsQueryParam("colors", "blue,red,green");
     return {
       status: 200,
-      body: json( `A multi collection format array was successfully received`),
+      body: json(`A multi collection format array was successfully received`),
     };
   }),
 );
--- a/packages/cadl-ranch-specs/http/collectionFormat/mockapi.ts
+++ b/packages/cadl-ranch-specs/http/collectionFormat/mockapi.ts
@@ -8,12 +8,12 @@ Scenarios.CollectionFormat_testMulti = passOnSuccess(
     if (req.originalRequest.originalUrl.includes("colors=blue&colors=red&colors=green")) {
       return {
         status: 200,
-        body: json({ message: `A multi collection format array was successfully received`}),
+        body: json({ message: `A multi collection format array was successfully received` }),
       };
     } else {
       return {
         status: 400,
-        body: json({ message: `Expected colors=blue&colors=red&colors=green after serialization`}),
+        body: json({ message: `Expected colors=blue&colors=red&colors=green after serialization` }),
       };
     }
   }),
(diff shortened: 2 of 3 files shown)


weidongxu-microsoft (approved):
LGTM on 200

I assume if validation fail, code should throw ValidationError?

haolingdong-msft:
If fails, `req.expect.containsQueryParam("colors", "blue,red,green");` will throw 400 `ValidationError`.

> LGTM on 200
> 
> I assume if validation fail, code should throw ValidationError?

weidongxu-microsoft:
> If fails, `req.expect.containsQueryParam("colors", "blue,red,green");` will throw 400 `ValidationError`.
> 
> > LGTM on 200
> > I assume if validation fail, code should throw ValidationError?

Then why
```
      return {
        status: 400,
        body: json(`Expected colors=blue&colors=red&colors=green after serialization`),
      };
```

haolingdong-msft:
> > If fails, `req.expect.containsQueryParam("colors", "blue,red,green");` will throw 400 `ValidationError`.
> > > LGTM on 200
> > > I assume if validation fail, code should throw ValidationError?
> 
> Then why return 400
> 
> ```
>       return {
>         status: 400,
>         body: json(`Expected colors=blue&colors=red&colors=green after serialization`),
>       };
> ```

I'm not the author of the code, but I guess it's because this case cannot use `req.expect` to verify, so return 400 directly.

iscai-msft:
@haolingdong-msft would you be able to update `req.expect.containsQueryParam` to also work for multiformat? We can have an optional parameter accept collection format

packages/cadl-ranch-specs/http/collectionFormat/mockapi.ts
```diff
@@ -8,24 +8,33 @@ Scenarios.CollectionFormat_testMulti = passOnSuccess(
...
-      status: 200,
-      body: json({ message: `A multi collection format array was successfully received` }),
-    };
+    if (
+      req.originalRequest.originalUrl.includes("colors=blue,red,green") ||
+      req.originalRequest.originalUrl.includes("colors=blue%2Cred%2Cgreen")
+    ) {
+      return {
+        status: 200,
+        body: json({ message: `A multi collection format array was successfully received` }),
```
weidongxu-microsoft, line 30: This should be a string?

packages/cadl-ranch-specs/http/collectionFormat/mockapi.ts
```diff
@@ -1,4 +1,4 @@
-import { passOnSuccess, mockapi, json } from "@azure-tools/cadl-ranch-api";
+import { passOnSuccess, mockapi, json, ValidationError } from "@azure-tools/cadl-ranch-api";
```
weidongxu-microsoft, line 1: where is `ValidationError` used?

packages/cadl-ranch-specs/http/collectionFormat/mockapi.ts
```diff
@@ -8,24 +8,33 @@ Scenarios.CollectionFormat_testMulti = passOnSuccess(
...
 Scenarios.CollectionFormat_testCsv = passOnSuccess(
   mockapi.get("/collectionFormat/csv", (req) => {
-    req.expect.containsQueryParam("colors", "blue,red,green");
-    return {
-      status: 200,
-      body: json({ message: `A multi collection format array was successfully received` }),
-    };
+    if (
+      req.originalRequest.originalUrl.includes("colors=blue,red,green") ||
+      req.originalRequest.originalUrl.includes("colors=blue%2Cred%2Cgreen")
```
weidongxu-microsoft, line 26: Just a note on this, autorest.testserver here
So I think Java encoding it since 3 years ago.
```
info: GET /queries/array/csv/string/valid?arrayQuery=ArrayQuery1%2Cbegin%21%2A%27%28%29%3B%3A%40%20%26%3D%2B%24%2C/?%23%5B%5Dend%2C%2C 200 6.420 ms - -
```

haolingdong-msft:
> @haolingdong-msft would you be able to update `req.expect.containsQueryParam` to also work for multiformat? We can have an optional parameter accept collection format

Hi @iscai-msft, I don't quite get your point. In multi collection format case, we use & to make `colors` query param as an array. Then when using `request.query["colors "]` ([code](https://github.com/Azure/cadl-ranch/blob/b47c44524c88aa4a68b67fa9b4fb4c897f97771e/packages/cadl-ranch-api/src/request-validations.ts#L96)) to get the value of `colors`, we will get `"blue,red,green"`. we can use `req.expect.containsQueryParam("colors", "blue,red,green");` to verify, but here we lose the context of using & format, so we verify the url directly.

weidongxu-microsoft:
`req.expect.containsQueryParam("colors", "blue,red,green")` seems fine for csv, nothing is lost?

I assume Isabella is thinking about improving `containsQueryParam` so that it would also work for multi. E.g. `containsQueryParam("coloers", "blue", "red", "green")` <-- this is my pseudo code.

Also current code is not very accurate. One can send `colors=blue&colors=red&colors=greeninvalid` and still pass the test.

weidongxu-microsoft:
@qiaozha as first author.

iscai-msft:
> `req.expect.containsQueryParam("colors", "blue,red,green")` seems fine for csv, nothing is lost (however we seems cannot use it directly since we'd like to handle percent-encoded comma as well)?
> 
> I assume Isabella is thinking about improving `containsQueryParam` so that it would also work for multi (maybe csv as well). E.g. `containsQueryParam("coloers", ["blue", "red", "green"])` <-- this is my pseudo code, not what you or Isabella had in mind.
> 
> Also current code is not very accurate. One can send `colors=blue&colors=red&colors=greeninvalid` and still pass the test.

Yep, but we should be able to handle various formats too, i.e. "csv" collection format, "pipes", "multi". So we can have a `collectionFormat` parameter that defaults to `csv` (i think that's what we're functionally defaulting to rn. Here are a few:
```
[CollectionFormat.CSV]: ",",
[CollectionFormat.SSV]: " ",
[CollectionFormat.TSV]: "\t",
[CollectionFormat.Pipes]: "|",
[CollectionFormat.Multi]: "colors=blue&colors=red&colors=green",
```

haolingdong-msft:
Hi @iscai-msft , https://github.com/Azure/cadl-ranch/issues/219, I create a seperate issue on this, I will work on this issue.  Could we merge this bug fix first as it blocks other languages.

packages/cadl-ranch-specs/http/collectionFormat/main.cadl
```diff
@@ -35,19 +35,23 @@ alias DefaultCollectionFormatParameter = {
   colors: string[];
 };
 
+model Message {
+  message: string;
```
weidongxu-microsoft, line 39: .net also not support property be same name as model... :-(

weidongxu-microsoft:
@MaryGao as Qiaoqiao OOF

iscai-msft (approved):

PR #218, opened 2023-03-01
fix(test): remove duplicated discriminator decorators

### LINKED ISSUE #217: [bug][test] duplicated discriminator property declared in test case
## Problem

See: https://github.com/Azure/cadl-ranch/blob/e2d7a8bb4bc40b55e624d01015ecdb1fa168f73e/packages/cadl-ranch-specs/http/special-words/main.cadl#L40-L50

Discriminator property `model.kind` has been declared twice, which is a problematic case. Some language emitters can tolerate that, but some cannot (like .Net). Unless we want to test the robustness of emitters, we should remove the duplicated declaration.

archerzz:
fix #217 

### DIFF
--- a/.changeset/tall-tomatoes-learn.md
+++ b/.changeset/tall-tomatoes-learn.md
@@ -1,0 +1,5 @@
+---
+"@azure-tools/cadl-ranch-specs": patch
+---
+
+fix test case on discriminator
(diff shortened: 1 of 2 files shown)


pshao25 (approved):

lirenhe:
Could this be handled at dotnet codegen side? I believe this does not introduce any invalid behavior, right?

archerzz:
> Could this be handled at dotnet codegen side? I believe this does not introduce any invalid behavior, right?

@lirenhe I've confirm with Mark on teams. He agreed to remove this duplicated decorator. I think it would be better to keep the emitter simpler.

lirenhe (approved):

iscai-msft (approved):

PR #226, opened 2023-03-06
add test for `@overload` and multi content-type

msyyc:

### DIFF
--- a/packages/cadl-ranch-specs/http/overload/main.cadl
+++ b/packages/cadl-ranch-specs/http/overload/main.cadl
@@ -1,0 +1,74 @@
+import "@cadl-lang/rest";
+import "@azure-tools/cadl-ranch-expect";
+
+using Cadl.Http;
+
+@doc("Overload")
+@supportedBy("dpg")
+@scenarioService("/overload")
+namespace Overload;
+
+@scenario
+@scenarioDoc("""
+send "hello world" with `conten-type` "text/plain";
+send binary data "hello world" with `conten-type` "application/octet-stream"
+""")
+@route("/bytesOrString")
+@post
+op uploadBytesOrString(
+  @body data: string | bytes,
+  @header contentType: "text/plain" | "application/octet-stream"
+): void;
+
+@scenario
+@scenarioDoc("""
+TODO: optimize cadl-ranch to skip it because the op is only for overload and no need to test
+""")
+@overload(uploadBytesOrString)
+@route("/bytesOrString/skip1")
+@post
+op uploadBytesOrStringSkip1(@body data: string, @header contentType: "text/plain"): void;
+
+@scenario
+@scenarioDoc("""
+TODO: optimize cadl-ranch to skip it because the op is only for overload and no need to test
+""")
+@overload(uploadBytesOrString)
+@route("/bytesOrString/skip2")
+@post
+op uploadBytesOrStringSkip2(@body data: bytes, @header contentType: "application/octet-stream"): void;
+
+model Data {
+  hello: string;
+}
+
+@scenario
+@scenarioDoc("""
+send "hello world" with `conten-type` "text/plain";
+send
+```json
+{ "hello": "world"}
+```
+with `conten-type` "application//json"
+""")
+@route("/JsonOrString")
+@post
+op uploadJsonOrString(@body data: string | Data, @header contentType: "text/plain" | "application/json"): void;
+
+@scenario
+@scenarioDoc("""
+TODO: optimize cadl-ranch to skip it because the op is only for overload and no need to test
+""")
+@overload(uploadJsonOrString)
+@route("/JsonOrString/skip3")
+@post
+op uploadJsonOrStringSkip3(@body data: string, @header contentType: "text/plain"): void;
+
+@scenario
+@scenarioDoc("""
+TODO: optimize cadl-ranch to skip it because the op is only for overload and no need to test
+""")
+@overload(uploadJsonOrString)
+@route("/JsonOrString/skip4")
+@post
+op uploadJsonOrStringSkip4(@body data: Data, @header contentType: "application/json"): void;
--- a/.changeset/rude-pigs-own.md
+++ b/.changeset/rude-pigs-own.md
@@ -1,0 +1,5 @@
+---
+"@azure-tools/cadl-ranch-specs": patch
+---
+
+add test for overload
--- a/packages/cadl-ranch-specs/cadl-ranch-summary.md
+++ b/packages/cadl-ranch-specs/cadl-ranch-summary.md
@@ -1535,6 +1535,50 @@ Expected input body:
 }
 ```
 
+### Overload_uploadBytesOrString
+
+- Endpoint: `post /overload/bytesOrString`
+
+send "hello world" with `conten-type` "text/plain";
+send binary data "hello world" with `conten-type` "application/octet-stream"
+
+### Overload_uploadBytesOrStringSkip1
+
+- Endpoint: `post /overload/bytesOrString/skip1`
+
+TODO: optimize cadl-ranch to skip it because the op is only for overload and no need to test
+
+### Overload_uploadBytesOrStringSkip2
+
+- Endpoint: `post /overload/bytesOrString/skip2`
+
+TODO: optimize cadl-ranch to skip it because the op is only for overload and no need to test
+
+### Overload_uploadJsonOrString
+
+- Endpoint: `post /overload/JsonOrString`
+
+send "hello world" with `conten-type` "text/plain";
+send
+
+```json
+{ "hello": "world" }
+```
+
+with `conten-type` "application//json"
+
+### Overload_uploadJsonOrStringSkip3
+
+- Endpoint: `post /overload/JsonOrString/skip3`
+
+TODO: optimize cadl-ranch to skip it because the op is only for overload and no need to test
+
+### Overload_uploadJsonOrStringSkip4
+
+- Endpoint: `post /overload/JsonOrString/skip4`
+
+TODO: optimize cadl-ranch to skip it because the op is only for overload and no need to test
+
 ### ProjectedName_jsonProjection
 
 - Endpoint: `post /projection/json`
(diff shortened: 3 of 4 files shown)


packages/cadl-ranch-specs/http/overload/main.cadl
```diff
@@ -0,0 +1,74 @@
...
+op uploadJsonOrString(@body data: string | Data, @header contentType: "text/plain" | "application/json"): void;
+
+@scenario
+@scenarioDoc("""
+TODO: optimize cadl-ranch to skip it because the op is only for overload and no need to test
+""")
+@overload(uploadJsonOrString)
+@route("/JsonOrString/skip3")
+@post
+op uploadJsonOrStringSkip3(@body data: string, @header contentType: "text/plain"): void;
```
haolingdong-msft, line 65: what does skip3 mean?

msyyc: just means "I have to define it to avoid compiler error but I don't want to generate API for it"

haolingdong-msft:
We may need to wait for other languages to support this or put this change to a minor version, because for languages (like Java, .net) not supported this, the generated code can't compile..

packages/cadl-ranch-specs/http/overload/main.cadl
weidongxu-microsoft: Is it just the overload to `uploadJsonOrString`, which means its name is irrelavent?

chunyu3: Why do we need to update the route? I think in service side, there is only one route for this.

packages/cadl-ranch-specs/http/overload/main.cadl
```diff
@@ -0,0 +1,74 @@
...
+namespace Overload;
+
+@scenario
+@scenarioDoc("""
+send "hello world" with `conten-type` "text/plain";
+send binary data "hello world" with `conten-type` "application/octet-stream"
+""")
+@route("/bytesOrString")
+@post
+op uploadBytesOrString(
```
m-nash, line 18: Should we have some way to mark this as "don't use" for the clients?  This isn't typically what `overload` means.  In this case we are trying to use `overload` as `overwrite with 1 or more items`.  I prefer we change the decorator or have a way to mark the original as "don't use".

@johanste @lmazuel @srnagar 

There was also discussion to use a different decorator like allow duplicate routes but I can't seem to find any documentation on it.  @srnagar do you remember the name of that decorator or have a link to the documentation?

packages/cadl-ranch-specs/http/overload/main.cadl
msyyc: > Is it just an overload to `uploadJsonOrString`, which means its name is irrelevant?

We have to make a different name for overload op otherwise cadl compiler will report error

msyyc: > Why do we need to update the route? I think in service side, there is only one route for this.

It is the limit of cadl-ranch. Without the `rout`, cadl-ranch will report error

packages/cadl-ranch-specs/http/overload/main.cadl
msyyc: Hi @m-nash The testcase is to test the [scenario1](https://gist.github.com/msyyc/87a79b917deec25639f9f10d20589e55#scenario-1) and [scenario2](https://gist.github.com/msyyc/87a79b917deec25639f9f10d20589e55#scenario-2). The testcase is limited by cadl-ranch so I think the scenario doc is much clear. 
In the scenario1 and 2, author could define very clear map for body type and content type which gives language emitter enough info to generate user-friendly SDK code like [here](https://github.com/Azure/autorest.python/blob/6b2c21e5b96ca9284fa8b9edc7f407c18774390d/packages/cadl-python/test/mock_api_tests/test_overload.py#L16-L17). With overload, users don't need to input `content-type` since SDK can judge it with input type which the abosolutely `overload` means.

packages/cadl-ranch-specs/http/overload/main.cadl
chunyu3: But there is only one rest api which will accept multiple content-type. 
If you want to use @overload to split to different sdk methods for multi content-type, the request Uri/route should be the same.

pshao25: > Without the rout, cadl-ranch will report error

Then we should fix cadl-ranch instead of putting this workaround.

packages/cadl-ranch-specs/http/overload/main.cadl
pshao25: If only python decides to support the `overload` thing, I would suggest put this spec and corresponding mockapi into python codegen repo. You just need to pass the mockapi path in your codegen repo to cadl-ranch server, and it could serve as a service in your own repo. Example: 
Mockapi: https://github.com/Azure/autorest.csharp/blob/2ff04cd6760e4371dc031d2022c3b43440bdb101/test/CadlRanchMockApis/src/Lro-Basic-Cadl.ts
Spec:
https://github.com/Azure/autorest.csharp/blob/2ff04cd6760e4371dc031d2022c3b43440bdb101/test/TestProjects/Lro-Basic-Cadl/Lro-Basic-Cadl.cadl
Test:
https://github.com/Azure/autorest.csharp/blob/2ff04cd6760e4371dc031d2022c3b43440bdb101/test/CadlRanchProjects.Tests/LroBasic-Cadl.cs

packages/cadl-ranch-specs/http/overload/main.cadl
iscai-msft: this is a cadl issue with route right now, we're blocked on cadl for this

iscai-msft:
@haolingdong-msft my expectation is java should still be able to generate, they just wouldn't actually generate overloads in this case. Does java code break on generation here?

haolingdong-msft:
> @haolingdong-msft my expectation is java should still be able to generate, they just wouldn't actually generate overloads in this case. Does java code break on generation here?
Java will break on generating convenient methods, I've tried locally...

weidongxu-microsoft:
Java does not block this test, just that when merge it, consider make a certain version bump and let dev know (so if the feature is not ready in codegen, we can avoid upgrading to that version).

The "overload" is very python syntax in Cadl. It does not have one on one mapping to Java (likely neither .NET) API. Hence there is high likelihood of compile error, if this feature is not supported but get generated.

m-nash:
> @haolingdong-msft my expectation is java should still be able to generate, they just wouldn't actually generate overloads in this case. Does java code break on generation here?

I have bounced back on forth on this quite a bit.  When the typespec creates a situation where we cannot know what to generate which of the following do we do.

1. Generate "junk" that we "hope" will be caught in apiview and fixed with manual customizations or client.tsp files.
2. Throw an error saying your typespec doesn't work please add client.tsp customizations or code customizations.

If we go with #1 we have to be very confident that things will actually get caught.  Even if we can achieve this high level of confidence we are kind of wasting our time to generate "junk" when we know that "junk" should never get used.  Its much more efficient use of our time to throw an error and expect correct typespec in the first place.

I am pretty firmly in the #2 camp at this point so anything that doesn't work for all languages should not get checked into cadl-ranch since it leaves a couple languages in a state where they have to ignore certain tests which defeats the purpose of using cadl-ranch as a collection of "valid templates".

Can the other 3 language representatives @iscai-msft, @srnagar, @joheredi and @lmazuel weigh in on which of those two strategies you think is best?

iscai-msft:
@m-nash this was actually something i was going to bring up in dpg sync today 😆 . I'm more in camp 1, at least for cadl ranch. Generally speaking I think unless something really breaks other languages, we should add it to cadl-ranch if it's a cadl feature and at least some languages support it. We've done this , for example in Python we don't differentiate between `int32`, `int64` etc (it's all integer), but we still have tests for all of these different kinds of ints, Python's job is just a little easier here because there isn't anything for us to do.

I'm confused how this would be junk for c# / languages that don't support overloads. My assumption is the overload functions wouldn't be picked up and you just generate the underlying operation. Is this incorrect?

packages/cadl-ranch-specs/http/overload/main.cadl
joheredi: @m-nash are you proposing to decorate the "general/overloaded" function with `do not use` to indicate emitters that they shouldn't generate that function but just the overloads? Or just mark them internally in the "code model" for emitters to decide what to do? If it is the latter I agree, I'm not sure I like the decorator idea.

FWIW this is the same pattern that [TypeScript uses for overloading](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads), where the overloaded function is not exposed, but we do need to generate it.

joheredi:
I agree with @iscai-msft. I would call it camp 3: we make sure the compiler/typespec-core marks the code model correctly so that emitters that consider some parts Junk can omit them. For example, in Typescript, we would have to generate the base function (but not expose it) plus all the overloads.

We shouldn't, on purpose, generate Junk in hopes that the archboard catches it, that won't scale.

srnagar:
When we have a TypeSpec that doesn't works for some languages, the emitters of those languages should throw an error. This will allow us to define necessary client.tsp or <language>-client.tsp that can then generate valid code for these languages right from the beginning. 

Generating invalid (junk) code is not desirable as it can accidentally slip through arch board reviews. Even if it's caught by arch board, the turnaround time to generate invalid code, waiting for arch board feedback and then revisiting the spec to add a client.tsp is too long and can lead to bad user experience for service teams.

However, I am fine with cadl-ranch having tests that work for some languages and other languages can either ignore those or have their <language>-client.tsp added to cadl-ranch repo to get those tests to work.

m-nash:
Spoke offline with the language reps.  Going forward all PRs into cadl-ranch should be correct representations for all Tier 1 languages.  This might mean that in order to check in we will need both a main.tsp and a client.tsp with the additional information needed for 1 or more languages.

@srnagar has an alternative suggestion for this PR specifically which he will add shortly.

johanste:
There is nothing that prevents a service from having different operations under the same route (using shared routes). 

However, in the examples above, there is nothing that indicates which methods to emit. And it cannot be computed if the "base" operation (the one with unions) needs to be emitted since the two constituent methods do not cover all cases of the base method. For `@overloads` the intent is that the individual overloading operations cover all possible permutations. 

It also doesn't cover the scenario where you want an overload with different routes/paths.

PR #234, opened 2023-03-14
Add test cases for nullable items

MaryGao:


pshao25 (changes requested):
Can we remove the union part? It is not going to be a valid case in .Net and Java.

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -1413,6 +1413,26 @@ Expected input body:
...
+```
+
+### Models_Property_Types_NullableModel_put
+
+- Endpoint: `put /models/properties/types/nullable-model`
+
+Expected input body:
+
+```json
+{ "property": null }
```
weidongxu-microsoft, line 1433: Personally, as REST API, I don't think PUT `{ "property": null }` semantically be any different from a PUT with `{}`.

But I won't block the case.

lirenhe: Will JS treat '{}' as undefined and 'null' as null?

MaryGao: Let me change the test case to JSON MERGE PATCH with partial update - https://www.rfc-editor.org/rfc/rfc7386#section-1.

```
{
         "a":"z",
         "c": {
       "f": null
         }
       }

```

MaryGao:
> Can we remove the union part? It is not going to be a valid case in .Net and Java.

@pshao25 @weidongxu-microsoft Is this decision finalized that we don't support nullable model/array/dictionary in .Net or Java? Or this is just current status.

pshao25 (approved):

weidongxu-microsoft:
> > Can we remove the union part? It is not going to be a valid case in .Net and Java.
> 
> @pshao25 @weidongxu-microsoft Is this decision finalized that we don't support nullable model/array/dictionary in .Net or Java? Or this is just current status.

What is the "union part" about?

Java support nullable in array/dictionary, and nullable model for json merge patch.

packages/cadl-ranch-specs/cadl-ranch-summary.md
weidongxu-microsoft: Yes, json merge patch would be great!

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -1513,6 +1553,29 @@ Expected input body:
...
+
+### Models_Property_Types_ModelNullableProperty_patch
+
+- Endpoint: `patch /models/properties/types/nullable-property`
+
+Expected input body:
+
+```json
+{ "id": 1, "property": null }
+```
```
MaryGao, line 1577: Here is the merge patch with partial update case.

packages/cadl-ranch-specs/http/models/property-types/main.tsp
```diff
@@ -170,16 +170,78 @@ model CollectionsModelProperty is ModelTemplate<InnerModel[]>;
...
 @operationGroup
 interface DictionaryString extends ModelOperations<DictionaryStringProperty, "{'k1': 'hello', 'k2': 'world'}"> {}
 
+// Test a model with a nullable dictionary property
+@doc("Model with collection with nullable values")
+model DictionaryNullableValueProperty is ModelTemplate<Record<string | null>>;
+@route("/dictionary/nullable-value")
+@operationGroup
+interface DictionaryNullableValue
+  extends ModelOperations<DictionaryNullableValueProperty, "{'k1': 'hello', 'k2': 'world', 'k3': null}"> {}
```
iscai-msft, line 193: what happens if we instead just have a nullable dictionary / array? (no model involved). would be good to have tests for that as well

iscai-msft: oh i see @MaryGao already left a comment about this, i'll follow the response to that question

iscai-msft:
> > Can we remove the union part? It is not going to be a valid case in .Net and Java.
> 
> @pshao25 @weidongxu-microsoft Is this decision finalized that we don't support nullable model/array/dictionary in .Net or Java? Or this is just current status.

i also have this question, thanks @MaryGao !

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -1513,6 +1553,29 @@ Expected input body:
 {"property": <don't include this property>}
 ```
 
+### Models_Property_Types_ModelNullableProperty_get
```
timotheeguerin, line 1556: Do those test belong in the Model Properties scenario and not their coresponding container
- Collection
- Dictionary

weidongxu-microsoft:
> > > Can we remove the union part? It is not going to be a valid case in .Net and Java.
> > 
> > 
> > @pshao25 @weidongxu-microsoft Is this decision finalized that we don't support nullable model/array/dictionary in .Net or Java? Or this is just current status.
> 
> i also have this question, thanks @MaryGao !

Already replied in https://github.com/Azure/cadl-ranch/pull/234#issuecomment-1469059819
No such decision in Java (no idea why the idea).

packages/cadl-ranch-specs/http/models/property-types/main.tsp
```diff
@@ -170,16 +170,78 @@ model CollectionsModelProperty is ModelTemplate<InnerModel[]>;
...
+@operationGroup
+interface ModelNullableProperty {
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "For testing"
+  @scenario
+  @scenarioDoc("""
+Expected response body::
+  ```json
+  {
+    "id": 1,
+    "property": "foo"
```
weidongxu-microsoft, line 223: It is `ModelNullableProperty` test, do you want property="foo" or property=null?

MaryGao: I design the case the property could be string or null, so here in `get` method it would be a foo string, and in `patch` method it would be a explicit null value.

.changeset/real-pets-mix.md
```diff
@@ -0,0 +1,5 @@
+---
+"@azure-tools/cadl-ranch-specs": minor
+---
+
+Add test cases for nullable values
```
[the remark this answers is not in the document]
MaryGao: Could you explore when it should be a patch and when should be a minor?

Here I add new subroute under property-types so I think it's minor change.

packages/cadl-ranch-specs/cadl-ranch-summary.md
MaryGao: I will keep the current position for these cases. Because the collection and dictionary cases are focusing on testing the response/request body as arrays or records. 

However in our nullable vaules we return normal models in which we contains some properties with nullable values.

timotheeguerin: Maybe I am misunderstanding but from what I see is the array iteself contains nullable value and the dictionary itself contains nullable values not the property that is of type array/dictionary being null which is what this spec is meant to test.

MaryGao: Got your point. I think we more likely request/response a payload in which its has a array property and in this container it may cover nullable item. I think in reality the more meaningful case would be the latter one. You could see the issue https://github.com/Azure/autorest.csharp/issues/3084.

```
alias nullableFloat = float32 | null;

model MyModel {
  listOfFloats?: nullableFloat[];   
}
```

This is why I design this case as property not array or dictionary. But we could add these cases in both position. How do you think?

timotheeguerin: I totally understand that this is most likely what you’ll see in services but I wanted to keep things to the minimum required in Cadl ranch and in this case the mode property  bring no value.

op getNullableArray(): nullalbleFloat[] is i feel like what we really care to test here.

timotheeguerin: Unless I’m missing something and there is something actually different about it being in a property

MaryGao: Okay thanks for the explaination! I think it makes sense and I'll change them to collection/dictionary sections.

MaryGao: I adjust the cases:

- Arrays/nullable-float -> NullableFloat[]
- Dictionary/nullable-float -> Record<NullableFloat>
- PropertyTypes/nullable-property -> ModelWithNullableProperty

packages/cadl-ranch-specs/http/models/property-types/main.tsp
```diff
@@ -183,3 +183,50 @@ model NeverProperty is ModelTemplate<never>;
...
 interface Never extends ModelOperations<NeverProperty, "<don't include this property>"> {}
+
+// Test a model with a model property which could be nullable
+@doc("Model with nullable properties")
+model ModelWithNullableProperty {
+  @doc("id property")
+  id: string;
+
+  @doc("nullable property")
+  property: string | null;
```
timotheeguerin, line 194: Sorry to bring new things but do we only care to test string or should we have a larger coverage of types that can be `null` as property.
i.e. Should we have a dedicated nullable model property spec that is pretty much like the optional model property spec but with `| null` instead of `?:`

packages/cadl-ranch-specs/http/models/property-types/main.tsp
```diff
@@ -183,3 +183,50 @@ model NeverProperty is ModelTemplate<never>;
...
+  ```json
+  {
+    "id": 1,
+    "property": "foo"
+  }
+  ```
+""")
+  @get
+  @doc("Get call")
+  get(): ModelWithNullableProperty;
```
timotheeguerin, line 214: Should we maybe not test the case where the service returns `null` instead(Or maybe both)

MaryGao: Similar optional property spec I keep both the getAll and getDefault. The first would return non-null value the latter would be a null.

.changeset/real-pets-mix.md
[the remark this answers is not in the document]
MaryGao: @timotheeguerin Do you think this is a patch or minor?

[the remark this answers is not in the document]
timotheeguerin: Don't have a super strong opinion but until we are in v1 we can follow the following rule:
- patch: new spec or fix
- minor: If there is something that actually change the flow(Breaking changes)
- NO Major
After release of dpg and we have a good set of spec we can move to semantic versioning so any new spec will increase minor

packages/cadl-ranch-specs/http/models/property-types/main.tsp
timotheeguerin: awesome!

packages/cadl-ranch-specs/http/models/property-nullable/mockapi.ts
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * Return the get and put operations
+ * @param route The route within /models/properties/optional/all/ for your function.
+ * @param value The value you are expecting and will return
+ */
+function createMockApis(route: string, value: any): MockApiGetPut {
+  const url = `/models/properties/nullable/${route}`;
+  const allUrl = `${url}/all`;
+  const defaultUrl = `${url}/default`;
```
timotheeguerin, line 20: Would just say here that `default` doesn't make that much sense for nullable as null is a value and we are not involving the default values.

Maybe name `withNull` and `withoutNull` or anything else you feel is more descriptive of the fact that we are testing sending null or not.

timotheeguerin (approved):
Looks great! made a last comment on the naming of the default operation, approving feel free to merge after resolving that!

PR #245, opened 2023-03-24
have lro operation return model

iscai-msft:

### DIFF
--- a/packages/cadl-ranch-specs/http/lro/lro-basic/main.tsp
+++ b/packages/cadl-ranch-specs/http/lro/lro-basic/main.tsp
@@ -9,22 +9,35 @@ using Azure.Core;
 @doc("All the basic lro test shall be added in this file")
 namespace Azure.Lro;
 
+@doc("Details about a user.")
+model User {
+  @doc("The name of user.")
+  name: string;
+}
+
 #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "For testing"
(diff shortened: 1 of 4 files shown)


timotheeguerin (approved):

packages/cadl-ranch-specs/http/lro/lro-basic/main.tsp
```diff
@@ -9,22 +9,35 @@ using Azure.Core;
...
   @pollingOperation(PollingSuccess.polling)
   @put
-  create(): string;
+  create(): User;
 
   @doc("The polling url")
   @route("/polling")
   @get
-  polling(): string;
+  polling(): User;
```
weidongxu-microsoft, line 38: Not sure whether polling needs a `status` property.

iscai-msft: i think python's core will look at "Succeeded". I'm going to keep it in for now to unblock people trying to move to latest `getLroMetadata`, definitely down to revisit later. thanks!

weidongxu-microsoft: OK for me.

Java will need to convert it to User model (as signature is `PollerFlux<User, User>`). And as "name" is required in User, some place would fail (in the case response only have `status`.
Not exactly same issue as my original comment.

PR #247, opened 2023-03-28
Add spread test cases

MaryGao:


chunyu3 (commented):
Can we add one more scenario:  Spread an Alias which contains model as property?
```
model Thing {
  @doc("name of the Thing")
  name: string;
  @doc("age of the Thing")
  age: int32;
}

alias ThingAliasWithModel = {
  @path id: string;
  ...Thing;
  @header top: int32;
};

op spreadAliasWithModel( ...ThingAliasWithModel ): OKResponse;
```

packages/cadl-ranch-specs/http/spread/main.tsp
```diff
@@ -0,0 +1,158 @@
...
+  name: string;
+}
+
+alias AliasNestedSpreadParameter = {
+  @path
+  id: string;
+
+  @header
+  `x-ms-test-header`: string;
+  ...ModelParameter;
```
MaryGao, line 142: @chunyu3 Case is updated here

packages/cadl-ranch-specs/http/spread/main.tsp
```diff
@@ -0,0 +1,158 @@
...
+
+using TypeSpec.Http;
+
+@doc("Test for the spread operator.")
+@supportedBy("dpg")
+@scenarioService("/spread")
+namespace Spread;
+
+@doc("This is a simple model.")
+model ModelBodyParameter {
```
weidongxu-microsoft, line 12: why is it not alias?

MaryGao:
@iscai-msft @timotheeguerin @weidongxu-microsoft Could you help review these cases?

packages/cadl-ranch-specs/http/spread/main.tsp
```diff
@@ -0,0 +1,158 @@
...
+@put
+op spreadModelAsRequestParameter(...ModelRequestParamter): OkResponse;
+
+alias AliasBodyParameter = {
+  name: string;
+};
+
+@scenario
+@scenarioDoc("""
+Test case for spread alias.
```
weidongxu-microsoft, line 60: Do we need to say that no model should be generated?

packages/cadl-ranch-specs/http/spread/main.tsp
```diff
@@ -0,0 +1,158 @@
...
+namespace Spread;
+
+@doc("This is a simple model.")
+model ModelBodyParameter {
+  name: string;
+}
+
+@scenario
+@scenarioDoc("""
+Test case for spread named model. Should generate request body model named ModelBodyParameter.
```
weidongxu-microsoft, line 18: Just a confirmation, we expect `ModelBodyParameter` model, as when it spread, it does not spread a alias, right?

packages/cadl-ranch-specs/http/spread/main.tsp
```diff
@@ -0,0 +1,158 @@
...
+  id: string;
+
+  @header
+  `x-ms-test-header`: string;
+  name: string;
+}
+
+@scenario
+@scenarioDoc("""
+Test case for spread named model with path and header parameter.
```
weidongxu-microsoft, line 41: What model do we expect here?

weidongxu-microsoft (commented):
Java should have supported the spread. Not sure about .NET

packages/cadl-ranch-specs/http/spread/main.tsp
```diff
@@ -0,0 +1,158 @@
...
+  id: string;
+
+  @header
+  `x-ms-test-header`: string;
+  ...ModelParameter;
+};
+
+@scenario
+@scenarioDoc("""
+Test case for spread alias with nested spread model as body parameter.
```
weidongxu-microsoft, line 147: Again, we do not expect any model, right??

packages/cadl-ranch-specs/http/spread/main.tsp
iscai-msft: yeah for `model`s we should generate them, even if they're spread. we should never generate an alias as a model

packages/cadl-ranch-specs/http/spread/main.tsp
iscai-msft: i agree, I would not include quantum models yet, there seems to be continued discussions over them, and don't think we should block this PR on this. Instead, you can make this an alias

weidongxu-microsoft: OK, in this case, there maybe no model (as it is path and header), whether it is alias or named model.

weidongxu-microsoft (approved):

packages/cadl-ranch-specs/http/spread/main.tsp
MaryGao: @lmazuel I'd like to hear your feedback here. For all spread cases, we are pretty sure what kind of http payload would be look like. But the questions for whether we are supposed to generate a model or should not generate any model currently it may be lack of consistency across language. Do you think  we should keep consistent and we need to specify in cadl-ranch?

MaryGao:
> Java should have supported the spread. Not sure about .NET

@chunyu3 Could you comfirm that these cases are supported in .NET?

packages/cadl-ranch-specs/http/spread/main.tsp
MaryGao: @iscai-msft We already have another case where it is an [alias ](https://github.com/Azure/cadl-ranch/blob/1f2b61d06a47d5e9b92508282062c9adcfb87b19/packages/cadl-ranch-specs/http/spread/main.tsp#L71-L79).

Not sure you could share some links for quantum models discusssion, before that it's better to comment this case out, right?

packages/cadl-ranch-specs/http/spread/main.tsp
MaryGao: For alias JS would not generate any named model, Could other language representatives confirm the behavior?

/cc @iscai-msft  @chunyu3

weidongxu-microsoft: no model for java.

packages/cadl-ranch-specs/http/spread/main.tsp
lmazuel: We should never generate a public model from an alias. There is no exception to that rule, as alias is just a way to simplify some syntax, and removing the alias later is NOT a breaking change in TypeSpec. If you rely on it for model generation, bad things will happen.

I can understand that for serialization purposes some language needs to generate a private internal model. Since it would be private, the name of that model doesn't matter.

packages/cadl-ranch-specs/http/spread/main.tsp
lmazuel: Like last comment, no-one should ever generate a model from an alias.

packages/cadl-ranch-specs/http/spread/main.tsp
```diff
@@ -0,0 +1,158 @@
...
+Test case for spread named model. Should generate request body model named ModelBodyParameter.
+
+Expected request body:
+```json
+{ "name": "foo" }
+```
+""")
+@route("/model/request-body")
+@put
+op spreadModelAsRequestBody(...ModelBodyParameter): OkResponse;
```
lmazuel, line 27: This is a valid scenario.

Must generate an operation and a model:
```
void spreadModelAsRequestBody(model_body_parameter: ModelBodyParameter)
```

Note that we don't have a parameter name in this scenario, I suggest it's guessed from the model name. See also:
https://github.com/Azure/typespec-azure/issues/2360

packages/cadl-ranch-specs/http/spread/main.tsp
lmazuel: @MaryGao I would remove any quantum model for now from this file.

packages/cadl-ranch-specs/http/spread/main.tsp
```diff
@@ -0,0 +1,158 @@
...
+Expected path parameter: id="1"
+Expected header parameter: x-ms-test-header="bar"
+Expected request body:
+```json
+{ "name": "foo" }
+```
+""")
+@route("/alias/request-parameter/{id}")
+@put
+op spreadAliasAsRequestParameter(...AliasRequestParameter): OkResponse;
```
lmazuel, line 93: No quantum model in this file for now. If necessary, let's call the folder "basic_spread"

lmazuel, line 93: This scenario is ok, this is not a quantum model since this is an alias.

Should generate something like:
```
void spreadAliasAsRequestParameter(id: string, name: string)
```

packages/cadl-ranch-specs/http/spread/main.tsp
```diff
@@ -0,0 +1,158 @@
...
+  "prop3":"foo3",
+  "prop4":"foo4",
+  "prop5":"foo5",
+  "prop6":"foo6"
+}
+```
+""")
+@route("/alias/multiple-parameters/{id}")
+@put
+op spreadAliasWithMultipleParameters(...AliasMultipleRequestParameters): OkResponse;
```
lmazuel, line 129: Feels C# specific, but that's ok, this is a fair test to have to confirm what impacts for operation with a lot of parameters.

In spirit, should generate something like:
```
void spreadAliasWithMultipleParameters(id: string, x_ms_test_header: string, prop1: string, prop2: string, prop3: string, prop4: string, prop5: string, prop6: string)
```

packages/cadl-ranch-specs/http/spread/main.tsp
chunyu3: For alias which only have one model as body, other properties are head or query or path, when spread alias in operation, .net will generate model for body. so the convenience method look like:
```
void spreadAliasNestedSpreadModelParameter(id: string, x_ms_test_header: string, body: ModelParameter)
```

lmazuel: @chunyu3 I realize there is good point it on the order of processing from TSP on spread VS aliases. Let me ask the question:
https://teams.microsoft.com/l/message/19:16f8f1ed70d044139be153707bdaee0e@thread.skype/1680633341627?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1680633341627&teamName=Azure%20SDK&channelName=TypeSpec%20(Cadl)%20Engineering&createdTime=1680633341627

lmazuel: @chunyu3 based on the discussion with @timotheeguerin , your convenience method is incorrect here, you should not have a model as a parameter.

lmazuel: After a full day of discussion, let's remove this scenario for now, as we have no clear conscencus on what it should do

MaryGao: Removed.

packages/cadl-ranch-specs/http/spread/main.tsp
MaryGao: Removed.

MaryGao:
> Can we add one more scenario: Spread an Alias which contains model as property?
> 
> ```
> model Thing {
>   @doc("name of the Thing")
>   name: string;
>   @doc("age of the Thing")
>   age: int32;
> }
> 
> alias ThingAliasWithModel = {
>   @path id: string;
>   ...Thing;
>   @header top: int32;
> };
> 
> op spreadAliasWithModel( ...ThingAliasWithModel ): OKResponse;
> ```

Pls notice the case is not updated because the discussion is not finalized https://github.com/Azure/cadl-ranch/pull/247#discussion_r1157859847

packages/cadl-ranch-specs/http/spread/main.tsp
```diff
@@ -0,0 +1,131 @@
+import "@typespec/http";
+import "@azure-tools/cadl-ranch-expect";
+
+using TypeSpec.Http;
+
+@doc("Test for the spread operator.")
+@supportedBy("dpg")
+@scenarioService("/spread")
+namespace Spread;
```
iscai-msft, line 9: can we have a subnamespace for "Alias" and one for "Model"?

iscai-msft: then we can change `spreadModelAsRequestBody` to just `requestBody` inside the `Spread.Model` subnamespace (we can decorate this namespace with an operation group decorator)

MaryGao: Thanks for the suggestion! I reorg these cases into `Model` and `Alias` subnamespace. Please note I also move these cases into `Parameters` top-level namespace because these cases are all for preparing request paramters.

packages/cadl-ranch-specs/http/parameters/spread/main.tsp
```diff
@@ -0,0 +1,141 @@
...
+  @put
+  op spreadAsRequestParameter(...RequestParameter): NoContentResponse;
+
+  alias MultipleRequestParameters = {
+    @path
+    id: string;
+
+    @header
+    `x-ms-test-header`: string;
+    prop1: string;
```
iscai-msft, line 105: i don't think we need this many entries in the alias, we can have one path, one header, and one `prop`

iscai-msft (approved):

PR #257, opened 2023-04-12
fix mock api issue

tadelesh:
fix mock api issue for nullable array (cc @MaryGao) and client path parameter.


MaryGao (approved):
Thanks for the fix! LGTM

tadelesh:
Add fix for nullable collection model definition problem. cc: @MaryGao

PR #259, opened 2023-04-12
fix collection format

iscai-msft:

### DIFF
--- a/packages/cadl-ranch-specs/http/parameters/collection-format/main.tsp
+++ b/packages/cadl-ranch-specs/http/parameters/collection-format/main.tsp
@@ -47,7 +47,7 @@ namespace Header {
   @scenarioDoc("""
 This test is testing sending a csv collection format array header parameters
 """)
-  @route("/csvHeader")
+  @route("/csv")
   op csv(
     @doc("Possible values for colors are [blue,red,green]")
     @header({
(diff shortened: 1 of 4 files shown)


packages/cadl-ranch-specs/http/parameters/collection-format/main.tsp
```diff
@@ -47,7 +47,7 @@ namespace Header {
   @scenarioDoc("""
 This test is testing sending a csv collection format array header parameters
 """)
-  @route("/csvHeader")
+  @route("/csv")
```
iscai-msft, line 50: @timotheeguerin is there any way that we can validate every route in the main tsp is covered in mockapi.ts? Right now we check based off of scenario name, could we add a step verifying the routes match?

packages/cadl-ranch-specs/http/parameters/collection-format/mockapi.ts
```diff
@@ -7,7 +7,7 @@ Scenarios.Parameters_CollectionFormat_Query_multi = passOnSuccess(
   mockapi.get("/parameters/collection-format/query/multi", (req) => {
     req.expect.containsQueryParam("colors", ["blue", "red", "green"], "multi");
     return {
-      status: 200,
+      status: 204,
```
iscai-msft, line 10: could we verify that the mockapis return the correct status code? i.e. 204 if they're returning NoContentResponse? This is more of a stretch ask, there might be some difficulties in adding this that i'm not anticipating though. thanks!

timotheeguerin: we could maybe but thats probably a lot of work so not sure its worth it

timotheeguerin (approved):

PR #261, opened 2023-04-13
resiliency tests update

iscai-msft:
elevating @timotheeguerin comment

The reason is because this scenario is covering the evolution of a service. 
old.tsp is how it was written and implemented at a certain point in time(for example 2022-01-01)
main.tsp is how the it was updated  at a later point in time (e.g. 2023-01-01)

The reason we have all those values is because we want to test the following cases:

- Client V1 (This is the client generated with old.tsp)
  -  `service: v1, api-version v1` Should be able to call the service when it was at the v1 implementation 
  -  `service: v2, api-version v1` Should still be able to call the service when it was at the v2 implementation by using the v1 api version
- Client V2 (This is the client generated with new.tsp)
  -  `service: v2, api-version v1` should be able to call the v2 implementation of the service with api-version v1
  -  `service: v2, api-version v2` should be able to call the v2 implementation of the service with api-version v2


packages/cadl-ranch-specs/http/resiliency/srv-driven-1/main.tsp
```diff
@@ -1,100 +1,148 @@
...
+  """)
+@client({
+  name: "ResiliencyServiceDrivenClient",
+})
+@service
+@server(
+  "http://localhost:3000/resiliency/service-driven/client:v1/service:{serviceDeploymentVersion}/api-version:{apiVersion}",
+  "Testserver endpoint",
+  {
+    @doc("Pass in either 'v1' or 'v2'. This represents a version of the service deployment in history. 'v1' is for the deployment when the service had only one api version. 'v2' is for the deployment when the service had api-versions 'v1' and 'v2'.")
```
iscai-msft, line 22: should we use `v1` and `v2` for the versioning of the client, the service deployment and api version? @lmazuel is also wondering if we want to do more accurate semantic versioning / using dates for api version versioning. I'm ok with any versioning with de for `api-version`, but I'm thinking if we want to do `old` and `new` for service deployment version and client

packages/cadl-ranch-specs/http/resiliency/srv-driven-1/main.tsp
```diff
@@ -1,100 +1,148 @@
...
+  With the above two calls, we test the following configurations from this service spec:
+  - A client generated from the first service spec can call the first deployment of a service with api version v1
+  - A client generated from the first service spec can call the second deployment of a service with api version v1
+
+  In the next service spec, we will test that we can grow this operation from accepting one content type to two.
+""")
+@route("/add-content-type")
+@doc("Test that currently accepts one content type. Will be updated in next spec to accept a new content type as well.")
+@post
+op addContentType(
```
iscai-msft, line 142: this test (esp if we want to grow up to take two different bodies as input) is part of a lot of the overload talks. I'm wondering if we want to hold off on this until later

packages/cadl-ranch-specs/http/resiliency/srv-driven-1/mockapi.ts
```diff
@@ -1,137 +1,30 @@
...
-    }
-  }),
-);
-
-/**
- * Initially the path exists but there is no delete method. After evolution this is a new method in a known path
- */
-Scenarios.DPGAddNewOperation = passOnSuccess(
-  mockapi.delete("/servicedriven/parameters", (req) => {
+Scenarios.Resiliency_ServiceDriven_V1_breakTheGlass = passOnSuccess(
```
iscai-msft, line 24: here i've added a 'break the glass' scenario that we want to test. Basically can a client call an endpoint on the service even if it's not a generated operation in the SDK. I'm using the "new operation" we added in the second deployment of the service for `api-version` `v2`. To me this is fair because that is a scenario for break the glass: "i have an old client but i want to call a new endpoint with a new api version". Thoughts?

packages/cadl-ranch-specs/http/resiliency/srv-driven/old.tsp
```diff
@@ -0,0 +1,148 @@
...
+  - A client generated from the first service spec can call the first deployment of a service with api version v1
+  - A client generated from the first service spec can call the second deployment of a service with api version v1
+
+  In the next service spec, we will test that we can grow this operation from accepting one content type to two.
+""")
+@route("/add-content-type")
+@doc("Test that currently accepts one content type. Will be updated in next spec to accept a new content type as well.")
+@post
+op addContentType(
+  @path contentTypePath: "json",
```
tadelesh, line 143: This path param is weird. I think for the first version, we should remove this path param. For the evolutional version, we shall define two operations with shared route (I don't know what is the final decision of the best practice for overload):
```
@scenario
@scenarioDoc("""
Operation only accept application/json content. After evolution, this operation could accept both application/json and image/jpeg.

Expected header: contentType=application/json
Expected request body:
```json
{
  "message": "hello"
}
Expected response body:
```json
{
  "message": "hello"
}
""")
@route("/contentType")
@post
op contentType(
  @body
  parameter: Message
): Message;

@scenario
@scenarioDoc("""
Operation only accept application/json content. After evolution, this operation could accept both application/json and image/jpeg.

Expected header: contentType=image/jpeg
Expected request body: binary date of the image
Expected response body:
```json
{
  "message": "hello"
}
""")
@route(
  "/contentType",
  {
    shared: true,
  }
)
@post
op contentTypeOfImage(
  @body
  parameter: bytes
): Message;
```

packages/cadl-ranch-specs/http/resiliency/srv-driven/main.tsp
```diff
@@ -0,0 +1,196 @@
...
+  2. Service deployment version: refers to a deployment version of the service. 'v1' represents the initial deployment of the service with a single api version. 'v2' represents the new deployment of a service with multiple api versions
+  3. Api version: The initial deployment of the service only supports api version 'v1'. The new deployment of the service supports api versions 'v1' and 'v2'.
+
+  With the above two calls, we test the following configurations from this service spec:
+  - A client generated from the first service spec can break the glass and call the second deployment of a service with api version v2
+  - A client generated from the second service spec can call the second deployment of a service with api version v2 with the updated changes
+
+  Tests that we can grow up by adding an operation.
+  """)
+@added(Versions.v2)
```
tadelesh, line 192: Do we need another test case of "add new operation of different method to the existed path"?

packages/cadl-ranch-specs/http/resiliency/srv-driven/main.tsp
```diff
@@ -0,0 +1,196 @@
...
+  2. Service deployment version: refers to a deployment version of the service. 'v1' represents the initial deployment of the service with a single api version. 'v2' represents the new deployment of a service with multiple api versions
+  3. Api version: The initial deployment of the service only supports api version 'v1'. The new deployment of the service supports api versions 'v1' and 'v2'.
+
+  With the above two calls, we test the following configurations from this service spec:
+  - A client generated from the second service spec can call the second deployment of a service with api version v1
+  - A client generated from the second service spec can call the second deployment of a service with api version v2 with the updated changes
+
+  Tests that we can grow up an operation from accepting one optional parameter to accepting two optional parameters.
+  """)
+  @route("/from-one-optional")
```
tadelesh, line 126: Do we need another test case of "add a new optional parameter to the operation with one required and one optional parameter"?

packages/cadl-ranch-specs/http/resiliency/srv-driven/main.tsp
```diff
@@ -0,0 +1,196 @@
...
+  We test the following configurations from this service spec:
+  - A client generated from the second service spec can call the second deployment of a service with api version v1
+  - A client generated from the second service spec can call the second deployment of a service with api version v2
+  """)
+@client({
+  name: "ResiliencyServiceDrivenClient",
+})
+@service
+@server(
+  "http://localhost:3000/resiliency/service-driven/client:v2/service:{serviceDeploymentVersion}/api-version:{apiVersion}",
```
tadelesh, line 28: I don't quite get the reason of separating the client and service version. IMO, the mock api with two api versions could already mock the service with two versions. Mock server should not be responsible for testing the behavior of the clients.

weidongxu-microsoft: My guess is for mockapi and dashboard to capture v1 and v2 separately (unsure).

Ideally, I would say we'd better to use api-version in query, as what Azure does, but here maybe it is more important to get mockapi to work.

packages/cadl-ranch-specs/http/resiliency/srv-driven/main.tsp
```diff
@@ -0,0 +1,196 @@
...
+@route("/add-content-type")
+@doc("Operation that we can grow up an operation from accepting one content type to two.")
+@post
+op addContentType(
+  @header `content-type`: "application/json" | "image/jpeg",
+
+  @doc("I am a body parameter. My only valid JSON entry is { url: \"http://example.org/myimage.jpeg\" }")
+  @body
+  parameter: PostInput
+): NoContentResponse;
```
weidongxu-microsoft, line 173: I seems not able to find any typespec sematic saying that `"image/jpeg"` is added in v2?

I am not against the case, but feeling this may be not mature enough. (i.e. maybe sharedRoute can handle this better, via adding another operation).

weidongxu-microsoft (commented):
Does ideally we'd like codegen generate 2 SDK, one from "old.tsp", one from "main.tsp", and use language tool to verify that new/main SDK is backward compatible to old SDK?

And runtime test would be done via new/main SDK (with "v1" and "v2")?

packages/cadl-ranch-specs/http/resiliency/srv-driven/main.tsp
timotheeguerin: we removed that one because didn't see this being a different use case, do you have an example were this could be treated differently?

packages/cadl-ranch-specs/http/resiliency/srv-driven/main.tsp
timotheeguerin: @weidongxu-microsoft for you comment on the expected behavior of this spec, the most important is the runtime test as described above. Checking there is no breaking change in the generated code between old and main is an added benefit you can also check but isn't the main purpose of this.

packages/cadl-ranch-specs/http/resiliency/srv-driven/main.tsp
timotheeguerin: yeah you are right, maybe we should omit this test until we have a better understanding.

packages/cadl-ranch-specs/http/resiliency/srv-driven/main.tsp
iscai-msft: we don't have one to an existing path, i feel it's not super necessary, do you have an example of how we would generate differently? thanks!

tadelesh: make sense.

packages/cadl-ranch-specs/http/resiliency/srv-driven/main.tsp
tadelesh: I see. It seems to be useful to show the result in the dashboard.

PR #262, opened 2023-04-14
Add HTTP auth tests

lmazuel:

Used in EventGrid for auth "Authorization: SharedAccessKey mykey".
See also https://github.com/Azure/typespec-azure/issues/2706


packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -221,6 +221,27 @@ Expect error code 403 and error body:
 }
 ```
 
+### Authentication_Http_valid
```
timotheeguerin, line 224: This should probably be Authentication.Http.sharedAccessKey (or without http)

lmazuel: Technically, I used "SharedAccessKey" since this is the actual value that EventGrid needs, but the test is designed to just confirm that the value passed as "scheme" is handled correctly. IOW, if I were to write "CogSevKey", this should work as well. It's why I tried to avid putting SharedAccessKey into the name.

Maybe Authentication.Http.CustomScheme ?

timotheeguerin: What do we expect the client to generate 

new SharedAcessKeyAuth("…")

or 

new HttpAuth("SharedAccessKey", "…")

for the first one I think we should have a test dedicated to it. For the 2Ms I guess it’a ok

lmazuel: I expect to generate something like:
client = client(endpoint, AzureKeyCredential("mykey"))

The fact that the scheme is "SharedAccessKey" is a hidden detail

timotheeguerin: I guess the question then is lets assume we have this other http scheme called `MasterKey`, would the Client generated code automatically inject that.

Basically do we expect this:
```
# In generted client.py
.headers["Authentication] = httpScheme + " " + value;
# where httpScheme would be whatever is the value in the spec
```
and the user would use that
```
client = client(endpoint, AzureKeyCredential("mykey"))
```


What do you expect those 2 things too look like if the scheme was `Basic`, where the value should be `username:password`

lmazuel: If it was `Basic`, I would not generate that, I'll definitely take a different credentials. The same way when it's "Bearer", we take a TokenCredential.

I guess it means there are some special values mapped to special types. I guess what I'm doing here, is to say that the fallback for unknown scheme is AzureKeyCredential.

timotheeguerin: I see, then yeah having this custom scheme naming then should be good. That would assume we can have this structure right:
```
Authentication
  Http:
     Bearer
     Basic
     Custom
  OAuth2
  ApiKey
    +

**Review by srnagar — APPROVED** [review]

File: pnpm-lock.yaml
Comment (timotheeguerin) [inline pnpm-lock.yaml:1]: do we need change to this file here, this feels like it might be outdated merge issue

File: packages/cadl-ranch-specs/cadl-ranch-summary.md [hunk 2]
```diff
@@ -21,6 +21,24 @@ Expect error code 403 and error body:
...
 
+### Authentication_Http_Custom_invalid
+
+- Endpoint: `get /authentication/http/custom/invalid`
+
+Expect error code 403 and error body:
+
+```json
+{
+  "error": "invalid-api-key"
```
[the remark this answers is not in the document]
timotheeguerin: For the sake of the test it doesn't really matter, the simpler the better

timotheeguerin (approved):

iscai-msft (approved):

packages/cadl-ranch-specs/http/authentication/http/custom/main.tsp
```diff
@@ -0,0 +1,39 @@
...
+import "@azure-tools/cadl-ranch-expect";
+
+using TypeSpec.Http;
+
+@scenarioService("/authentication/http/custom")
+@doc("Illustrates clients generated with generic HTTP auth.")
+@useAuth({
+  type: AuthType.http,
+  scheme: "SharedAccessKey",
+})
```
weidongxu-microsoft, line 11: It seems hard to understand from tsp, that this is to create a `SharedAccessKey <value>` on `Authorization` header.
(maybe the `Authorization` is kind of inferred, as `http` type is mostly on it; but `scheme` as prefix is quite confusing to anyone out of codegen)

PR #263, opened 2023-04-17
Add other types of collection format array query testcase

qiaozha:


packages/cadl-ranch-api/src/types.ts
```diff
@@ -52,3 +52,11 @@ export interface MockResponseBody {
   contentType: string;
   rawContent: string | undefined;
 }
+
+export enum CollectionFormat {
```
timotheeguerin, line 56: can we leave this as a union and not an enum. string enums are often just more annoying to deal with in TS.

qiaozha: okay, changed into union.

timotheeguerin (approved):

PR #273, opened 2023-04-20
Add client format on duration tests

iscai-msft:

resolve: https://github.com/Azure/cadl-ranch/issues/288


tadelesh:
@iscai-msft will we release a new tcgc version to replace the dev version in this PR? or maybe this PR need to wait `@dataFormat`(https://github.com/microsoft/typespec/pull/1796) replace the `@clientFormat`?

iscai-msft:
@tadelesh i think we're good with using dev versions in cadl-ranch, esp in this case bc everyone needs to upgrade to support `@clientFormat` with seconds before the next public release of tsp (to unblock mvad)

iscai-msft:
also @tadelesh can't approve it because i'm the original author, but this PR lgtm, thanks!

timotheeguerin (approved):

iscai-msft:
@srnagar @m-nash can one of you guys take a look? need a strongly typed language approval. thanks!

packages/cadl-ranch-specs/http/azure/client-generator-core/client-format/duration/main.tsp
```diff
@@ -0,0 +1,90 @@
...
+
+@global.Azure.ClientGenerator.Core.operationGroup
+@route("/property")
+namespace Property {
+  model ISO8601DurationProperty {
+    value: duration;
+  }
+
+  model SecondsDurationProperty {
+    @global.Azure.ClientGenerator.Core.clientFormat("seconds")
```
m-nash, line 48: didn't we agree that we need 2 parameters here?  `units` and `format`.  I think there are two real examples both wanting duration in seconds as the units but one wanting a double and one wanting an int?

timotheeguerin: or maybe have the format be `"seconds-int"`, `"seconds-float64"`

m-nash: That could work @timotheeguerin, typically would lean towards 2 params here but the combined single param is less work we can do that.

@iscai-msft I would lean towards tcgc parsing / splitting this into two parameters that get passed to emitters.

iscai-msft: @m-nash @joheredi is working on that right now, we do have 2 options here: we can wait for him to be done, or we can go ahead and release the test now. the more i think though i think you're right, we should wait until all of the support is into tcgc

packages/cadl-ranch-specs/http/azure/client-generator-core/client-format/duration/main.tsp
```diff
@@ -0,0 +1,90 @@
...
+    input: duration
+  ): NoContentResponse;
+
+  @route("/seconds")
+  @scenario
+  @scenarioDoc("""
+  Test seconds client format for a duration parameter.
+  Expected query parameter `input=35.6`
+  """)
+  op seconds(
```
srnagar, line 33: We'll need multiple routes for different formats - int and double.

packages/cadl-ranch-specs/http/azure/client-generator-core/client-format/duration/main.tsp
```diff
@@ -0,0 +1,90 @@
...
+import "@azure-tools/cadl-ranch-expect";
+import "@azure-tools/typespec-client-generator-core";
+
+using TypeSpec.Http;
+using Azure.ClientGenerator.Core;
+
+@doc("Test for clientFormat decorator on duration.")
+@supportedBy("dpg")
+@scenarioService("/azure/client-generator-core/client-format/duration")
+namespace _Specs_.Azure.ClientGenerator.Core.ClientFormat.Duration;
```
srnagar, line 11: Just curious why we have _Specs_ as part of the namespace. It's in some tests and not in [others](https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/authentication/api-key/main.tsp#L9).

packages/cadl-ranch-specs/http/azure/client-generator-core/client-format/duration/main.tsp
tadelesh: @iscai-msft after the tcgc update, I could help to update the test then.

packages/cadl-ranch-specs/http/azure/client-generator-core/client-format/duration/main.tsp
tadelesh: Will add the test after tcgc update.

packages/cadl-ranch-specs/http/azure/client-generator-core/client-format/duration/main.tsp
tadelesh: The namespace for cadl ranch test is always corresponding to the test folder structure. For test under `azure`, the folder structure is always the same as the core lib namespace segments, so I guess `_Specs_` is added to prevent any possible collision.

timotheeguerin: yeah the problem is if we keep it then it conflict with the Azure.Core or other azure library namespace and then having `@scenarioService` will make it think all th eoperations inside of Azure.Core should be scenarios

tadelesh:
Replaced by #316.

PR #211, opened 2023-02-14
Add body item convert to fix datetime compare issue

pshao25:
# Cadl Ranch Contribution Checklist:

- [ ] I have written a [scenario spec](../docs/writing-scenario-spec.md)
- [ ] I have **meaningful** `@scenario` names. Someone can look at the list of scenarios and understand what I'm covering.
- [ ] I have written a [mock API](../docs/writing-mock-apis.md)
- [ ] I have used `@scenarioDoc`s for extra scenario description and to tell people **how to pass** my mock api check.

PR #274, opened 2023-04-21
Fix datetime issue in dictionary

pshao25:
`bodyEquals` just compares the values by taking them as literal strings. But if the value is datetime, this does not make sense, because the datetime could have different formats. Here `coerceBodyEquals` compares datetimes by first changing them into ISO format. We have same fix in the test server.


packages/cadl-ranch-specs/http/dictionary/mockapi.ts
```diff
@@ -22,7 +22,7 @@ function createModelMockApis(route: string, value: any): MockApiGetPut {
       };
     }),
     put: mockapi.put(url, (req) => {
-      req.expect.bodyEquals(value);
+      req.expect.coercedBodyEquals(value);
```
lmazuel, line 25: Please explain what you are fixing, your PR has no context

pshao25: `bodyEquals` just compares the values by taking them as literal strings. But if the value is datetime, this does not make sense, because the datetime could have different formats. Here `coerceBodyEquals` compares datetimes by first changing them into ISO format. We have same fix in the test server.

And we already have several fixes before.
This is for first solution: https://github.com/Azure/cadl-ranch/pull/91, https://github.com/Azure/cadl-ranch/pull/140
This is for second solution: https://github.com/Azure/cadl-ranch/pull/211

I just thought @timotheeguerin and @iscai-msft fully understand what I'm doing here, so I didn't put the description. :) Updated now.

PR #284, opened 2023-04-24
refine internal test

tadelesh:

Refine internal test for transitive closure.


packages/cadl-ranch-specs/http/azure/client-generator-core/internal/main.tsp
```diff
@@ -10,20 +10,35 @@ using Azure.ClientGenerator.Core;
 @scenarioService("/azure/client-generator-core/internal")
 namespace _Specs_.Azure.ClientGenerator.Core.Internal;
 
-@doc("This is an internal model.")
-model InternalModel {
+@doc("This is a model.")
+model Model {
```
timotheeguerin, line 14: Should we not have 2 different models to test the internal case? Wouldn't Model get generated here because `get` is not internal and then defeat the point of `getInternal`

timotheeguerin: oh I see, this is to test it does actually get generated despite being internal somewhere else

timotheeguerin: Ok so I think we might want some different scenarios here:
- Public model
- Shared model
- Internal model


So i'd say lets organize the scenarios this way:

```tsp
@scenario
op publicOnly(): PublicModel;

@scenario
op internalOnly(): InternalOnly;

// This scenario is both operations
@scenario
interface SharedModels {
  public(): SharedModel;
  internal(): SharedModel;
}
```

and so we have 4 endpoints but the last 2 are under the same scenario

timotheeguerin: (feel free to use different names if that make more sense than public)

tadelesh: Good suggestion. `public` is good, easy to understand. I've updated the tsp and mockapi.

packages/cadl-ranch-specs/http/azure/client-generator-core/internal/main.tsp
```diff
@@ -32,30 +36,56 @@ Expected response body:
...
-@post
+@route("/internal")
+@get
 @global.Azure.ClientGenerator.Core.internal
-op postInternal(@body body: ModelOnlyUsedByInternalOperation): ModelOnlyUsedByInternalOperation;
+op internalOnly(@query name: string): PublicModel;
+
+@route("/shared")
+interface Shared {
+  @scenario
```
timotheeguerin, line 61: Put the scenario on the interface so both operations belong to the same scenario

packages/cadl-ranch-specs/http/azure/client-generator-core/internal/mockapi.ts
```diff
@@ -1,31 +1,22 @@
...
-    }
-    return {
-      status: 200,
-      body: json({ id: 1, name: req.body["name"] }),
-    };
-  }),
-);
+Scenarios.Azure_ClientGenerator_Core_Internal_publicOnly = passOnSuccess(createInternalMockApis("public"));
+Scenarios.Azure_ClientGenerator_Core_Internal_internalOnly = passOnSuccess(createInternalMockApis("internal"));
+Scenarios.Azure_ClientGenerator_Core_Internal_Shared_public = passOnSuccess(createInternalMockApis("shared/public"));
```
timotheeguerin, line 21: Those 2 should be under the same scenario you can pass a list to `passOnSuccess`

packages/cadl-ranch-specs/http/azure/client-generator-core/internal/main.tsp
```diff
@@ -32,30 +36,45 @@ Expected response body:
...
-  "id": 1,
   "name": <any string>
 }
 ```
 """)
-@route("/postInternal")
-@post
-@global.Azure.ClientGenerator.Core.internal
-op postInternal(@body body: ModelOnlyUsedByInternalOperation): ModelOnlyUsedByInternalOperation;
+interface Shared {
```
timotheeguerin, line 71:
```suggestion
@operationGroup
interface Shared {
```

I think you need that so those operations don't end up at the root right?

timotheeguerin (approved):
One little comment left about operationGroup but good to merge after

PR #297, opened 2023-04-26
add record of unknown type test

tadelesh:

resolve: https://github.com/Azure/cadl-ranch/issues/251


tadelesh:
resolve: https://github.com/Azure/cadl-ranch/issues/251

packages/cadl-ranch-specs/http/type/property/value-types/mockapi.ts
```diff
@@ -91,3 +91,14 @@ Scenarios.Type_Property_ValueTypes_DictionaryString_put = passOnSuccess(dictiona
 const neverMock = createMockApis("never", undefined);
 Scenarios.Type_Property_ValueTypes_Never_get = passOnSuccess(neverMock.get);
 Scenarios.Type_Property_ValueTypes_Never_put = passOnSuccess(neverMock.put);
+
+const recordUnknownMock = createMockApis("record/unknown", {
```
timotheeguerin, line 95: I think this test should go in the dictionary test, this is basically that.

tadelesh: Oh, I see. There already exists a test case of `Record<unknown>` under dictionary. I will close this PR.

tadelesh:
Currently, there is a lint rule that disallow `Record<unknown>`: https://github.com/Azure/typespec-azure/blob/8c23a8e9a2bd3fe7462560370f6e0ea15332bb72/packages/typespec-azure-core/src/rules/record-types.ts#L5. If we need to support any JSON object using `Record<unknown>`, we should remove such lint.

packages/cadl-ranch-specs/http/type/property/value-types/mockapi.ts
timotheeguerin: yeah it is a bit different but the goal of this test is to test different types for a model property
while the other is to test the different types for the value of a record.

So in this test we want to test one Record but we don't need to test every single record combination on top.
This shouldn't make any difference in the generation no?

tadelesh: Do you mean the we only need one Record<> test for type-property test and left tests for all kinds of types for Record value in dictionary test? If so, I'm OK to close this PR.

timotheeguerin: yep, I don't think it brings extra value

PR #298, opened 2023-04-26
Fix page key for value

pshao25:
According to its spec
```
@global.Azure.Core.pagedResult
model CustomPageModel<T> {
  @global.Azure.Core.items
  @doc("List of items.")
  items: T[];

  @global.Azure.Core.nextLink
  @doc("Link to fetch more items.")
  nextLink?: string;
}
```

The key to get items should be "items".

### DIFF
--- a/packages/cadl-ranch-specs/http/azure/core/basic/mockapi.ts
+++ b/packages/cadl-ranch-specs/http/azure/core/basic/mockapi.ts
@@ -77,7 +77,7 @@ Scenarios.Azure_Core_Basic_list = passOnSuccess(
 Scenarios.Azure_Core_Basic_listWithPage = passOnSuccess(
   mockapi.get("/azure/core/basic/page", (req) => {
     const responseBody = {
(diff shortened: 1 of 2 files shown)


packages/cadl-ranch-specs/http/azure/core/basic/mockapi.ts
```diff
@@ -77,7 +77,7 @@ Scenarios.Azure_Core_Basic_list = passOnSuccess(
 Scenarios.Azure_Core_Basic_listWithPage = passOnSuccess(
   mockapi.get("/azure/core/basic/page", (req) => {
     const responseBody = {
-      value: [validUser],
```
timotheeguerin, line 80: I don't think this one is correct, @tadelesh already fixed the `listWithCustomPage` operation in that PR.

https://github.com/Azure/cadl-ranch/pull/290/files

This operation use the Azure.Core standard `Page`

PR #321, opened 2023-05-09
tsp-bump-deps, include overrides

weidongxu-microsoft:
One comment is that codegen would wish the script has an option to add the "overrides", instead of modifying the existing "overrides". (it may get script more complex, as it would need to take different "overrides" format from different package manager, and it may need to merge the 2 list)

Test in Java (same script)
https://github.com/Azure/autorest.java/pull/2135
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2767191&view=logs&j=ca395085-040a-526b-2ce8-bdc85f692774&t=c310a698-906e-53a5-78f7-614142e7c740


packages/typespec-bump-deps/src/cli/cli.ts
```diff
@@ -27,7 +27,7 @@ async function main() {
   for (const packageJsonPath of packageJsonPaths) {
     const content = await readFile(packageJsonPath);
     const packageJson = JSON.parse(content.toString());
-    const depTypes = ["dependencies", "devDependencies", "peerDependencies"];
+    const depTypes = ["dependencies", "devDependencies", "peerDependencies", "overrides"];
```
MaryGao, line 30: Can we add the `overrides` section in package.json in this script if absent?

timotheeguerin (approved):

packages/typespec-bump-deps/src/cli/cli.ts
weidongxu-microsoft: I think we can do this, though probably with a bit more discussion with Isabella & Timothee on parameter.

I kind of think `typespec-bump-deps --add-rush-overrides <file-path>` may do.
Though you will likely need to call it twice, one on `package.json`s to modify the deps, then call it on `pnpm-config.json` with that option to add overrides.

For npm it maybe `typespec-bump-deps --add-npm-overrides <file-path>`. (I assume Java and .NET may use this)

Let us continue this discussion when we are decided that JS codegen would indeed use the overrides.

MaryGao: No, I don't need to call twice to upgrade them. I only need to call `typespec-bump-deps --add-rush-overrides <file-path>` and Rush would manage all rest of changes for me.

packages/typespec-bump-deps/src/cli/cli.ts
```diff
@@ -27,7 +27,7 @@ async function main() {
...
     for (const depType of depTypes) {
       const deps = packageJson[depType];
       if (deps === undefined) continue;
```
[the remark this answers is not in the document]
pshao25: This is our expected:
1. We bump the version of dependencies in the first package.json.
2. We need to add the overrides section in the second package.json which doesn't have any dependencies. And in that overrides section, there are the bumped dependencies. 

Is that doable?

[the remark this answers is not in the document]
weidongxu-microsoft: Current script won't.

Should be doable in script. But please confirm that this solution indeed work for .NET, before we actually implement it here.

packages/typespec-bump-deps/src/cli/cli.ts
MaryGao: `typespec-bump-deps --add-rush-overrides <file-path>` would be good to me. I would expect the following logic if enable `--add-rush-overrides` option:

1. Check if existing section `globalOverrides`; if yes, step 2; if no, adding one section;
2. For every `knownPackages` override its latest dev into `globalOverrides`

packages/typespec-bump-deps/src/cli/cli.ts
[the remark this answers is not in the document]
timotheeguerin: @pshao25 can you share exactly what state of the package.json worked for you. 

I am asking this because the script in its current state only change version of packages already in the package.json, it doesn't add anything. If we want to add new entries we need to make sure we can include the right set of dependencies.

@MaryGao same thing for `rush` could we get the exact value that is working for you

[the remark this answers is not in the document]
weidongxu-microsoft: I assume they want to add/merge all these libs
https://github.com/Azure/cadl-ranch/blob/781add01b1544fe5df6a428676b0887c879ead85/packages/typespec-bump-deps/src/cli/cli.ts#L5C1-L11

packages/typespec-bump-deps/src/cli/cli.ts
```diff
@@ -23,7 +23,12 @@ async function main() {
   console.log("The following is a mapping between packages and the versions we will update them to");
   // eslint-disable-next-line no-console
   console.log(packageToVersionRecord);
-  const packageJsonPaths = process.argv.slice(2);
+  const args = process.argv.slice(2);
+  const addRushOverridesArg = "--add-rush-overrides";
+  const addNpmOverridesArg = "--add-npm-overrides";
+  const addRushOverrides = args.includes(addRushOverridesArg);
+  const addNpmOverrides = args.includes(addNpmOverridesArg);
+  const packageJsonPaths = args.filter((arg) => arg !== addRushOverridesArg && arg !== addNpmOverridesArg);
```
weidongxu-microsoft, line 31: I probably should use yargs, but maybe I do it in another PR.

packages/typespec-bump-deps/src/cli/cli.ts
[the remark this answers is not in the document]
pshao25: It's kind of tech debt, but we have two `pakage.json`s in our repo. The first [one](https://github.com/Azure/autorest.csharp/blob/e8c8b670c684b9680bd8b747f5daedbb138b4759/src/CADL.Extension/Emitter.Csharp/package.json) contains the actual dependecies, which we want to bump the typespec version. The second [one](https://github.com/Azure/autorest.csharp/blob/e8c8b670c684b9680bd8b747f5daedbb138b4759/package.json) adds the first one as dependency, which is the entry we install the package, i.e., we install packages with this package.json. 

Our expectation is: after installation, the root folder has typespec of dev version. The only way I could think of now is: in the second package.json add the overrides section and put all the updated dependecies from the first one into it.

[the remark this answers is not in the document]
weidongxu-microsoft: Java has 2 package.json too.

You can call it twice, first without the option to typespec project only update the dependencies, second with the option to test project and also add overrides.

pshao25 (approved):

MaryGao (approved):

packages/typespec-bump-deps/src/cli/cli.ts
timotheeguerin: There is actually a built-in parser for args in node now:
https://nodejs.org/api/util.html#utilparseargsconfig

It is simple but that will cover those use cases very well I think.

weidongxu-microsoft: Changed to parseArgs.

This TS error on type get me puzzled for quite some time.

TS won't accept
```ts
  const options = {
    "add-rush-overrides": {
      type: "boolean",
    },
    "add-npm-overrides": {
      type: "boolean",
    },
  };
```

So I had to explicitly write have the union there. Not sure if there is better way to write this.
```ts
  const booleanTypeStr: "string" | "boolean" = "boolean";
  const options = {
    "add-rush-overrides": {
      type: booleanTypeStr,
    },
    "add-npm-overrides": {
      type: booleanTypeStr,
    },
  };
```

timotheeguerin: hhm, weird, let me try out

timotheeguerin: ok just had to use `as const` so typscript keeps the exact type and doesn't downgrade to `string`

```
const options = {
    "add-rush-overrides": {
      type: "boolean",
    },
    "add-npm-overrides": {
      type: "boolean",
    },
  } as const;
```

timotheeguerin (approved):

PR #329, opened 2023-05-18
Add docs to enums and enum values.

tjprescott:
Required PR due to: https://github.com/Azure/typespec-azure/pull/3006

This will also require a release once merged.


.changeset/modern-worms-exercise.md
```diff
@@ -0,0 +1,2 @@
+---
```
timotheeguerin, line 1: This should have been the changelog, you can click the comment on the PR to auto populate with the commit message

packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
```diff
@@ -20,7 +20,9 @@ using TypeSpec.Versioning;
 )
 namespace _Specs_.Azure.Core.Basic;
 
+@doc("The version of the API.")
 enum Versions {
+  @doc("The version 2022-12-01-preview.")
```
timotheeguerin, line 25: comment here too but do we really need that doc on version enum, not sure what are people going to be commenting apart from the same thing again

.changeset/modern-worms-exercise.md
tjprescott: It said to use `--empty` since there is no functional, relevant change to cadl ranch.

packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
tjprescott: This PR is just to make the e2e test pass.

.changeset/modern-worms-exercise.md
timotheeguerin: It won’t bum the version when releasing without change

timotheeguerin: make a patch change

packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
tjprescott: The point of the rule is to make sure that enums and enum values have documentation. It's a slippery slope to try and decide that some enums are self-explanatory enough not to require it. If they are, then you could just suppress it.

timotheeguerin: But would the version enum be carried over to the generated client?

timotheeguerin: sure in general we can't, but for `Versions` we can know it is the version enum. I am saying that as this is a change all the spec will most likely need a change and I don't think is very valuable. But not gonna block on this.

tjprescott:
@timotheeguerin so somehow you and I are both codeowners and you have approved this, but I can't merge it.

timotheeguerin:
hhm, I think this is because DPG teams didn't want change to scenarios merged without approval from them

lmazuel (approved):

PR #330, opened 2023-05-19
model extends/is record<>

msyyc:


lmazuel (changes requested):
- I would not put this in inheritance. TSP uses "extends", but it's the concept of "additional properties". Inheritance is misleading here.
- I need a GET as well

timotheeguerin:
note also `extends` vs `is` is different, you probably still want `is`. Extends will say i am inheriting a Record<string>, is will just say I am a `Record<string>`

See the differently generated openapi 
https://cadlplayground.z22.web.core.windows.net/?c=bW9kZWwgTcQGUmVjb3JkVW5rbm93biBleHRlbmRzIMYWPHXGFz4gewogIG5hbWU6IHN0cmluZzsKfQoK2EYyIGnfQsZC

weidongxu-microsoft:
On swagger output of
```ts
model ModelRecordUnknown2 is Record<unknown> {
  name: string;
}
```

Does not have
```json
    ModelRecordUnknown2:
      type: object
      additionalProperties: {}
```

It does not have anything about "name"?

msyyc:
> note also `extends` vs `is` is different, you probably still want `is`. Extends will say i am inheriting a Record, is will just say I am a `Record<string>`
> 
> See the differently generated openapi https://cadlplayground.z22.web.core.windows.net/?c=bW9kZWwgTcQGUmVjb3JkVW5rbm93biBleHRlbmRzIMYWPHXGFz4gewogIG5hbWU6IHN0cmluZzsKfQoK2EYyIGnfQsZC
> 
> I don't know how that affect generated client however but I could imagine something like that
> 
> ```
> 
> class WithExtends extends Dictionary<string, string> {
> 
>   public Name {get; set;}
> }
> 
> 
> class WithIs implements IDictionary<string, string> {
>   ... need to imploement dictionary interface
>   public Name {get; set;}
> }
> ```

@timotheeguerin Like weidong [said](https://github.com/Azure/cadl-ranch/pull/330#issuecomment-1556453560), we shall encourage users `extends` instead of `is` otherwise users can't get expected output about

msyyc:
> * I would not put this in inheritance. TSP uses "extends", but it's the concept of "additional properties". Inheritance is misleading here.
> * I need a GET as well

I add new folder `record` under `type/model`

tadelesh:
I prefer to add it under to `type/property` named with `additional-properties`. @timotheeguerin I have a question for it. From TSP type relations, it seems custom model with properties are derived from Record. It that mean all custom models are already support additional properties? What's the difference for an explicit `Model extends Record<unknown>`?

timotheeguerin:
Hhm yeah seems like we have a bug in autorest and openapi3 emitter, those properties should not be dropped. I think the idea was there though that with `is` you have `additionalProperties` on the schema itself. Vs with `extends` it use an `allOf`.

I think the takeaway is mostly both can happen and I think we should recommend people to use the `is` variant. If you generate the same thing then maybe its ok

timotheeguerin:
> .

there is a [discussion on Teams about that](https://teams.microsoft.com/l/message/19:906c1efbbec54dc8949ac736633e6bdf@thread.skype/1679502780311?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1679502780311&teamName=Azure%20SDK&channelName=TypeSpec%20(Cadl)%20Discussion%20%F0%9F%90%AE&createdTime=1679502780311) 

short answer, no models do not have additional properties in the TypeSpec system(I made a wrong statement before), we might need to update the type graph a little bit
```
model Foo { } !== model Foo is Record<unknown> {}
```

msyyc:
> @msyyc Yuchao Yan FTE I prefer to add it under to `type/property` named with `additional-properties`. @timotheeguerin Timothee Guerin FTE I have a question for `Record<unknown>`. From TSP type relations, it seems custom model with properties are derived from Record. It that mean all custom models are already support additional properties? What's the difference for an explicit `Model extends Record<unknown>`?

@tadelesh Make sense. I will move it to `type/property/additional-properties`

msyyc:
@timotheeguerin @lmazuel @tadelesh @weidongxu-microsoft 
Updated:
I understand that language emitter shall have same behavior for `extends Record<>` or `is Record<>`, I add test for both scenario.

packages/cadl-ranch-specs/http/type/property/additional-properties/mockapi.ts
```diff
@@ -0,0 +1,51 @@
...
+      const expectedBody = JSON.parse(JSON.stringify(body));
+      req.expect.coercedBodyEquals(expectedBody);
+      return {
+        status: 204,
+      };
+    }),
+  };
+}
+
+const extendRecordUnknown = createMockApis("extends/record/unknown", {
```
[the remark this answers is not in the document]
tadelesh: I perfer `extends`.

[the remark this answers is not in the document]
timotheeguerin: not sure we need `/record/` in the route. if we already are scopped in `/additional-properties` this feels redundant

packages/cadl-ranch-specs/http/type/property/additional-properties/main.tsp
```diff
@@ -0,0 +1,79 @@
...
+    {
+      doc: TDoc,
+    }
+  )
+  @get
+  @doc("Get call")
+  get(): TModel;
+
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "For testing"
+  #suppress "@azure-tools/typespec-azure-core/bad-record-type" "For testing"
```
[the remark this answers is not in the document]
timotheeguerin: yeah shouldn't be needed here, if it is sounds like we might have a bug

packages/cadl-ranch-specs/http/type/property/additional-properties/main.tsp
```diff
@@ -0,0 +1,79 @@
...
+namespace Type.Property.AdditionalProperties;
+
+#suppress "@azure-tools/typespec-azure-core/bad-record-type" "For testing"
+@doc("The model extends from Record<unknown> type.")
+model ModelExtendsRecordUnknown extends Record<unknown> {
+  @doc("The name property")
+  name: string;
+}
+
+#suppress "@azure-tools/typespec-azure-core/bad-record-type" "For testing"
```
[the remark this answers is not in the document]
timotheeguerin: I think it complains about just using `Record<unknown>` if you used `Record<string>` then it would not. But for testing here its fine to suppress.

packages/cadl-ranch-specs/http/type/property/additional-properties/mockapi.ts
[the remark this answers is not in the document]
msyyc: I think `extendsRecordUnknown` is better

packages/cadl-ranch-specs/http/type/property/additional-properties/main.tsp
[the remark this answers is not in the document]
msyyc: It is necessary otherwise CI fails:
image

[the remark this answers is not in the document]
weidongxu-microsoft: OK, seems lint check for body more than for response :-)

packages/cadl-ranch-specs/http/type/property/additional-properties/main.tsp
```diff
@@ -0,0 +1,79 @@
...
+using TypeSpec.Http;
+using Azure.ClientGenerator.Core;
+
+@doc("Illustrates various property types for models")
+@scenarioService("/type/property/additionalProperties")
+namespace Type.Property.AdditionalProperties;
+
+#suppress "@azure-tools/typespec-azure-core/bad-record-type" "For testing"
+@doc("The model extends from Record<unknown> type.")
+model ModelExtendsRecordUnknown extends Record<unknown> {
```
MaryGao, line 15: @timotheeguerin I'd like to confirm my understanding here.

Any model which extends Record<> would enbale its `additionalProperties` flag, no matter `extends Record<string>` or `extends Record<float>`.

packages/cadl-ranch-specs/http/type/property/additional-properties/main.tsp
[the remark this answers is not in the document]
timotheeguerin: hhm ok, yeah I think we have an issue. The rule seems to be complaining when you reference reference the `model is Record<string>` again

Filed an issue https://github.com/Azure/typespec-azure/issues/3054

packages/cadl-ranch-specs/http/type/property/additional-properties/main.tsp
timotheeguerin: well it would enable additional properties of the type provide in the record. I would not expect `is Record<string>` and `is Record<float>` to produce the same model.

I guess yeah we could add more types than just unknown. Similar to what we have for property type.

msyyc: I think Record<unknown> is wider than Record<string> and Record<float>, etc. I add the test for check int/string/boolean: https://github.com/Azure/cadl-ranch/blob/2f4816a30f06c18c049dbaf0c2a1e8cd9cf18f83/packages/cadl-ranch-specs/http/type/property/additional-properties/mockapi.ts#L44-L49
So there is no need to add Record<string/...> I think.

timotheeguerin: @msyyc I think it can still be valuable to make sure you generate the right thing but we can come back to it

packages/cadl-ranch-specs/http/type/property/additional-properties/mockapi.ts
```diff
@@ -0,0 +1,51 @@
...
+      return {
+        status: 204,
+      };
+    }),
+  };
+}
+
+const extendRecordUnknown = createMockApis("extendsRecordUnknown", {
+  name: "ModelExtendsRecordUnknown",
+  prop1: 32,
```
timotheeguerin, line 37: can we just rename those properties to be describing the type they are suppose to set. Will make it easier.

timotheeguerin (approved):

packages/cadl-ranch-specs/http/type/property/additional-properties/main.tsp
weidongxu-microsoft: > @timotheeguerin I'd like to confirm my understanding here:
> 
> Any model which extends Record<> would enbale its `additionalProperties` flag, no matter `extends Record<string>` or `extends Record<float>`.
> 
> If yes, do you think we need to also add other types cases?

I think any `Record<>` would enable the `additionalProperties` feature.
However the detail of the type/class in language would likely be a bit different depends on the type in Record.
E.g. Java would have `Map<String, String>` instead of `Map<String, Object>`, if type in Record is `string`.

msyyc: @weidongxu-microsoft @MaryGao I add `Record<string>` and `Record<float>`. But actually Python doesn't care about the type, so please feel free to commit directly if you need.

packages/cadl-ranch-specs/http/type/property/additional-properties/mockapi.ts
msyyc: done

packages/cadl-ranch-specs/http/type/property/additional-properties/main.tsp
weidongxu-microsoft: nit, if we prefer service to use `is Record<>`, we can have less type coverage on `extends Record<>` (maybe only one `Record<unknown>` be enough for `extends`).

@chunyu3 @pshao25 Is .NET OK with this PR?

packages/cadl-ranch-specs/http/type/property/additional-properties/main.tsp
```diff
@@ -0,0 +1,131 @@
+import "@typespec/http";
+import "@azure-tools/cadl-ranch-expect";
+import "@azure-tools/typespec-client-generator-core";
+import "@azure-tools/typespec-azure-core";
+
+using TypeSpec.Http;
+using Azure.ClientGenerator.Core;
+
+@doc("Illustrates various property types for models")
```
tadelesh, line 9:
```suggestion
@doc("Tests for additional properties of models")
```

packages/cadl-ranch-specs/http/type/property/additional-properties/main.tsp
```diff
@@ -0,0 +1,131 @@
...
+// ********************************************** Record<unknown> **********************************************
+#suppress "@azure-tools/typespec-azure-core/bad-record-type" "For testing"
+@doc("The model extends from Record<unknown> type.")
+model ExtendsUnknownAddtionalProperties extends Record<unknown> {
+  @doc("The name property")
+  name: string;
+}
+
+@route("/extendsRecordUnknown")
+@operationGroup
```
tadelesh, line 60: Shall we group them with "extends or is" instead of group them one by one?

tadelesh (commented):
some minor comment

tadelesh (commented):
some minor comment

packages/cadl-ranch-specs/http/type/property/additional-properties/main.tsp
msyyc: each interface has two operation: put and get. I follow the same pattern like value-types.

packages/cadl-ranch-specs/http/type/property/additional-properties/main.tsp
```diff
@@ -0,0 +1,131 @@
...
+  id: float32;
+}
+
+@route("/extendsRecordFloat")
+@operationGroup
+interface ExtendsFloat extends ModelOperations<ExtendsFloatAddtionalProperties, "{'id': 42.42, 'prop': 42.42}"> {}
+
+#suppress "@azure-tools/typespec-azure-core/bad-record-type" "For testing"
+@doc("The model is from Record<float32> type.")
+model IsFloatAddtionalProperties is Record<float32> {
```
qiaozha, line 124: Can you add test cases for extending from Record of Model. as well as model arrays ? we have a customer asking for this https://github.com/Azure/autorest.typescript/issues/2033. Thanks.

msyyc: Added testcase for Model and Model[]. @qiaozha @chunyu3 Please help review

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -3553,6 +3553,226 @@ Expected input body:
...
+```
+
+### Type_Property_AdditionalProperties_IsModelArray_put
+
+- Endpoint: `put /type/property/additionalProperties/isRecordModelArray`
+
+Expected input body:
+
+```json
+{ "prop": [{ "state": "ok" }, { "state": "ok" }] }
```
weidongxu-microsoft, line 3723: why 2 state? JSON dict can have dupliate key?

weidongxu-microsoft (approved):
Make sure you've tested the mockapi in python.

qiaozha (approved):
LGTM

tadelesh (approved):
Though seems a little bit weird to give both `is` and `extends` example, but should all be valid cases for non-branding.

lmazuel (approved):
Was approved by enough smart people to remove my "require change" :)

PR #334, opened 2023-05-24
Add mulitclient test case

qiaozha:

Add test cases for this customization in client hierarchy design https://github.com/Azure/typespec-azure/blob/main/docs/howtos/DataPlane%20Generation%20-%20DPG/02client.md


packages/cadl-ranch-specs/http/clientHierarchy/renaming/main.tsp
timotheeguerin, line 1: are all of those using the same service? should we import a common file then?

packages/cadl-ranch-specs/http/client-hierarchy/default/client.tsp
```diff
@@ -0,0 +1,14 @@
...
+using Azure.ClientGenerator.Core;
+
+// @doc("""
+//   This is to show that if we don't do any customization. The client side should be able to call the api like
+//   const client = new MultipleClient();
+//   client.createOrUpdate()
+//   client.get() 
+//   client.createWithHeaders()
+//   client.getArray()
+//   """)
```
qiaozha, line 14: Because there are no customization in this scenario, I am not very sure how to keep this doc.

packages/cadl-ranch/src/scenarios-resolver.ts
```diff
@@ -27,13 +27,31 @@ interface CadlRanchScenarioFile {
...
-  logger.info(`Found ${scenarios.length} scenarios.`);
+  scenarios.sort();
+  const result: string[] = [];
+  let shouldSkip = false;
+  scenarios.forEach((scenario, index) => {
+    if (shouldSkip) {
+      shouldSkip = false;
+    } else {
+      if (index + 1 < scenarios.length) {
+        if (scenario.replace("client.tsp", "main.tsp") === scenarios[index + 1]) {
```
timotheeguerin, line 41: Does this mean that if it finds a `main.tsp` after a `client.tsp` then it skips it?

packages/cadl-ranch/src/scenarios-resolver.ts
```diff
@@ -27,13 +27,31 @@ interface CadlRanchScenarioFile {
...
   await ensureScenariosPathExists(scenariosPath);
-  const pattern = `${normalizePath(scenariosPath)}/**/main.tsp`;
+  const pattern = [`${normalizePath(scenariosPath)}/**/client.tsp`, `${normalizePath(scenariosPath)}/**/main.tsp`];
   logger.debug(`Looking for scenarios in ${pattern}`);
   const scenarios = await findFilesFromPattern(pattern);
-  logger.info(`Found ${scenarios.length} scenarios.`);
+  scenarios.sort();
+  const result: string[] = [];
+  let shouldSkip = false;
+  scenarios.forEach((scenario, index) => {
```
timotheeguerin, line 36: what about using a `.filter` instead

packages/cadl-ranch/src/scenarios-resolver.ts
qiaozha: yes, because I have sorted it, client,tsp should appear first.

qiaozha: Actually I was thinking skip the next main.tsp when I found client.tsp, that's why I don't use filter at first, because it can only filter the current item, right ? but then I realize we can skip main.tsp and compare with the previous one see if it's client.tsp, then skip main.tsp when found main.tsp. That way filter would work.

timotheeguerin: I would maybe instead use a set and do a getDirName and join main.tsp instead as it feels more reliable than just checking the previous item in the sorted list

qiaozha: Okay, I changed this, not sure if I fully understood. anyway, let me know if it's not.

packages/cadl-ranch-specs/http/client/structure/renamed-operation/mockapi.ts
```diff
@@ -0,0 +1,34 @@
+import { passOnSuccess, mockapi } from "@azure-tools/cadl-ranch-api";
+import { ScenarioMockApi } from "@azure-tools/cadl-ranch-api";
+
+export const Scenarios: Record<string, ScenarioMockApi> = {};
+Scenarios.Client_Structure_RenamedOperation_one = passOnSuccess(
+  mockapi.patch("/client/structure/renamed-operation/one", (req) => {
+    return { status: 200 };
+  }),
+);
+Scenarios.Client_Structure_RenamedOperation_two = passOnSuccess(
```
timotheeguerin, line 10:
```suggestion
Scenarios.Client_Structure_RenamedOperation = passOnSuccess([
...

]);
```

change to this and move all the mockapi registrations inside

packages/cadl-ranch/src/scenarios-resolver.ts
```diff
@@ -31,16 +31,20 @@ export async function findScenarioCadlFiles(scenariosPath: string): Promise<Cadl
   const pattern = [`${normalizedScenarioPath}/**/client.tsp`, `${normalizedScenarioPath}/**/main.tsp`];
   logger.debug(`Looking for scenarios in ${pattern}`);
   const fullScenarios = await findFilesFromPattern(pattern);
+  logger.info(`Found ${fullScenarios.length} full scenarios.`);
   const scenarioSet = new Set(fullScenarios);
   const scenarios = fullScenarios.filter((scenario) => {
     // Exclude main.tsp that have a client.tsp next to it, we should use that instead
-    return scenario.endsWith("/main.tsp") && scenarioSet.has(normalizePath(join(dirname(scenario), "client.tsp")));
+    return (
```
timotheeguerin, line 38: why was this not working before with building the path. Doing a .replace is less reliable if you end up with something in the path above it would change it.

packages/cadl-ranch-specs/http/client/structure/renamed-operation/mockapi.ts
qiaozha: Should have updated.

packages/cadl-ranch/src/scenarios-resolver.ts
qiaozha: This is not working in my another laptop which is a windows system yesterday. I changed it to use path.sep, but I am using mac now. will not go to the office in the rest week. so I am not sure if this change works for windows or not

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -618,6 +618,101 @@ Expected response body:
...
+
+clientB.renamedTwo();
+clientB.renamedFour();
+clientB.renamedSix();
+```
+
+### Client_Structure_RenamedOperation
+
+- Endpoints:
+  - `get /one`
```
qiaozha, line 649: Is this endpoints too general ? @timotheeguerin

packages/cadl-ranch-specs/http/client/structure/default/mockapi.ts
```diff
@@ -0,0 +1,24 @@
+import { passOnSuccess, mockapi } from "@azure-tools/cadl-ranch-api";
+import { ScenarioMockApi } from "@azure-tools/cadl-ranch-api";
+
+export const Scenarios: Record<string, ScenarioMockApi> = {};
+Scenarios.Client_Structure_Service = passOnSuccess([
+  mockapi.patch("/one", (req) => {
```
qiaozha, line 6: how to make sure the tests are all covered if they all share the same endpoints ?

timotheeguerin: the full url for all of those is 
```
http://localhost:3000/client/structure/{client}/...
```

so make sure each endpoint here is `/client/structure/{client}/...` and replace `{client}` with the scenario name

packages/cadl-ranch/src/actions/validate-mock-apis.ts
```diff
@@ -3,6 +3,7 @@ import pc from "picocolors";
 import { findScenarioCadlFiles, loadScenarioMockApiFiles } from "../scenarios-resolver.js";
 import { importTypeSpec, importCadlRanchExpect } from "../cadl-utils/import-cadl.js";
 import { createDiagnosticReporter } from "../utils/diagnostic-reporter.js";
+import { sep } from "path";
```
timotheeguerin, line 6: We should not use any of the native path utils, without normalizing it is better to make sure everything is using `/` otherwise when comparing path it can always result in different result across platforms. I am wondering if those issue you are having are the same as this #382

timotheeguerin: I'll make a separate PR fixing this, so you can revert those changes

timotheeguerin: PR fixing it https://github.com/Azure/cadl-ranch/pull/383

packages/cadl-ranch/src/scenarios-resolver.ts
```diff
@@ -1,7 +1,7 @@
 import { ScenarioMockApi } from "@azure-tools/cadl-ranch-api";
 import { Scenario } from "@azure-tools/cadl-ranch-expect";
 import { Operation } from "@typespec/compiler";
-import { join, relative, resolve } from "path";
+import { join, relative, resolve, dirname, sep } from "path";
```
timotheeguerin, line 4: Same here you can revert those changes I believe after the other PR is merged

packages/cadl-ranch-specs/http/client/structure/default/mockapi.ts
qiaozha: But the summary is saying the otherwise, should I change anything in generating the summary ?

packages/cadl-ranch/src/actions/validate-mock-apis.ts
qiaozha: Yeah, I think so, I will revert those changes after #383 get merged.

packages/cadl-ranch-specs/http/client/structure/default/mockapi.ts
timotheeguerin: Yeah summary doesn’t take the server url into account. We should fix that

qiaozha: Got it, I will see how to do that? maybe we should add some validation checks here to make sure the mockapi.ts is reflecting the real apis?

qiaozha: Because it should have reported errors now, but it's not.

timotheeguerin: Yeah but it it’s a little tricky when we have some path parameter to make sure things match and not find false positive.

packages/cadl-ranch-specs/http/client/structure/default/mockapi.ts
```diff
@@ -3,22 +3,22 @@ import { ScenarioMockApi } from "@azure-tools/cadl-ranch-api";
 
 export const Scenarios: Record<string, ScenarioMockApi> = {};
 Scenarios.Client_Structure_Service = passOnSuccess([
-  mockapi.patch("/one", (req) => {
+  mockapi.patch("client/structure/default/one", (req) => {
```
timotheeguerin, line 6:
```suggestion
  mockapi.patch("/client/structure/default/one", (req) => {
```

think they should start with /

packages/cadl-ranch/src/actions/validate-mock-apis.ts
qiaozha: updated, should have resolved this comment.

packages/cadl-ranch/src/scenarios-resolver.ts
qiaozha: updated.

packages/cadl-ranch-specs/http/client/structure/default/mockapi.ts
qiaozha: will track with #385

packages/cadl-ranch-specs/http/client/structure/renamed-operation/client.tsp
weidongxu-microsoft, line 1: Do we need a "renamed-operation" test suite? "multi-client" already has "rename" in it, and "two-operation-group" already has "operation group" in it.

PS: "two-operation-group" seems better named "multiple-operation-group"?

packages/cadl-ranch-specs/http/client/structure/common/service.tsp
```diff
@@ -0,0 +1,60 @@
...
+  3. one client can have more than one operations groups
+  4. split one interface into two clients
+  5. have two clients with operations come from different interfaces
+  6. have two clients with a hierarchy relation.
+  """)
+@server(
+  "http://localhost:3000/client/structure/{client}",
+  "",
+  {
+    client: ClientTypeString,
```
weidongxu-microsoft, line 23: Why not `client: ClientType`? I assume `@knownValues` is an outdate way to define extensible enum.

If it is an enum, just use an enum. (playground is good, not sure how emitter works)

packages/cadl-ranch-specs/http/client/structure/renamed-operation/client.tsp
qiaozha: They are mostly coming from this document https://github.com/Azure/typespec-azure/blob/main/docs/howtos/DataPlane%20Generation%20-%20DPG/02client.md

packages/cadl-ranch-specs/http/client/structure/renamed-operation/client.tsp
```diff
@@ -0,0 +1,38 @@
...
+  name: "RenamedOperationClient",
+  service: Client.Structure.Service,
+})
+@scenario
+namespace Client.Structure.RenamedOperation;
+
+// Those operations are renamed at the root
+op renamedOne is Client.Structure.Service.one;
+op renamedThree is Client.Structure.Service.Foo.three;
+op renamedFive is Client.Structure.Service.Foo.three;
```
tadelesh, line 30: It looks like the difference between this test and mutli-operation-group is this has client level methods?

packages/cadl-ranch-specs/http/client/structure/renamed-operation/client.tsp
weidongxu-microsoft: I be good if no one else has opinion.

What I feel is that tests take time on dev from all languages to write, and also take future efforts to maintain. Hence if feature can be tested with less file/folder, the better.

packages/cadl-ranch-specs/http/client/structure/renamed-operation/client.tsp
qiaozha: yeah

packages/cadl-ranch-specs/http/client/structure/common/service.tsp
qiaozha: changed to ClientType.

packages/cadl-ranch-specs/cadl-ranch-summary.md
qiaozha: pending https://github.com/Azure/cadl-ranch/pull/386 to fix it and will be fixed after rerun the generate-scenarios-summary

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -618,6 +618,101 @@ Expected response body:
...
+const client = new MultiClient("default");
+client.one();
+client.two();
+client.three();
+client.four();
+client.five();
+client.six();
+```
+
+### Client_Structure_TwoOperationGroup
```
weidongxu-microsoft, line 692: @timotheeguerin Let us know if the test case hierarchy (e.g. `Client_Structure_TwoOperationGroup`) is good for you.

I am a bit unsure about "Structure", but don't have good recommendation.

weidongxu-microsoft (commented):
LGTM

packages/cadl-ranch-specs/http/client/structure/multi-client/mockapi.ts
```diff
@@ -0,0 +1,24 @@
+import { passOnSuccess, mockapi } from "@azure-tools/cadl-ranch-api";
+import { ScenarioMockApi } from "@azure-tools/cadl-ranch-api";
+
+export const Scenarios: Record<string, ScenarioMockApi> = {};
+Scenarios.Client_Structure_MultiClient = passOnSuccess([
+  mockapi.patch("/client/structure/multi-client/one", (req) => {
+    return { status: 200 };
```
timotheeguerin, line 7: I think all of those needs to be 204, as the operation returns `void`

weidongxu-microsoft (approved):

tadelesh (approved):

packages/cadl-ranch-specs/cadl-ranch-summary.md
qiaozha: fixed

packages/cadl-ranch-specs/http/client/structure/multi-client/mockapi.ts
qiaozha: updated, should have been resolved.

packages/cadl-ranch-specs/http/client/structure/multi-client/mockapi.ts
```diff
@@ -0,0 +1,24 @@
...
+import { ScenarioMockApi } from "@azure-tools/cadl-ranch-api";
+
+export const Scenarios: Record<string, ScenarioMockApi> = {};
+Scenarios.Client_Structure_MultiClient = passOnSuccess([
+  mockapi.patch("/client/structure/multi-client/one", (req) => {
+    return { status: 204 };
+  }),
+  // mockapi.patch("/client/structure/multi-client/two", (req) => {
+  //   return { status: 204 };
+  // }),
```
qiaozha, line 11: Commented two, four, six out, because we don't reference them in client.tsp. But my question is, how to make sure one, three, five have been called twice? @weidongxu-microsoft @timotheeguerin

weidongxu-microsoft: I don't have a good idea. Though I don't think this be a thing that must be verified in mockapi.

I assume as long as generated client is good, test uses both client calls 3 methods, that be enough (if endpoint is not right, mockapi would give error). No need to verify in mockapi to verify whether it is called once, or twice. If it get called a third time, it does not mean much...

timotheeguerin (approved):

PR #335, opened 2023-05-24
Empty model test

MaryGao:

### DIFF
--- a/packages/cadl-ranch-specs/http/type/model/empty/main.tsp
+++ b/packages/cadl-ranch-specs/http/type/model/empty/main.tsp
@@ -1,0 +1,30 @@
+import "@typespec/http";
+import "@azure-tools/cadl-ranch-expect";
+
+using TypeSpec.Http;
+
+/**
+ * Those tests are meant to test different behavior of records if they are used as input, output or both.
+ * This is something that might not matter in your emitter.
+ *
+ * This is valuable if you for example only decide to create a public constructor if a model is used as input.
+ */
+@doc("Illustrates usage of empty model used in operation's parameters and responses.")
+@scenarioService("/type/model/empty")
+namespace Type.Model.Empty;
+
+@doc("Empty model used in operation parameters")
+model EmptyInput {}
+
+@doc("Empty model used in operation return type")
+model EmptyOutput {}
+
+@scenario
+@scenarioDoc("Send a POST request with the following body {}")
+@route("/input")
+op input(@body input: EmptyInput): void;
+
+@scenario
+@scenarioDoc("Send a GET request which return the following body {}")
+@route("/output")
+op output(): EmptyOutput;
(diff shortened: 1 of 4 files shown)


tadelesh (approved):

packages/cadl-ranch-specs/http/type/model/empty/main.tsp
```diff
@@ -0,0 +1,24 @@
...
+
+@scenario
+@scenarioDoc("Send a POST request with the following body {}")
+@route("/input")
+op input(@body input: EmptyInput): void;
+
+@scenario
+@scenarioDoc("Send a GET request which return the following body {}")
+@route("/output")
+op output(): EmptyOutput;
```
weidongxu-microsoft, line 24: I assume both can be on same endpoint, one as `@get`, anothor `@put`? No need for input/output in endpoint.

e.g. https://github.com/Azure/cadl-ranch/blob/58fdd380c43881abfbd92e9c52adc0d19bd1cef3/packages/cadl-ranch-specs/http/type/property/optional/main.tsp#L20-L69

weidongxu-microsoft:
@chunyu3 @pshao25 for review for .NET

packages/cadl-ranch-specs/http/type/model/empty/main.tsp
MaryGao: Sure I'll update to the same endpoint.

chunyu3: can we add a roundtrip empty model which will be used as both input and output (may or may not in the same operation)

packages/cadl-ranch-specs/http/type/model/empty/main.tsp
```diff
@@ -0,0 +1,24 @@
...
+
+@doc("Empty model used in operation parameters")
+model EmptyInput {}
+
+@doc("Empty model used in operation return type")
+model EmptyOutput {}
+
+@scenario
+@scenarioDoc("Send a POST request with the following body {}")
+@route("/input")
```
chunyu3, line 18: missing `@post`?

MaryGao: TSP compiler would automatically add @post here, I'll update the action explicitly.

packages/cadl-ranch-specs/http/type/model/empty/main.tsp
weidongxu-microsoft: @chunyu3 Do you propose to add another op, or just convert the current `@put` to roundtrip (so that it returns 200 + `{}`)?

chunyu3: @weidongxu-microsoft  Both are fine.

lmazuel (approved):

PR #339, opened 2023-05-26
Minor adjustments to unblock Azure Core pull requests

daviwil:
Hey folks!  Here are two small tweaks to cadl-ranch specs to respond to the following Azure Core PRs:

- https://github.com/Azure/typespec-azure/pull/3067
- https://github.com/Azure/typespec-azure/pull/3070

This is my first time contributing changes to specs in this repo so let me know if I've missed anything.

### DIFF
--- a/packages/cadl-ranch-specs/http/azure/core/lro/rpc/main.tsp
+++ b/packages/cadl-ranch-specs/http/azure/core/lro/rpc/main.tsp
@@ -130,15 +130,15 @@ Expected response body:
 }
 ```
 """)
+@route("/same-poll-result")
 interface SamePollResult {
   #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Will use Azure.Core when LongRunningRpcOperation is ready"
   @doc("Creates a Job")
-  @route("/same-poll-result/jobs")
+  @route("/jobs")
   @global.Azure.Core.pollingOperation(SamePollResult.getJob)
   createJob is LocalCore.LongRunningRpcOperation<JobData, JobResult>;
 
   @doc("Gets the status of a Job")
-  @route("/same-poll-result")
   getJob is ResourceOperations.ResourceRead<JobResult>;
 }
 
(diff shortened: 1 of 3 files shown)


packages/cadl-ranch-specs/http/azure/core/lro/rpc/main.tsp
```diff
@@ -130,15 +130,15 @@ Expected response body:
 }
 ```
 """)
+@route("/same-poll-result")
 interface SamePollResult {
```
timotheeguerin, line 134: nothing against this change but how is this changing for the new rule? Wasn't it just that it shouldn't have route at all?

timotheeguerin (approved):

packages/cadl-ranch-specs/http/azure/core/lro/rpc/main.tsp
timotheeguerin: ok I see, the other operation is a custom copy of the core one so it doesn't complain about it.

I think the better choice here is just to suppress on the operation because that's by design we have this here

daviwil: The point of the new rule is to discourage the use of @route directly on an operation because someone might think that it will set the entire route of the operation, not just the fragment for the current container level.  The real fix is to put the route prefix on the container.  

If you want, I can just suppress instead, I'm fine with it either way.

timotheeguerin: sorry thought you were already gone on leave, was just updating your branch already.

Yeah I think it is better for cadl-ranch as the route is really specific to how we want the test server to be structured and not really matter that it doesn't follow the standard(most cald-ranch specs are not azure.core)

timotheeguerin: actually nvm, I'll revert you are probably right

iscai-msft (approved):

weidongxu-microsoft:
Thanks David and Timothee.

We are in progress to migrate to the (already merged but not yet released) LongRunningRpcOperation in Azure.Core. So hopefully we will not need the local template in util.tsp after next TypeSpec release.

PS: There is a bit of pending question in that PR. Hope we can get through them before next release. https://github.com/Azure/typespec-azure/pull/2983#discussion_r1203671504

PR #341, opened 2023-05-29
fix wrong header check for custom http

tadelesh:
Minor issue from https://github.com/Azure/cadl-ranch/pull/262

### DIFF
--- a/packages/cadl-ranch-specs/http/authentication/http/custom/mockapi.ts
+++ b/packages/cadl-ranch-specs/http/authentication/http/custom/mockapi.ts
@@ -8,7 +8,7 @@ const validAndInvalidScenarios = getValidAndInvalidScenarios(
   "http/custom",
   "invalid-api-key",
   function addOptionalParamOldApiVersionNewClientValidate(req: MockRequest): void {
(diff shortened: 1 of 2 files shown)


packages/cadl-ranch-specs/http/authentication/http/custom/mockapi.ts
```diff
@@ -8,7 +8,7 @@ const validAndInvalidScenarios = getValidAndInvalidScenarios(
   "http/custom",
   "invalid-api-key",
   function addOptionalParamOldApiVersionNewClientValidate(req: MockRequest): void {
-    req.expect.containsHeader("Authorization", "SharedAccessKey valid-key");
+    req.expect.containsHeader("authorization", "SharedAccessKey valid-key");
```
weidongxu-microsoft, line 11: Is this method always take lower-case?

tadelesh: Yes. Refer: https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules__types_node_http_d_._http_.incominghttpheaders.html.

weidongxu-microsoft (approved):

PR #342, opened 2023-05-30
spec, update lro/rpc test scenario

weidongxu-microsoft:
Ref https://github.com/Azure/typespec-azure/issues/3108

Recreate a similar legacy LRO in cadl-ranch, with [healthinsights](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/HealthInsights/healthinsights.common/primitives.tsp) in mind ([recording](https://github.com/Azure/azure-sdk-for-java/blob/2369bfcf07309646f646aa05d94cdad40e151ba0/sdk/healthinsights/azure-health-insights-cancerprofiling/src/test/resources/session-records/CancerProfilingClientTest.test.json#L18)). Basically it is a "post to create" LRO.

Verified that `logicalResult` and `pollingInfo.responseModel` (this probably only used by Java) is both `JobResult`.

I've removed `@pollingOperation` operation link (and the `getPoll` op). `getLroMetadata` does not need it. And it was not correct in the first place.

Here is commit to modify healthinsights template, to describe correctly that the `TPollResponse` is `JobResult` model, not `OperationStatus<>` model. https://github.com/Azure/cadl-ranch/pull/342/commits/ac09c0b7d503f0b24238297d794dbcf14f41038e

---

I think I should move this scenario out of `http/azure/core/lro/rpc` folder, as it won't be from Azure.Core.

Any suggestion? `http/azure/legacy/lro`?


packages/cadl-ranch-specs/http/azure/core/lro/rpc/main.tsp
```diff
@@ -132,90 +134,10 @@ Expected response body:
...
-  "comment": "async job",
-  "status": "Succeeded",
-  "results": ["job1 result"]
-}
-```
-""")
-@route("/different-poll-result")
-interface DifferentPollResult {
-  #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Will use Azure.Core when LongRunningRpcOperation is ready"
+  #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Legacy LRO uses local template"
```
chunyu3, line 137: The above doc may need to change. the status should be changed from `inprogress` to `running`

chunyu3 (approved):

chunyu3:
LGTM except a little modification needed in doc.

packages/cadl-ranch-specs/http/azure/core/lro/rpc/main.tsp
weidongxu-microsoft: Thanks for catching this. Indeed I forgot update the scenarioDoc after mockapi change.

msyyc (approved):

tadelesh (approved):

weidongxu-microsoft:
@timotheeguerin @lmazuel 
Would you kindly take a quick look on the PR? Also, any recommendation on folder name or namespace.

The referred issue https://github.com/Azure/typespec-azure/issues/3108 is not solved yet, but this have been reviewed by almost all SDK languages.
If there is future change to the legacy LRO issue, I will sync it to cadl-ranch.

.changeset/curvy-pens-do.md
timotheeguerin, line 1: > @timotheeguerin @lmazuel
> Would you kindly take a quick look on the PR? Also, any recommendation on folder name or namespace (as it is not really "RPC").
> 
> The referred issue https://github.com/Azure/typespec-azure/issues/3108 is not solved yet, but this PR have been reviewed by almost all SDK languages. Hence would likely to make a available to SDK dev first.
> If there is future change to the legacy LRO issue, I will sync it to cadl-ranch.

commenting here so its easier to reply, `rpc` being the name for non standard resource operations this should be ok having it in the path I think(I would think the name of the operation in azure.core whenever we get that done would have it in the name)

But maybe moving that into another subfolder to say this is the legacy flow `http/azure/lro/rpc/legacy`. That way it leave space for the standard lro rpc operation specs

 We could always rename after, a bit of a pain for everyone but we also don't have the whole picture yet so not sure we can do better.

weidongxu-microsoft: Got it. I will add the legacy segment.

The rpc discussion is pretty long, around the [LongRunningRpcOperation](https://github.com/Azure/typespec-azure/blob/83e1e57/packages/typespec-azure-core/lib/operations.tsp#L48). I thought it would provide building block for these legacy LRO, but turn out it is not (it more like a LongRunningResourceAction, for a non-resource).

weidongxu-microsoft: Also update the scenario case name to "CreateResourcePollViaOperationLocation".

There is other different legacy LRO, e.g. CreateResourcePollViaLocation, CreateResourcePollStatusViaOperationLocationFinalViaLocation :-)

PR #346, opened 2023-06-02
fix doc for http auth test

msyyc:
supplement for https://github.com/Azure/cadl-ranch/pull/341

### DIFF
--- a/packages/cadl-ranch-specs/cadl-ranch-summary.md
+++ b/packages/cadl-ranch-specs/cadl-ranch-summary.md
@@ -37,7 +37,7 @@ Expect error code 403 and error body:
 
 - Endpoint: `get /authentication/http/custom/valid`
 
-Expects header 'Authorization': 'SharedAccessKey valid-key'
+Expects header 'authorization': 'SharedAccessKey valid-key'
 
 ### Authentication_OAuth2_invalid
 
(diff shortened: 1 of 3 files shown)


packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -37,7 +37,7 @@ Expect error code 403 and error body:
 
 - Endpoint: `get /authentication/http/custom/valid`
 
-Expects header 'Authorization': 'SharedAccessKey valid-key'
```
timotheeguerin, line 40: So is that correct, I didn’t see the original pr but aren’t headed supposed to be case insensitive and we should update the validation to relate that. Isn’t that Pascal case version the most common?

weidongxu-microsoft: They are same. Header name is case insensitive.

timotheeguerin: Yeah so aren’t we being too restrictive here. (Aren’t some generator sending using Authorization) Wouldn’t the real fix just to make the validateHeader check be case insensitive

msyyc: Got it. Since the header name is case insensitive, it is not necessary to update the doc.

PR #348, opened 2023-06-02
Add repeatability tests

lmazuel:


packages/cadl-ranch-specs/http/azure/core/traits/mockapi.ts
```diff
@@ -17,6 +17,14 @@ Scenarios.Azure_Core_Traits_smokeTest = passOnSuccess(
     if (req.params.id !== "1") {
       throw new ValidationError("Expected path param id=1", "1", req.params.id);
     }
+
+    if (!("repeatability-request-id" in req.headers)) {
```
srnagar, line 21: Are all headers in `req.headers` always lower case?

packages/cadl-ranch-specs/http/azure/core/traits/mockapi.ts
```diff
@@ -17,6 +17,14 @@ Scenarios.Azure_Core_Traits_smokeTest = passOnSuccess(
...
+      throw new ValidationError("Repeatability-Request-ID is missing", "A UUID string", undefined);
+    }
+    if (!("repeatability-first-sent" in req.headers)) {
```
srnagar, line 24: Do we also want to validate the format and not just check for the presence of the header?

packages/cadl-ranch-specs/http/azure/core/traits/mockapi.ts
lmazuel: Yes, I actually tested explicitly :)
image

packages/cadl-ranch-specs/http/azure/core/traits/mockapi.ts
lmazuel: We could, my JS skills stopped here for this first PR :)

srnagar: It would be great to validate the format as we don't want different languages using different datetime format for this header.

pnpm-lock.yaml
weidongxu-microsoft, line 1: not an expert: should lockfileVersion be updated?

packages/cadl-ranch-specs/http/special-headers/repeatability/mockapi.ts
```diff
@@ -0,0 +1,25 @@
...
+    if (!("repeatability-first-sent" in req.headers)) {
+      throw new ValidationError(
+        "Repeatability-First-Sent is missing",
+        "A date-time in headers format",
+        req.headers["Repeatability-First-Sent"],
+      );
+    }
+    return {
+      status: 204,
+    };
```
weidongxu-microsoft, line 23: Ideally, the test should return 5xx upon first request (simulate service error), and return 204 upon second request (the retry), and verify that value of both header is exactly same on the first request and the retry.

Though how best to do this in mockapi is a bit beyond me.

packages/cadl-ranch-specs/http/special-headers/repeatability/main.tsp
```diff
@@ -0,0 +1,28 @@
...
+@scenario
+@scenarioDoc("""
+Check we recognize Repeatability-Request-ID and Repeatability-First-Sent.
+""")
+@doc("""
+Check we recognize Repeatability-Request-ID and Repeatability-First-Sent.
+""")
+@post
+@route("/doSomething")
+op doSomething(
```
weidongxu-microsoft, line 22: nit, do we have better name than `doSomething`? Maybe just `action` or `post`?

weidongxu-microsoft (approved):

packages/cadl-ranch-specs/http/special-headers/repeatability/main.tsp
```diff
@@ -0,0 +1,28 @@
...
+namespace SpecialHeaders.Repeatability;
+
+@scenario
+@scenarioDoc("""
+Check we recognize Repeatability-Request-ID and Repeatability-First-Sent.
+""")
+@doc("""
+Check we recognize Repeatability-Request-ID and Repeatability-First-Sent.
+""")
+@post
```
tadelesh, line 20: shall we also cover put and delete?

pnpm-lock.yaml
msyyc: there is no dependency update so I think the file shall not be committed.

timotheeguerin: Yeah it shouldn’t be needed but I think this says it use newer version of the lock file so it should be good to update. Got to make sure nobody revert to an older later

packages/cadl-ranch-specs/http/special-headers/repeatability/main.tsp
timotheeguerin: I think depending on all teh scenarios we need (maybe different verb from @tadelesh) comment and different retry flow this would be better to be named describing what it is checking.
Base scenario would probably just check that you are passing the headers and it would succeed. So maybe 
```
immediateSuccess
```

packages/cadl-ranch-specs/http/azure/core/traits/mockapi.ts
lmazuel: We would need to find a JS parser for that, I'm sure that exists :)

packages/cadl-ranch-specs/http/special-headers/repeatability/mockapi.ts
lmazuel: I believe testing that the retry behavior of azure-core is outside of the scope of this test. The scope of this test is to make sure the codegen do not generate parameters, and send random values and "now".

weidongxu-microsoft: Got it. Java will then keep its local mock test for verifying the correctness of retry behavior.

timotheeguerin: @lmazuel I can update this test to include all the other scenarios. Just worried if we just include this one like htat it basically will come down to generators having to reupdate it after

weidongxu-microsoft: I think we can start with this one. My major concern is the name. It could be `doSomething` or `post` (if we plan adding `put` and `delete`).

timotheeguerin: yeah that's true, I'm leaning more towards `immediateSuccess` because that's what shows up in the dashboard and will make more sense as to what it is doing.

packages/cadl-ranch-specs/http/special-headers/repeatability/main.tsp
tadelesh: It should be no difference for the generated code. Since the test for repeatable behavior should not be the goal of cadl-ranch, I just add an operation to azure traits and rename this operation.

packages/cadl-ranch-specs/http/azure/core/traits/mockapi.ts
```diff
@@ -29,6 +29,48 @@ Scenarios.Azure_Core_Traits_smokeTest = passOnSuccess(
         "bar": "456",
         "etag": "11bdc430-65e8-45ad-81d9-8ffa60d55b59",
         "x-ms-client-request-id": req.headers["x-ms-client-request-id"],
+        "repeatability-result": "accepted",
```
lmazuel, line 32: This line should be removed as well

packages/cadl-ranch-specs/http/azure/core/traits/mockapi.ts
```diff
@@ -29,6 +29,48 @@ Scenarios.Azure_Core_Traits_smokeTest = passOnSuccess(
...
+    }
+
+    if (!/^[0-9a-f]{8}-([0-9a-f]{4}-){3}[0-9a-f]{12}$/i.test(req.headers["repeatability-request-id"])) {
+      throw new ValidationError(
+        `Repeatability-Request-ID should be a UUID string`,
+        "A UUID string",
+        req.headers["repeatability-request-id"],
+      );
+    }
+    if (isNaN(Date.parse(req.headers["repeatability-first-sent"]))) {
```
lmazuel, line 58: Can someone confirmed what format `Date.parse` is parsing?

packages/cadl-ranch-specs/http/special-headers/repeatability/main.tsp
```diff
@@ -0,0 +1,39 @@
...
+
+model RepeatableResponse {
+  @doc("The status code.")
+  @statusCode
+  statusCode: 204;
+
+  @visibility("read")
+  @header("Repeatability-Result")
+  @doc("Indicates whether the repeatable request was accepted or rejected.")
+  repeatabilityResult?: "accepted" | "rejected";
```
timotheeguerin, line 21: reason we need this? Wasn't the NoContent all we needed?

Also this feels invalid returning 204 with a body

packages/cadl-ranch-specs/http/azure/core/traits/mockapi.ts
```diff
@@ -33,3 +33,44 @@ Scenarios.Azure_Core_Traits_smokeTest = passOnSuccess(
...
+    }
+
+    if (!("repeatability-request-id" in req.headers)) {
+      throw new ValidationError("Repeatability-Request-ID is missing", "A UUID string", undefined);
+    }
+    if (!("repeatability-first-sent" in req.headers)) {
+      throw new ValidationError("Repeatability-First-Sent is missing", "A date-time in headers format", undefined);
+    }
+
+    if (!/^[0-9a-f]{8}-([0-9a-f]{4}-){3}[0-9a-f]{12}$/i.test(req.headers["repeatability-request-id"])) {
```
timotheeguerin, line 50: can we move this to an helper and reuse it in the `special-headers/repeatability` too.

packages/cadl-ranch-specs/http/azure/core/traits/mockapi.ts
lmazuel: Checked online

lmazuel:
I can't approve my own PR, but works for me

weidongxu-microsoft:
@chunyu3 Opinion from .NET?

packages/cadl-ranch-specs/http/azure/core/traits/mockapi.ts
tadelesh: @timotheeguerin I've moved the validation to `cadl-ranch-api`, could you help to review?

packages/cadl-ranch-api/src/request-validations.ts
```diff
@@ -132,3 +132,32 @@ const coerceDate = (targetObject: Record<string, unknown>): Record<string, unkno
   stringRep = stringRep.replace(/(\d\d\d\d-\d\d-\d\d[Tt]\d\d:\d\d:\d\d)(\.\d{3,7})?([Zz]|[+-]00:00)/g, "$1Z");
   return JSON.parse(stringRep);
 };
+
+/**
+ * Check whether the value follow the right format.
+ */
+export const validateValueFormat = (value: string, format: "uuid" | "rfc7123" | "rfc3339"): void => {
+  switch (format) {
```
chunyu3, line 140: Will we add more format validation in future?

chunyu3 (approved):

chunyu3:
> @chunyu3 Opinion from .NET?

Looks good for .NET

packages/cadl-ranch-api/src/request-validations.ts
tadelesh: If new format check is needed, then we could add more format.

timotheeguerin (approved):

tadelesh (approved):

PR #352, opened 2023-06-07
Bump versions

tadelesh:


tadelesh:
@timotheeguerin In May release, the TSP version update is [a patch version bump](https://github.com/Azure/cadl-ranch/pull/325/files) for cadl-ranch, but this month, it become a minor version bump. Could you double check and help with the release?

timotheeguerin (approved):
Yeah, I think it is ok we bump minor when increasing typespec.

PR #353, opened 2023-06-12
remove `etag` in response body to conform definition

msyyc:


packages/cadl-ranch-specs/http/azure/core/traits/main.tsp
```diff
@@ -66,8 +66,7 @@ Expected response body:
 ```json
 {
   "id": 1,
-  "name": "Madge",
-  "etag": "11bdc430-65e8-45ad-81d9-8ffa60d55b59"
+  "name": "Madge"
```
msyyc, line 69: `etag` is not defined in response body: https://github.com/Azure/cadl-ranch/blob/e91bb0289957f46a2b875b5cf9bf6b3a41d2bb6e/packages/cadl-ranch-specs/http/azure/core/traits/main.tsp#L41-L49

weidongxu-microsoft: Should we improve scenario doc stating that SDK expect `etag` header and `bar` header in response (I assume they are)?

tadelesh: > `etag` is not defined in response body:
> 
> https://github.com/Azure/cadl-ranch/blob/e91bb0289957f46a2b875b5cf9bf6b3a41d2bb6e/packages/cadl-ranch-specs/http/azure/core/traits/main.tsp#L41-L49

The response does not need to define etag header explicitly, it is defined in the operation level with `SupportsConditionalRequests` trait.

tadelesh: I see. Sorry for my mis-understand. I agree to add doc to mention all response headers.

msyyc: doc is updated.

weidongxu-microsoft (approved):

tadelesh (approved):

PR #355, opened 2023-06-13
add encode test for datetime and bytes

tadelesh:

### DIFF
--- a/packages/cadl-ranch-specs/http/encode/bytes/main.tsp
+++ b/packages/cadl-ranch-specs/http/encode/bytes/main.tsp
@@ -19,7 +19,7 @@ namespace Query {
   @route("/default")
   @scenario
   @scenarioDoc("""
-  Test default encode for bytes query parameter.
+  Test default encode (base64) for bytes query parameter.
   Expected query parameter: 
   value=dGVzdA== (base64 encode of test)
   """)
@@ -93,7 +93,7 @@ namespace Property {
   @route("/default")
   @scenario
   @scenarioDoc("""
-  Test operation with request and response model contains bytes properties with default encode.
+  Test operation with request and response model contains bytes properties with default encode (base64).
   Expected request body:
   ```json
   {
@@ -177,7 +177,7 @@ namespace Header {
   @route("/default")
   @scenario
   @scenarioDoc("""
-  Test default encode for bytes header.
+  Test default encode (base64) for bytes header.
   Expected header: 
   value=dGVzdA== (base64 encode of test)
   """)
--- a/packages/cadl-ranch-specs/http/encode/bytes/main.tsp
+++ b/packages/cadl-ranch-specs/http/encode/bytes/main.tsp
@@ -20,7 +20,7 @@ namespace Query {
   @scenario
   @scenarioDoc("""
   Test default encode (base64) for bytes query parameter.
-  Expected query parameter: 
+  Expected query parameter:
   value=dGVzdA== (base64 encode of test)
   """)
   op default(
@@ -32,7 +32,7 @@ namespace Query {
   @scenario
   @scenarioDoc("""
   Test base64 encode for bytes query parameter.
-  Expected query parameter: 
+  Expected query parameter:
   value=dGVzdA== (base64 encode of test)
   """)
   op base64(
@@ -45,7 +45,7 @@ namespace Query {
   @scenario
   @scenarioDoc("""
   Test base64url encode for bytes query parameter.
-  Expected query parameter: 
+  Expected query parameter:
   value=dGVzdA (base64url encode of test)
   """)
   op base64url(
@@ -58,7 +58,7 @@ namespace Query {
   @scenario
   @scenarioDoc("""
   Test base64url encode for bytes array query parameter.
-  Expected query parameter: 
+  Expected query parameter:
   value=dGVzdA, dGVzdA
   """)
   op base64urlArray(
@@ -178,7 +178,7 @@ namespace Header {
   @scenario
   @scenarioDoc("""
   Test default encode (base64) for bytes header.
-  Expected header: 
+  Expected header:
   value=dGVzdA== (base64 encode of test)
   """)
   op default(
@@ -190,7 +190,7 @@ namespace Header {
   @scenario
   @scenarioDoc("""
   Test base64 encode for bytes header.
-  Expected header: 
+  Expected header:
   value=dGVzdA== (base64 encode of test)
   """)
   op base64(
@@ -203,7 +203,7 @@ namespace Header {
   @scenario
   @scenarioDoc("""
   Test base64url encode for bytes header.
-  Expected header: 
+  Expected header:
   value=dGVzdA (base64url encode of test)
   """)
   op base64url(
@@ -216,8 +216,8 @@ namespace Header {
   @scenario
   @scenarioDoc("""
   Test base64url encode for bytes array header.
-  Expected header: 
-  value=dGVzdA, dGVzdA
+  Expected header:
+  value=dGVzdA,dGVzdA
   """)
   op base64urlArray(
     @header({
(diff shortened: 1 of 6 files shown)


packages/cadl-ranch-specs/http/encode/bytes/main.tsp
```diff
@@ -0,0 +1,109 @@
...
+@scenario
+@scenarioDoc("""
+Test different encode for bytes query parameters.
+Expected query parameter: 
+default=dGVzdA== (base64 encode of test)
+base64=dGVzdA==
+base64url=dGVzdA (base64url encode of test)
+base64-array=dGVzdA==, dGVzdA==
+""")
+op query(
```
timotheeguerin, line 26: Can we have one scenario per encoding style like it is in the encode/duration spec

packages/cadl-ranch-specs/http/encode/bytes/main.tsp
```diff
@@ -0,0 +1,109 @@
...
+
+@route("/query")
+@scenario
+@scenarioDoc("""
+Test different encode for bytes query parameters.
+Expected query parameter: 
+default=dGVzdA== (base64 encode of test)
+base64=dGVzdA==
+base64url=dGVzdA (base64url encode of test)
+base64-array=dGVzdA==, dGVzdA==
```
weidongxu-microsoft, line 24: Should we test base64url array, instead of base64 array (if we had to choose one)? As latter is default.

packages/cadl-ranch-specs/http/encode/datetime/main.tsp
```diff
@@ -0,0 +1,124 @@
...
+  @query
+  @encode(DateTimeKnownEncoding.rfc3339)
+  rfc3339: utcDateTime,
+
+  @query
+  @encode(DateTimeKnownEncoding.rfc7231)
+  rfc7231: utcDateTime,
+
+  @query("unix-timestamp")
+  @encode(DateTimeKnownEncoding.unixTimestamp, int32)
```
weidongxu-microsoft, line 40: nit, I assume timestamp is usually long (aka int64)
https://learn.microsoft.com/en-us/dotnet/api/system.datetimeoffset.tounixtimeseconds?view=net-7.0

packages/cadl-ranch-specs/http/encode/datetime/main.tsp
```diff
@@ -0,0 +1,124 @@
...
+  unixTimestamp: utcDateTime,
+
+  @query({
+    name: "rfc3339-array",
+    format: "csv",
+  })
+  rfc3339Array: rfc3339Datetime[]
```
weidongxu-microsoft, line 47: Again, should we instead test an array of non-default encoding?

packages/cadl-ranch-specs/http/encode/datetime/mockapi.ts
```diff
@@ -0,0 +1,47 @@
...
+  }),
+);
+
+Scenarios.Encode_Datetime_Header_get = passOnSuccess(
+  mockapi.get("/encode/datetime/header/get", (req) => {
+    req.expect.containsHeader("default", "Fri, 26 Aug 2022 14:38:00 GMT");
+    req.expect.containsHeader("rfc3339", "2022-08-26T18:38:00.000Z");
+    req.expect.containsHeader("rfc7231", "Fri, 26 Aug 2022 14:38:00 GMT");
+    req.expect.containsHeader("unix-timestamp", "1686566864");
+    req.expect.containsHeader("rfc7231-array", "Fri, 26 Aug 2022 14:38:00 GMT,Fri, 26 Aug 2022 16:38:00 GMT");
```
weidongxu-microsoft, line 42: Something like `["Fri, 26 Aug 2022 14:38:00 GMT", "Fri, 26 Aug 2022 16:38:00 GMT"], "csv"` ?

Or header does not have this method?
I am somewhat worried about the comma in `Fri, 26 Aug 2022 14:38:00 GMT`, as it would likely be percentage encoded in Java.

weidongxu-microsoft (approved):
Let us know if the runtime pass for Python.

packages/cadl-ranch-specs/http/encode/datetime/mockapi.ts
tadelesh: Currently, for header array, Python will not do url-encoding for header, and just contact value with `,`. What will Java do for headers? Maybe I should change to test unix-timestamp.

weidongxu-microsoft: Then how do backend know which comma `,` to break?

tadelesh: I've changed to unix-timestamp to set aside this header encode problem. We may discuss it in another thread.

packages/cadl-ranch-specs/http/encode/bytes/main.tsp
timotheeguerin: it would be nice that the dashboard shows that you support each encoding independently. 

Seeing just that it is supported in the query but not seeing we can encode bytes as base64, base64url feels limited

tadelesh: I see. Is there any better way to define these operations? Decorator could not be used in template param.

timotheeguerin: hhm, what were you trying to reuse, it feels like we do have unfortunatelly  do a lot of duplication

tadelesh: Separated. You may see there are many duplication; I did not find good way to reduce them.

packages/cadl-ranch-specs/http/encode/datetime/main.tsp
```diff
@@ -0,0 +1,282 @@
...
+    @encode(DateTimeKnownEncoding.unixTimestamp, int64)
+    value: utcDateTime
+  ): NoContentResponse;
+
+  @route("/rfc7231-array")
+  @scenario
+  @scenarioDoc("""
+  Test rfc7231 encode for datetime array query parameter.
+  Expected query parameter: 
+  value=Fri, 26 Aug 2022 14:38:00 GMT, Fri, 26 Aug 2022 16:38:00 GMT
```
weidongxu-microsoft, line 78: Same question as csv header, is it a meaningful case?

Do we had to use `format=csv` when data contains comma? Can it be multi or piple?

packages/cadl-ranch-specs/http/encode/datetime/main.tsp
```diff
@@ -0,0 +1,282 @@
...
+@encode(DateTimeKnownEncoding.unixTimestamp, int64)
+scalar unixTimestampDatetime extends utcDateTime;
+
+@operationGroup
+@route("/query")
+namespace Query {
+  @route("/default")
+  @scenario
+  @scenarioDoc("""
+  Test default encode for datetime query parameter.
```
chunyu3, line 25: Can you write the default encoding in the scenarioDoc?
```suggestion
  Test default encode rfc3339 for datetime query parameter.
```

packages/cadl-ranch-specs/http/encode/datetime/main.tsp
tadelesh: No. Header could only use csv.

weidongxu-microsoft: This is query.

tadelesh: For Python, we do url encoding for each of the array item, then contact them with raw `,`. It means the final url will be `k=<url encoded v1>,<url encoded v2>` and it could be parsed by cadl-ranch server correctly. I don't know if this is a common pattern.

tadelesh:
@timotheeguerin I've separated the tests. Do you have any further concern?

timotheeguerin (approved):
@tadelesh sorry missed the update, this looks great! thanks for updating.

packages/cadl-ranch-specs/http/encode/datetime/main.tsp
tadelesh: Change to unixTimestamp to prevent possible encode issue.

PR #357, opened 2023-06-19
Add tests for multiple api version

pshao25:


packages/cadl-ranch-specs/http/server/version/query/main.tsp
```diff
@@ -0,0 +1,33 @@
...
+
+model Options {
+  versioned: Versions;
+}
+
+@scenario
+@scenarioDoc("An simple operation in a parameterized server.")
+@route("/myOp")
+@head
+op myOp(): OkResponse;
```
timotheeguerin, line 33: how does it resolve the version here? there is no reference to it

packages/cadl-ranch-specs/http/server/version/path/main.tsp
```diff
@@ -0,0 +1,40 @@
...
+
+model Options {
+  versioned: Versions;
+}
+
+@scenario
+@scenarioDoc("An simple operation in a parameterized server.")
+@route("/myOp")
+@head
+op myOp(): OkResponse;
```
timotheeguerin, line 40:
```suggestion
@route("/valid")
@head
op valid(): NoContentResponse;
```

or the entire service  can be the scenario instead so it doesn't add extra nodes in the support dashboard

packages/cadl-ranch-specs/http/server/version/path/main.tsp
```diff
@@ -0,0 +1,40 @@
...
+)
+namespace Server.Version.Path;
+
+enum Versions {
+  v1_1: "v1.1",
+  v1_2: "v1.2",
+}
+
+@knownValues(Versions)
+scalar APIVersion extends string;
```
timotheeguerin, line 30:
```suggestion
enum APIVersion {
  v1_1: "v1.1",
  v1_2: "v1.2",
}
```

you don't need known values for Azure services anymore, enum are viewed extensible

packages/cadl-ranch-specs/http/server/version/query/main.tsp
weidongxu-microsoft: Please consider a better name. `myOp` shows on dashboard, and it be meaningless.

Also describe what you want the tester to do in `scenarioDoc`. What's the input (query/body) you are expecting?

pshao25: The generated url at sdk side should contains the version. See my updated description.

packages/cadl-ranch-specs/http/server/version/union/main.tsp
```diff
@@ -0,0 +1,66 @@
...
+Server with multiple api versions and operation with customer-defined api version.
+
+Expected path with api version parameter: /server/version/union/customerDefined?api-version=2.0.0
+""")
+@route("/customerDefinedWithDefault")
+@head
+op customerDefinedWithDefault(
+  @doc("Api version.")
+  @query
+  apiVersion: "2.0.0"
```
weidongxu-microsoft, line 54: Question: is any other language plan to support this?

packages/cadl-ranch-specs/http/server/version/query/main.tsp
```diff
@@ -0,0 +1,37 @@
...
+    @doc("Need to be set as 'http://localhost:3000' in client.")
+    endpoint: url,
+  }
+)
+namespace Server.Version.Query;
+
+enum Versions {
+  v1_1: "v1.1",
+  v1_2: "v1.2",
+}
```
weidongxu-microsoft, line 23: I assume this case already covered it?
https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/resiliency/srv-driven/main.tsp#L40-L47

packages/cadl-ranch-specs/http/server/version/union/main.tsp
```diff
@@ -0,0 +1,66 @@
...
+#suppress "@azure-tools/typespec-azure-core/use-standard-operations" "For testing purpose"
+@doc("Server with multiple api versions and operation with template-defined api version.")
+@scenario
+@scenarioDoc("""
+Server with multiple api versions and operation with template-defined api version.
+
+Expected path with api version parameter: /server/version/union/Pet/dog?api-version=1.0.0
+""")
+@delete
+op templateDefined is ResourceDelete<Pet>;
```
timotheeguerin, line 66: what is this operation, does using azure.core here bring any value?

We should try to keep all azure things into the Azure folder.

packages/cadl-ranch-specs/http/server/version/union/main.tsp
```diff
@@ -0,0 +1,66 @@
...
+
+using Azure.Core;
+using TypeSpec.Http;
+using TypeSpec.Rest;
+using TypeSpec.Versioning;
+
+@doc("Illustrates server with customer-defined and template-defined api versions.")
+@scenarioService("/server/version/union")
+@useDependency(Azure.Core.Versions.v1_0_Preview_1)
+namespace Server.Version.Union;
```
timotheeguerin, line 14:
```suggestion
namespace Server.Version.Union;
```

What is the meaning of `union` here? what exactly do you expect a client to be doing with this spec in the generated code(not the url sent)

packages/cadl-ranch-specs/http/server/version/union/main.tsp
pshao25: This case is for: when customer defined api-version and templated defined api-version are put together, there should be no issues.

packages/cadl-ranch-specs/http/server/version/union/main.tsp
pshao25: `Union` is copied from `authentication/union`, I'm OK if you have a better name. This is a real case, that customer would use template from Azure.Core and define their operation with their own api-version in the same spec file.

packages/cadl-ranch-specs/http/server/version/union/main.tsp
pshao25: Why not?

weidongxu-microsoft: Why should service allowed to specify api-version that potentially different to the global one?

tadelesh: I agree that all the api version should be defined globally.

packages/cadl-ranch-specs/http/server/version/union/main.tsp
tadelesh: What is the relationship between azure core template api version and service api version? I wonder if service should define its own api version instead of reuse the template api version?

packages/cadl-ranch-specs/http/server/version/union/main.tsp
pshao25: See the definition of Operation:
```
op Operation<
  TParams extends TypeSpec.Reflection.Model,
  TResponse,
  Traits extends TypeSpec.Reflection.Model = {},
  TErrorResponse = Azure.Core.Foundations.ErrorResponse
>(
  ...TraitProperties<
    Traits & VersionParameterTrait<ApiVersionParameter>,
    TraitLocation.ApiVersionParameter
  >,
  ...TParams
): TResponse | TErrorResponse;
```

packages/cadl-ranch-specs/http/server/version/union/main.tsp
pshao25: Azure core template defines api version in its implementation. Then if user doesn't use template, he might define his own api version. Put it in another way, user just spread everything in the template, then there would be an apiversion paramter, right?

packages/cadl-ranch-specs/http/server/version/query/main.tsp
pshao25: If I understand correctly, this case doesn't care about how service evolves. I just want to make sure the api version I set would generate correct url. At SDK side, my expectation is when I initiate the client with v1.1, then the url the client would send request is  `/v1.1/server/version/path/valid`

timotheeguerin: This is about the query spec, the path one looks good and does what I expect. The query one says it will send this url
```
Expected path with api version parameter: /server/version/query/valid?api-version=v1.1
```
but there is 0 reference to a parameter called `api-version` in the spec unless I'm missing it.

pshao25: `v1.1` is what I expect user to pass. For e.g., 
```C#
var response = await new QueryClient(host, new QueryClientOptions(QueryClientOptions.ServiceVersion.V1_1)).ValidAsync();
```
When user creates a class, he should indicate that the api version is `v1.1`. The fully generated client and corresponding test is in https://github.com/Azure/autorest.csharp/pull/3514 under folder [test/CadlRanchProjects/server/version/query/src/Generated/](https://github.com/Azure/autorest.csharp/pull/3514/files#diff-5541013fd5e3a4a736ac6e77f52fe25066f59ff294b899a7ca2529615934aa11)

packages/cadl-ranch-specs/http/server/version/union/main.tsp
pshao25: >  Which from what I understand is you have 2 different operations with different versioning scheme.

I don't think we support 2 different versioning values in the same client. For the client generated from this spec, its `api-version` should be either `v1.1` or `v1.2`, and it applies to all the operations.

Because api-version is a special parameter, what I want to test is it is handled correctly no matter it is defined in azure.core, or in raw operation, or in both of them in the same spec. I don't care if its value is different or same (though it cannot be different).

packages/cadl-ranch-specs/http/server/version/union/main.tsp
tadelesh: I mean the API version value should be defined as a global enum.

packages/cadl-ranch-specs/http/server/version/query/main.tsp
pshao25: OK, seems you are saying this is not a valid case?

Our current behavior is if there is no api-version parameter in the operation, we will generate one, otherwise SDK cannot work.

timotheeguerin: Depends on what you mean by valid case. That spec is a valid representation of some http service but it has no version being used.

That part "the SDK cannot work" is what needs to be worked on, not adding magic parameters that will prevent non azure SDK from working in the future and/or conflict with someone that end up depending on it implicitly.

packages/cadl-ranch-specs/http/server/version/union/main.tsp
pshao25: Though Azure.Core templates just render down to a simple operation, api-version is a such special parameter that we once not support this spec. Therefore it is valuable I think.

packages/cadl-ranch-specs/http/server/version/query/main.tsp
pshao25: I edited my comment with the link to [AnmalyDetector](https://github.com/Azure/azure-rest-api-specs/blob/4854f07c1ff8b67930ad5bc6a3042ad73b241946/specification/cognitiveservices/AnomalyDetector/multivariate/routes.tsp).

pshao25: @m-nash  Currently for an operation without api-version parameter like 
```
op valid(): NoContentResponse;
```
We will add the api-version parameter to it. But Tim mentioned it will prevent non azure SDK from working in the future. Do we have plan to support non azure SDK?

packages/cadl-ranch-specs/http/server/version/union/main.tsp
timotheeguerin: so what you are trying to test here is the `Azure.Core.Foundations.ApiVersionParameter`, I don't think doing a mix bring much value, not really sure we are going to see that ever.

So few items:
1. Move the those test under Azure
2. The path one is good
3. Have a AzureCoreApiVersionParameter spec that use that reference with a custom operation. Or use `Azure.Core.Foundation.Operation` with empty objects. Don't include anything extra(no Pet)
4. The query one should be renamed to something like NonStandardQuery.

packages/cadl-ranch-specs/http/server/version/query/main.tsp
timotheeguerin: Like in the spec, the resulting apiVersion paramter is implicitly set by your SDK currently. 

```tsp
op valid(): NoContentResponse;
```

if you add apiVersion query parameter magically and someone generate an SDK but need that parameter they might not add it to their spec as the SDK "worked" and then removing/changing that behavior will break them.

m-nash: @pshao25 as far as I know we made an explicit choice to put emitters in @azure-tools namespace on npm to indicate these emitters are for azure sdk.  Usage outside of azure sdk is not explicitly supported although it might work in some cases.  This doesn't mean we will never support generating non azure sdk, but if we did I would imagine that emitter would go under the @typespec namespace.

@timotheeguerin do you think we should have a linter to just block the case where a customer doesn't define api-version  but is an azure service since this is a requirement for all azure services?  Or are you aware of a case of an azure service that doesn't have an api-version and passed api stewardship?

PR #361, opened 2023-06-21
Add Job in CI checking changes work with latest dev version

timotheeguerin:
Because typespec-azure PR depends on this repo be valid for the latest dev version we need to also make sure nothing gets merged that is not up to the new standard(new warning, deprecation, linter rule etc.)

fix https://github.com/Azure/typespec-azure/issues/3170


.github/workflows/ci.yml
```diff
@@ -77,3 +77,51 @@ jobs:
...
+      - run: npm install -g pnpm@7.27.1
+        name: Install PNPM
+
+      # Install deps and build the bump package first
+      # - run: pnpm install
+      #   name: Install dependencies
+
+      # - run: npm run build
+      #   name: Build
+      #   working-directory: ./typespec-bump-deps
```
tjprescott, line 104: Remove this?

.github/workflows/ci.yml
```diff
@@ -77,3 +77,51 @@ jobs:
...
+
+      - run: npx -y @azure-tools/typespec-bump-deps@latest ./packages/cadl-ranch-specs/package.json ./packages/cadl-ranch/package.json ./packages/cadl-ranch-expect/package.json
```
tjprescott, line 106: Is this actually what pulls the latest from typespec-azure?

timotheeguerin: this will bump all the version in those pacakge.json to be the latest `-dev` available. This is the same package(defined in this repo) used by the emitter for their nightly build

.github/workflows/ci.yml
timotheeguerin: ah yeah, had issue originally because the next command kept trying to load the unbuilt version of the typespec-bump-deps first

iscai-msft (approved):

PR #364, opened 2023-06-28
add test about nextLink for paging operation

msyyc:
https://github.com/Azure/autorest.python/issues/1984

### DIFF
--- a/packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
+++ b/packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
@@ -206,6 +206,41 @@ Expected response body:
 """)
 op listWithPage is global.Azure.Core.Foundations.Operation<{}, global.Azure.Core.Page<User>>;
 
+#suppress "@azure-tools/typespec-azure-core/use-standard-operations" "For testing global.Azure.Core.Page"
+@scenario
+@doc("List with Azure.Core.Page<>. And it contains nextLink.")
+@route("/page/nextLink")
+@scenarioDoc("""
+Should only generate models named User and UserOrder.
+
+Should not generate visible model like Page.
+
+Expected query parameter: api-version=2022-12-01-preview
+
+Expected response body:
+```json
+{
+  "value":[
+     {
+        "id":1,
+        "name":"Madge",
+        "etag": "11bdc430-65e8-45ad-81d9-8ffa60d55b59"
+     },
+     "nextLink": "/azure/core/basic/page/nextLink/containBrace?%24skiptoken={"bar"%3a"foo"}"
+  ]
+}
+{
+  "value":[
+     {
+        "id":1,
+        "name":"Madge",
+        "etag": "11bdc430-65e8-45ad-81d9-8ffa60d55b59"
+     }
+  ]
+}
+""")
+op listWithPageNextLink is global.Azure.Core.Foundations.Operation<{}, global.Azure.Core.Page<User>>;
+
 @friendlyName("{name}ListResults", T)
 @global.Azure.Core.pagedResult
 model CustomPageModel<T> {
--- a/packages/cadl-ranch-specs/http/azure/core/basic/mockapi.ts
+++ b/packages/cadl-ranch-specs/http/azure/core/basic/mockapi.ts
@@ -83,6 +83,23 @@ Scenarios.Azure_Core_Basic_listWithPage = passOnSuccess(
   }),
 );
 
+Scenarios.Azure_Core_Basic_listWithPageNextLink = passOnSuccess([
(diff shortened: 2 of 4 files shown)


packages/cadl-ranch-specs/http/azure/core/basic/mockapi.ts
```diff
@@ -74,14 +74,22 @@ Scenarios.Azure_Core_Basic_list = passOnSuccess(
...
+Scenarios.Azure_Core_Basic_listWithPage = passOnSuccess([
   mockapi.get("/azure/core/basic/page", (req) => {
     const responseBody = {
       value: [validUser],
+      nextLink: '/azure/core/basic/page/nextLink?%24skiptoken={"bar"%3a"foo"}',
     };
     return { status: 200, body: json(responseBody) };
   }),
-);
+  mockapi.get("/azure/core/basic/page/nextLink", (req) => {
```
msyyc, line 85: It won't break existing test since final result is same with before.

packages/cadl-ranch-specs/http/azure/core/basic/mockapi.ts
```diff
@@ -74,14 +74,22 @@ Scenarios.Azure_Core_Basic_list = passOnSuccess(
...
+    req.expect.containsQueryParam("$skiptoken", '{"bar":"foo"}');
+    const responseBody = {
+      value: [],
```
weidongxu-microsoft, line 88: I am a bit uncomfortable with the empty in 2nd page. Assume all SDK pagination can handle it.

weidongxu-microsoft (approved):
@chunyu3 for .NET

packages/cadl-ranch-specs/http/azure/core/basic/mockapi.ts
msyyc: autorest.testserver has [similar testcase](https://github.com/Azure/autorest.testserver/blob/b6195830209d783dcfef44803f4748f35c1241e0/src/test-routes/paging.ts#L65-L92) long time ago so I infer all SDK has supported it.

iscai-msft: It also was originally added to autorest.testserver because a service had this behavior, I can't remember the service though but Python failed on it so we added it as a test

packages/cadl-ranch-specs/http/azure/core/basic/mockapi.ts
iscai-msft: @msyyc I think this should be a separate operation and test. I'm assuming we're explicitly testing empty next links, is that correct? If so we can call the test `emptyNextLink` or somethign like that

msyyc: Got it. Let me add a new test to check it

packages/cadl-ranch-specs/http/azure/core/basic/mockapi.ts
msyyc: Got it. I will update a new check if we support next link contains `{}`

weidongxu-microsoft: Personally, I'd prefer unit test in autorest.python for this particular case of `{}`.

"{} in nextLink" is a too limited case. One can then have "= in nextLink", ", in nextLink", "% in nextLink"... It does not worth it.

If we want to add to cadl-ranch, I'd prefer it cover more characters like `=` etc.

packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
```diff
@@ -206,6 +206,41 @@ Expected response body:
...
+Expected response body:
+```json
+{
+  "value":[
+     {
+        "id":1,
+        "name":"Madge",
+        "etag": "11bdc430-65e8-45ad-81d9-8ffa60d55b59"
+     },
+     "nextLink": "/azure/core/basic/page/nextLink/containBrace?%24skiptoken={"bar"%3a"foo"}"
```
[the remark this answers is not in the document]
weidongxu-microsoft: Technically, `{}` is unsafe (but not reserved) char that "should" be encoded. But practically client/server seems can handle them (my Firefox send them un-encoded...).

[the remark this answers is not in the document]
msyyc: @weidongxu-microsoft @tadelesh, I forward e-mail to you. It seems wrong behavior of service. So....., not sure whether we need to add the odd test.
image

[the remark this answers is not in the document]
tadelesh: I agree not to add this odd test.

PR #366, opened 2023-07-03
add test for request id header

tadelesh:

This test is testing for logic of auto-generated uuid client id header for azure service in dpg. Since all languages' emitter should support it, I added it to cadl-ranch. But I'm not sure where is the good place to put it.

Also, since it is a client-level behavior, I added three sub-clients to the root namespace.


packages/cadl-ranch-specs/http/special-headers/request-id/main.tsp
```diff
@@ -0,0 +1,59 @@
...
+  @scenarioDoc("""
+Standard request id header. SDK should not genreate `requestId` paramerter but use policy to auto-set the header.
+""")
+  @doc("""
+Standard request id header.
+""")
+  @post
+  @route("/standard")
+  op standard(
+    @header("x-ms-request-id")
```
weidongxu-microsoft, line 39: `x-ms-client-request-id` ?

weidongxu-microsoft: I am not sure whether you want to put standard client-request-id together with non-standard client-request-id in same service/client.

As there is only one `RequestIdPolicy`, (I assume) it can only handle one case for client-request-id.

chunyu3, line 39: Is it `x-ms-client-request-id`? typo?

tadelesh: fixed.

tadelesh: I agree that one client with one request id policy. Thus, I use multi-client to bypass this problem. I don't know if there is better solution.

weidongxu-microsoft: Just ignore the standard one? As it already tested in traits.

Even multiple-client might have some implication for different codegen, as they are required to do per-client configure on this RequestIdPolicy. (knowing this, Java should be able to do it, but not sure about other codegen)

packages/cadl-ranch-specs/http/special-headers/request-id/main.tsp
```diff
@@ -0,0 +1,59 @@
...
+  @scenarioDoc("""
+Standard request id header. SDK should not genreate `clientRequestId` paramerter but use policy to auto-set the header.
+""")
+  @doc("""
+Standard request id header.
+""")
+  @post
+  @route("/standard")
+  op standard(
+    @header("x-ms-client-request-id")
```
MaryGao, line 39: https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-core/lib/traits.tsp#L133

Could you also add a test for azure-core - SupportsClientRequestId?

weidongxu-microsoft: https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/azure/core/traits/main.tsp#L62

MaryGao: Thanks for the info, I think we could add a test case for `NoClientRequestId`.

https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-core/lib/traits.tsp#L148

weidongxu-microsoft: `NoClientRequestId` is up to you (and maybe other langauges).

Java's logic is "x-ms-client-request-id" is **always** sent, regardless of TSP or Swagger or whatever. The only exception maybe when find this "client-request-id" header in client.
So Java won't test on `NoClientRequestId`, as it won't ever happen in Java SDK.

tadelesh: I've removed the standard ones and update the doc for traits test.

packages/cadl-ranch-specs/http/special-headers/request-id/main.tsp
```diff
@@ -0,0 +1,27 @@
+import "@typespec/http";
```
iscai-msft, line 1: I think it would be better to have the folder structure as `headers/request-id`. Thoughts @timotheeguerin ?

tadelesh: `request-id` shall be a special header of azure service. I prefer either put it under `azure/request-id` or keep the current `special-headers`.

packages/cadl-ranch-specs/http/special-headers/request-id/main.tsp
tadelesh: @weidongxu-microsoft @MaryGao do you have any further concern?

weidongxu-microsoft (approved):
LGTM

Please check Isabella's comment.

packages/cadl-ranch-specs/http/special-headers/request-id/main.tsp
timotheeguerin: hard to say if it is that standard compared to repatibility headers or not. We can keep it there for now but maybe rename to the full header name `client-request-id

packages/cadl-ranch-specs/http/special-headers/request-id/main.tsp
```diff
@@ -0,0 +1,27 @@
...
+Non-standard request id header. SDK should not genreate `clientRequestId` paramerter but use policy to auto-set the header.
+Expected header parameters:
+- x-ms-client-request-id=<any uuid string>
+""")
+@doc("""
+Non-standard request id header.
+""")
+@get
+@route("/non-standard")
+op nonStandard(
```
timotheeguerin, line 24: I would maybe not say `nonStandard` as the scenario name. EIther change to something like `valid` or better just move `@scenario` and `@scenarioDoc` on the namespace so the whole file is the scenario.

iscai-msft (approved):

packages/cadl-ranch-specs/http/special-headers/client-request-id/main.tsp
```diff
@@ -0,0 +1,27 @@
...
+namespace SpecialHeaders.ClientRequestId;
+
+@doc("""
+Get operation with azure client request id header.
+""")
+@get
+@route("/")
+op get(
+  @header("client-request-id")
+  clientRequestId: string
```
MaryGao, line 26: optional?

MaryGao (approved):
LGTM

chunyu3 (approved):

PR #368, opened 2023-07-05
Add paging tests

pshao25:


packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
```diff
@@ -206,6 +206,120 @@ Expected response body:
...
+  @doc("The name of the input.")
+  inputName: string;
+}
+
+#suppress "@azure-tools/typespec-azure-core/use-standard-operations" "For testing global.Azure.Core.Page"
+@scenario
+@doc("List with int32 parameter Azure.Core.Page<>.")
+@route("/int-page")
+@scenarioDoc("""
+Expected query parameter: api-version=2022-12-01-preview&another=1
```
weidongxu-microsoft, line 250: What's the difference from `ResourceOperations.ResourceList`, except `another=1`?

If the only difference is the query parameter, why it be paging tests, not query param tests?

packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
```diff
@@ -206,6 +206,120 @@ Expected response body:
...
+
+Expected body parameter: {"inputName": "Madge"}
```
weidongxu-microsoft, line 252: Why List have a body?

packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
pshao25: The difference is I give it an additional parameter. In this example it is `ListItemInput` which contains `another`.

> why it be paging tests, not query param tests

Because what I want to test is paging operation could accept these kind of parameters.

packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
pshao25: We have real case from confidential ledger.

weidongxu-microsoft: It's a `@get`!

packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
weidongxu-microsoft: I'd object adding arbitrary query parameter to paging test.

If you want to test query parameter, go to the parameter folder.

packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
pshao25: Yes, it is a get and it has body. And it is a real case from confidential ledger.

packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
pshao25: It is a real case from load testing.

packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
```diff
@@ -206,6 +206,120 @@ Expected response body:
...
+     }
+  ]
+}
+""")
+@get
+op listWithExtensibleEnumParameters is global.Azure.Core.Foundations.Operation<
+  ListItemInput<ListItemInputExtensibleEnum>,
+  global.Azure.Core.Page<User>
+>;
+
```
tadelesh, line 322: Does the generation logic for .NET is different for an operation with query param with an pageable operation with query param? It looks like testing for different types of query param.

pshao25: Yes, we handle paging operation and normal operation separately.

Yeah it is not a must for us to test different types. I added just for more coverage.

tadelesh (approved):

chunyu3 (commented):
Can we put pageable test to separate directory named `pageable`?

packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
```diff
@@ -207,6 +207,56 @@ Expected response body:
...
 op listWithPage is global.Azure.Core.Foundations.Operation<{}, global.Azure.Core.Page<User>>;
 
+@doc("The parameters for listing users.")
+model ListItemInput {
+  @doc("The body of the input.")
+  @body
+  bodyInput: ListItemInputBody;
+
+  @doc("Another query parameter.")
+  @query
```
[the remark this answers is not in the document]
pshao25: Yes, this test is testing paging operation with additional parameters.

chunyu3 (approved):

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -313,6 +313,34 @@ Expected response body:
   ]
 }
 
+### Azure_Core_Basic_listWithParameters
```
iscai-msft, line 316: looks good, I do think it'd be good to make either a paging operation group, or do a subfolder for paging, because we're getting more and more tests. I won't block merging on this, but could you make a PR separating them out? Thanks!

iscai-msft (approved):

PR #371, opened 2023-07-07
Add test for @include over @internal

tadelesh:

This test is from openai real use case: https://github.com/Azure/autorest.csharp/issues/3519. The new added test could not be tested by mock server, different languages' SDK should add test for their generated code.
The logic emitter should accomplish:
1. `isInternal` will return `true` for `InternaIncludelModel` model
2. `isInclude` will return `true` for `InternaIncludelModel` model
3. `InternaIncludelModel` will be included in `getAllModels`
4. SDK should generate `InternaIncludelModel` and make it public to user.

### DIFF
--- a/.changeset/nine-otters-change.md
+++ b/.changeset/nine-otters-change.md
@@ -2,4 +2,4 @@
 "@azure-tools/cadl-ranch-specs": patch
 ---
 
-Add test for @include over @internal
+Add test for `@include` over `@internal`
--- a/.changeset/nine-otters-change.md
+++ b/.changeset/nine-otters-change.md
@@ -1,0 +1,5 @@
+---
+"@azure-tools/cadl-ranch-specs": patch
+---
+
+Add test for @include over @internal
--- a/packages/cadl-ranch-specs/http/azure/client-generator-core/internal/main.tsp
+++ b/packages/cadl-ranch-specs/http/azure/client-generator-core/internal/main.tsp
@@ -27,6 +27,7 @@ model InternalIncludeModel {
   nested: NestedIncludeModel;
 }
 
+@doc("This is a model referred by model decorated with @include. It should be generated and exported.")
 model NestedIncludeModel {
   comment: string;
 }
--- a/packages/cadl-ranch-specs/http/azure/client-generator-core/internal/main.tsp
+++ b/packages/cadl-ranch-specs/http/azure/client-generator-core/internal/main.tsp
@@ -24,6 +24,11 @@ model InternalModel {
 @global.Azure.ClientGenerator.Core.include
 model InternalIncludeModel {
   name: string;
+  nested: NestedIncludeModel;
+}
+
+model NestedIncludeModel {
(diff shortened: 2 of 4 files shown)


.changeset/nine-otters-change.md
```diff
@@ -0,0 +1,5 @@
+---
+"@azure-tools/cadl-ranch-specs": patch
+---
+
+Add test for @include over @internal
```
timotheeguerin, line 5:
```suggestion
Add test for `@include` over `@internal`
```
nit: so it doesn't tag people

packages/cadl-ranch-specs/http/azure/client-generator-core/internal/main.tsp
```diff
@@ -20,6 +20,12 @@ model InternalModel {
   name: string;
 }
 
+@doc("This is a model only used by internal operation. Also, it is decorated with @include. It should be generated and exported.")
+@global.Azure.ClientGenerator.Core.include
+model InternaIncludelModel {
```
timotheeguerin, line 25:
```suggestion
model InternalIncludeModel {
```

typo

packages/cadl-ranch-specs/http/azure/client-generator-core/internal/main.tsp
```diff
@@ -20,6 +20,12 @@ model InternalModel {
...
+  name: string;
```
weidongxu-microsoft, line 26: Should we add a nested model to also test transitivity?

Or, if we don't sure how transitivity works for `@include`, I am find to leave as this.

weidongxu-microsoft (approved):
LGTM

Please take care of Timothee's comment.

PS: whether model is output-only or input-output actually affect Java class (guess .NET too), but that's language problem, no need to worry in tsp test.

packages/cadl-ranch-specs/http/azure/client-generator-core/internal/main.tsp
tadelesh: I think transitivity should be the work of `isInclude` (though it seems current TCGC has not implemented it yet). I've created an [issue](https://github.com/Azure/typespec-azure/issues/3264) to track..

tadelesh:
@msyyc @qiaozha @pshao25 please review from your language's perspective.

packages/cadl-ranch-specs/http/azure/client-generator-core/internal/main.tsp
pshao25: I think we should figure out this issue now.

pshao25 (changes requested):
Checked the user's spec, it has nested model. So let's figure out how to handle it in this PR.

tadelesh:
> Checked the user's spec, it has nested model. So let's figure out how to handle it in this PR.

Added nested model in current test, and a fix for @include transitivity: https://github.com/Azure/typespec-azure/pull/3267.

PR #381, opened 2023-07-26
feat(specs): refactor and add new tests to model inheritance

### LINKED ISSUE #379: feat(test): add test cases for enum type discriminator property
## Description

In swagger, you can define the discriminator property as enum type. This should be supported in typespec as well. It's a reasonable business case.

archerzz:
- split inheritance test cases into 4 sub test cases
- and test cases for single discriminator and enum discriminator

resolve #379


packages/cadl-ranch-specs/http/type/model/inheritance/enum-discriminator/main.tsp
```diff
@@ -0,0 +1,50 @@
...
+Receive model with extensible enum discriminator type.
+Expected response body:
+```json
+
+{"kind": "golden", "weight": 10}
+```
+""")
+@route("/extensible-enum")
+@get
+op getExtensibleEnum(): Dog;
```
weidongxu-microsoft, line 37: `getValid` or `getModel` be enough? `Dog` is not an enum.

archerzz: Yes. Previously I have test cases regarding fixed enum. But it looks like after `0.46.0` there will be error if you use fixed enum as discriminator type, which makes sense. So I deleted thec ases regarding fixed enum. I'll make a change.

archerzz: change is made

packages/cadl-ranch-specs/http/type/model/inheritance/enum-discriminator/main.tsp
```diff
@@ -0,0 +1,50 @@
...
+Send model with extensible enum discriminator type.
+Expected request body:
+```json
+
+{"kind": "golden", "weight": 10}
+```
+""")
+@route("/extensible-enum")
+@put
+op putModel(@body input: Dog): OkResponse;
```
timotheeguerin, line 50: Should we have the same scenarios as for the non enum tests?
- get
- put
- get unknown discriminator
- get missing discriminator property? (not sure this one is really something that can happen so would be ok not having it in either too)

packages/cadl-ranch-specs/http/type/model/inheritance/non-discriminator/main.tsp
```diff
@@ -0,0 +1,53 @@
+import "@typespec/http";
+import "@azure-tools/cadl-ranch-expect";
+
+using TypeSpec.Http;
+
+@doc("Illustrates non-discriminator inheritance model.")
+@scenarioService("/type/model/inheritance/non-discriminator")
+namespace Type.Model.Inheritance.NonDiscriminator;
```
timotheeguerin, line 8:
```suggestion
namespace Type.Model.Inheritance.NoDiscriminator;
```

or `NotDiscriminated` I think would be better naming.

packages/cadl-ranch-specs/http/type/model/inheritance/non-discriminator/main.tsp
```diff
@@ -0,0 +1,53 @@
...
+@scenarioDoc("""
+Generate and send model.
+Expected input body:
+```json
+{"name": "abc", "age": 32, "smart": true}
+```
+""")
+@route("/valid")
+@post
+op postValid(@body input: Siamese): OkResponse;
```
timotheeguerin, line 35:
```suggestion
op postValid(@body input: Siamese): NoContentResponse;
```

packages/cadl-ranch-specs/http/type/model/inheritance/multiple-discriminator/main.tsp
```diff
@@ -0,0 +1,222 @@
...
+@route("/model")
+@scenarioDoc("""
+Generate and send polymorphic model in multiple levels inheritance with 2 discriminators.
+Expected input body:
+```json
+{"age": 1, "kind": "shark",  "sharktype": "goblin"}
+```
+""")
+@put
+op putModel(@body input: Fish): OkResponse;
```
timotheeguerin, line 62:
```suggestion
op putModel(@body input: Fish): NoContentResponse;
```

return NoContentresponse or `void` for operations without body.

packages/cadl-ranch-specs/http/type/model/inheritance/multiple-discriminator/main.tsp
```diff
@@ -0,0 +1,222 @@
...
+@doc("This is base model for polymorphic multiple levels inheritance with a discriminator.")
+@discriminator("kind")
+model Fish {
+  age: int32;
+}
+
+@doc("The second level model in polymorphic multiple levels inheritance and it defines a new discriminator.")
+@discriminator("sharktype")
+model Shark extends Fish {
+  kind: "shark";
```
timotheeguerin, line 19:
```suggestion
  kind: "shark";
  sarktype: string; 
```
recommended way is to still provide the base class with the discirminator property. Omitting it is a legacy feature from when we didn't allow property override

timotheeguerin: maybe we can add a separate test for the 2 syntax in the single-discriminator

packages/cadl-ranch-specs/http/type/model/inheritance/multiple-discriminator/main.tsp
```diff
@@ -0,0 +1,222 @@
+import "@typespec/http";
+import "@azure-tools/cadl-ranch-expect";
+
+using TypeSpec.Http;
+
+@doc("Illustrates multiple level inheritance with multiple discriminators.")
+@scenarioService("/type/model/inheritance/multiple-discriminator")
+namespace Type.Model.Inheritance.MultipleDiscriminator;
```
timotheeguerin, line 8: Would mmaybe rename this one to nested discriminator. Multiple discriminator makes me think that there is a multiple discriminator property on the same model

packages/cadl-ranch-specs/http/type/model/inheritance/enum-discriminator/main.tsp
archerzz: added `getMissingDiscriminator` and `getWrongDiscriminator` cases

packages/cadl-ranch-specs/http/type/model/inheritance/non-discriminator/main.tsp
archerzz: rename the whole scenario `not-discriminated`

packages/cadl-ranch-specs/http/type/model/inheritance/non-discriminator/main.tsp
archerzz: fixed

packages/cadl-ranch-specs/http/type/model/inheritance/multiple-discriminator/main.tsp
archerzz: fixed

packages/cadl-ranch-specs/http/type/model/inheritance/multiple-discriminator/main.tsp
archerzz: done

packages/cadl-ranch-specs/http/type/model/inheritance/multiple-discriminator/main.tsp
archerzz: renamed the whole scenario to `nested-discriminator`.

archerzz:
@timotheeguerin I've adopted your comments. List of changes:
-  add `getMissingDiscriminator` and `getWrongDiscriminator` case to enum discriminator tests
- refactor non-discriminator scenario
    - rename scenario to `not-discriminated`
    - change response type from `OkResponse` to `NoContentResponse`
- fix return type of put operations in multiple scenario
- refactor scenario `multiple-discriminator`
    - rename to `nested-discriminator`
    - improve `Shark` declaration
 - add test case for legacy way of defining base model in `single-discriminator` scenario

packages/cadl-ranch-specs/http/type/model/inheritance/enum-discriminator/main.tsp
archerzz: > Was that an error or a warning? If it is a warning we should maybe just suppress it and add keep test for it. This is something I could see being allowed in very specific cases where all the discriminated types are know for sure

@timotheeguerin I double checked. It should be a warning originally. It's coverted as error in `pnpm validate-secenarios`. Two rules are triggered:
- https://github.com/Azure/typespec-azure/blob/0a95aee0a1d075b161ed2b8953578a81e9ebc611/packages/typespec-azure-core/src/lib.ts#L102C6-L102C33
- https://github.com/Azure/typespec-azure/blob/0a95aee0a1d075b161ed2b8953578a81e9ebc611/packages/typespec-azure-core/src/lib.ts#L121

packages/cadl-ranch-specs/http/type/model/inheritance/nested-discriminator/main.tsp
```diff
@@ -0,0 +1,223 @@
...
+namespace Type.Model.Inheritance.NestedDiscriminator;
+
+@doc("This is base model for polymorphic multiple levels inheritance with a discriminator.")
+@discriminator("kind")
+model Fish {
+  age: int32;
+}
+
+@doc("The second level model in polymorphic multiple levels inheritance and it defines a new discriminator.")
+@discriminator("sharktype")
```
iscai-msft, line 17: @archerzz my understanding was csharp wasn't able to handle multiple discriminators in one inheritance line https://github.com/Azure/typespec-azure/issues/2219

timotheeguerin: they are not but this is still good to keep this test, we just separated it so it doesn't prevent the others from being generated.

That way we have clear documentation that this is not a feature supported in all languages instead of having to figure out why its working sometimes in the future when someone ask for it and we forgot.

archerzz: correct. that's one of the initiatives of the pr. we split the nested discriminator scenario from the original scenario, so that in csharp generator test cases we can select those scenarios we support.

packages/cadl-ranch-specs/http/type/model/inheritance/single-discriminator/main.tsp
```diff
@@ -0,0 +1,171 @@
...
+  friends?: Bird[];
+  hate?: Record<Bird>;
+  partner?: Bird;
+}
+
+@doc("Define a base class in the legacy way. Discriminator property is not explicitly defined in the model.")
+@discriminator("kind")
+model Dinosaur {
+  size: int32;
+}
```
weidongxu-microsoft, line 44: What do you mean by "the legacy way"?

TypeSpec doc https://microsoft.github.io/typespec/standard-library/built-in-decorators#examples-1
```ts
@discriminator("kind")
model Pet{ kind: string }

model Cat extends Pet {kind: "cat", meow: boolean}
model Dog extends Pet  {kind: "dog", bark: boolean}
```

Swagger 2.0
https://swagger.io/specification/v2/
> Adds support for polymorphism. The discriminator is the schema property name that is used to differentiate between other schema that inherit this schema. The property name used MUST be defined at this schema and it MUST be in the required property list. When used, the value MUST be the name of this schema or any schema that inherits it.

So I think TypeSpec at least "recommend" to have discriminator defined as property, and Swagger requires so.

timotheeguerin: I think thats the point of this test, it is a valid representation without the `kind:` property on the base class. It might not be recommended but it is still valid.

weidongxu-microsoft: I see. Just wondering why "the legacy way" come from.

Tried on playground, it will produce valid Swagger. So, Yeah, I am OK with it in test case.

timotheeguerin: it used to be the only way before we supported overriding properties in model inheritance

```
model Foo {
a: string}

model Bar extends Foo {
a: string; // this used to be an error, you couldn't redefine properties from parent models
}
```

timotheeguerin (approved):

weidongxu-microsoft (approved):

tadelesh (approved):

PR #386, opened 2023-07-27
fix-generate-summary-missing-endpoint-url

### LINKED ISSUE #384: generated scenario summary is missing the baseurl
We have defined baseurl in the @ server decorator in the multiclient scenarios PR.
```
@server(
  "http://localhost:3000/client/structure/{client}",
  "",
  {
    client: string,
  }
)
```
And the expected summary should be something like `/client/structure/{client}/one`, but it's now generated as

```
- Endpoints:
  - `get /one`
  - `get /three`
  - `get /three`
  - `get /one`
  - `get /three`
  - `get /three`
```
We should add baseurl info from the tsp compiler result when generating the summary.

qiaozha:

fixes #384


packages/cadl-ranch-expect/src/decorators.ts
```diff
@@ -133,8 +133,20 @@ function getRouteSegments(program: Program, target: Operation | Interface | Name
...
+  return (template ? template : "/") + segments.map((x) => (x.startsWith("/") ? x.substring(1) : x)).join("/");
+}
+
+function getHostTemplate(program: Program, namespace?: Namespace): string | undefined {
+  if (namespace === undefined) {
+    return undefined;
+  }
+  const server = getServers(program, namespace);
+  if (server && server.length === 1) {
+    return server[0].url.split("localhost:3000")[1];
```
timotheeguerin, line 147:
```suggestion
    return new URL(server[0].url).path
```
Or whatever else is the correct property on the URL object for the entire path
I think that would be more reliable.

qiaozha: Actually we can't, there's a case like `{endpoint}//server/path/multiple/{apiVersion}` and if we use URL, it will report invalid url errors.

timotheeguerin: hhm right, could maybe catch the error and then keep the entire server url

packages/cadl-ranch-expect/src/decorators.ts
```diff
@@ -133,8 +133,26 @@ function getRouteSegments(program: Program, target: Operation | Interface | Name
...
+  const server = getServers(program, namespace);
+  if (server && server.length === 1) {
+    if (server[0].url.indexOf("localhost:3000") > -1) {
+      return server[0].url.split("localhost:3000")[1];
+    } else if (server[0].url.indexOf("{endpoint}") > -1) {
+      return server[0].url.split("{endpoint}")[1];
+    } else {
+      return server[0].url;
+    }
+  }
```
qiaozha, line 154: I am changing it this way.

qiaozha: @timotheeguerin one reason I kind of don't like URL is, it automatically encoding something 
image

qiaozha: The above is the result I get after I change the code like 
```typescript
  if (server && server.length === 1) {
    try {
      return new URL(server[0].url).pathname;
    } catch (e) {
      if (server[0].url.indexOf("{endpoint}") > -1) {
        return server[0].url.split("{endpoint}")[1];
      } else {
        return server[0].url;
      }
    }
  }
```

weidongxu-microsoft: 1. It is not actually "host". Seems more like `getRouteSegmentFromServer`?
2. Hope there is more generic solution for `localhost:3000` and `{endpoint}`, but I don't have good solution. Maybe remove URL scheme and start with first `/`, but I guess there will be bug on this as well.

weidongxu-microsoft (commented):
Fine by me.

packages/cadl-ranch-expect/src/decorators.ts
timotheeguerin: yeah I guess that's ok we can revisit if we get into more cases that don't work out

timotheeguerin (approved):

packages/cadl-ranch-expect/src/decorators.ts
qiaozha: changed the name to getRouteSegmentFromServer.

weidongxu-microsoft (approved):

tadelesh (approved):

PR #388, opened 2023-08-02
Make typespec-bump-deps testable

hallipr:


packages/typespec-bump-deps/test/cli.test.ts
timotheeguerin, line 1: would name this file `typespec-bump-deps.test.ts` to test that file

packages/typespec-bump-deps/test/cli.test.ts
```diff
@@ -0,0 +1,35 @@
+import { expect } from "chai";
+import "mocha";
```
timotheeguerin, line 2: Was this because typescript was complaining? I think you might need to add `mocha` to the tsconfig.json instead like https://github.com/microsoft/typespec/blob/4e685d2cd8141099d7337ec60b48a4fe6bb84454/packages/lint/tsconfig.json#L8

hallipr: I just copied the test file from https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-api/test/expectation.test.ts

Importing expect and mocha looks right, no?

timotheeguerin: ah didn't realized that test was done this way, ideally you can just have it once in the tsconfig.json so each test doesn't need this import with side effects.

Or change it to be explicit like

```ts
import { describe, it } from "mocha";
```

hallipr: I think I prefer the explicit import

timotheeguerin (approved):

PR #393, opened 2023-08-03
use patch for the apis

qiaozha:

### DIFF
--- a/.changeset/tender-trees-relax.md
+++ b/.changeset/tender-trees-relax.md
@@ -1,0 +1,4 @@
+---
+"@azure-tools/cadl-ranch-specs": patch
+---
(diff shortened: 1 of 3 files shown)


weidongxu-microsoft (approved):

tadelesh (approved):

.changeset/tender-trees-relax.md
timotheeguerin, line 1: Why did we change that? was it because the mock api defined patch?

It is just a little weird using patch for those, maybe post if we didn't want get

qiaozha: > was it because the mock api defined patch?
Yeah,  I will change it to post then.  Also, I think we should improve the mock api validation script to validate the http method.

PR #398, opened 2023-08-05
Add content negotiation specs

timotheeguerin:


weidongxu-microsoft (commented):
TypeSpec looks good to me (haven't read the mockapi yet).

SDK mostly call it shared-route case.

---

Below discussion relates to the shared-route SDK design.

In this particular case of content negotiation, there could be a question to .NET/Java on protocol API: whether or not there should be a single shared `getAvatar` protocol API (instead of 2 `getAvatarAsPng` and `getAvatarAsJpeg`) as REST API level it is a single API. If so, how to get the name. 

We may use a "client.tsp" to specify a single protocol API (or have emitter try infer such)
```ts
@client({
  name: "SameBodyClient",
  service: ContentNegotiation,
})
interface SameBodyClient {
  @convenientAPI(false, "java")
  getAvatar(@header accept: "image/png" | "image/jpeg"): bytes;

  @protocolAPI(false, "java")
  getAvatarAsPng is SameBody.getAvatarAsPng;

  @protocolAPI(false, "java")
  getAvatarAsJpeg is SameBody.getAvatarAsJpeg;
}
```
(note "scope" is not supported yet)
@srnagar 

@chunyu3  for awareness.

packages/cadl-ranch-specs/http/content-negotiation/main.tsp
```diff
@@ -0,0 +1,63 @@
...
+  model PngImageAsJson {
+    @header contentType: "image/jpeg";
+    content: bytes;
+  }
+
+  @sharedRoute
+  op getAvatarAsPng(@header accept: "image/png"): PngImage;
+
+  @sharedRoute
+  op getAvatarAsJson(@header accept: "application/json"): PngImageAsJson;
```
weidongxu-microsoft, line 62: SDK would either require a "client.tsp", or these `getAvatarAs##` in different interface having different operation name.

As SDK would ignore interface by default, then these API would have conflict API name.

packages/cadl-ranch-specs/http/content-negotiation/mockapi.ts
```diff
@@ -0,0 +1,65 @@
...
+        };
+      case "application/json":
+        return {
+          status: 200,
+          body: json({
+            value: pngFile.toString("base64"),
+          }),
+        };
+      default:
+        throw new ValidationError("Unsupported Accept header", `"image/png" | "image/jpeg"`, req.headers["accept"]);
```
weidongxu-microsoft, line 62:
```suggestion
        throw new ValidationError("Unsupported Accept header", `"image/png" | "application/json"`, req.headers["accept"]);
```

packages/cadl-ranch-specs/http/content-negotiation/mockapi.ts
```diff
@@ -0,0 +1,65 @@
...
+            rawContent: jpegImage,
+          },
+        };
+      default:
+        throw new ValidationError("Unsupported Accept header", `"image/png" | "image/jpeg"`, req.headers["accept"]);
+    }
+  }),
+);
+
+Scenarios.ContentNegotiation_DifferentBody = passOnSuccess(
```
weidongxu-microsoft, line 43: Wondering why "SameBody" uses `passByKey`, while here uses `passOnSuccess`?

packages/cadl-ranch-specs/http/content-negotiation/main.tsp
chunyu3: Seems `SameBody` and `DifferentBody` are the same. It only return differently according to the accept header. I suppose those tow interface will set different input with different content-type.

chunyu3:
> TypeSpec looks good to me (haven't read the mockapi yet).
> 
> SDK mostly call it shared-route case.
> 
> Below discussion relates to the shared-route SDK design.
> 
> In this particular case of content negotiation, there could be a question to .NET/Java on protocol API: whether or not should there be a single shared `getAvatar` protocol API (instead of 2 `getAvatarAsPng` and `getAvatarAsJpeg`) as REST API level it is a single API. If so, how to get the name.
> 
> We may use a "client.tsp" to specify a single protocol API (or have emitter try infer such)
> 
> ```ts
> @client({
>   name: "SameBodyClient",
>   service: ContentNegotiation,
> })
> interface SameBodyClient {
>   // need to disable on python/javascript
>   @convenientAPI(false)
>   @route("same-body")
>   @sharedRoute
>   getAvatar(@header accept: "image/png" | "image/jpeg"): bytes;
> 
>   @protocolAPI(false, "java")
>   getAvatarAsPng is SameBody.getAvatarAsPng;
> 
>   @protocolAPI(false, "java")
>   getAvatarAsJpeg is SameBody.getAvatarAsJpeg;
> }
> ```
> 
> (note "scope" is not supported yet)
> 
> Or we may need an overload of `@protocolAPI` which can specify a method name e.g. `@@protocolAPI(SameBody.getAvatarAsPng, true, "getAvatar")`
> 
> Or we wait for project TOM to solve that... @srnagar
> 
> @chunyu3 for awareness.

Yes. I think we may need to define an original operation `op getAvatar(@header accept: "image/png" | "image/jpeg"): bytes;` in `main.tsp`, and will add above two convenience api  with @sharedRoute in `client.tsp`

packages/cadl-ranch-specs/http/content-negotiation/main.tsp
chunyu3: This testcase is only for multiple-content-type for response.  Shall we add tests for multiple content-type for input body parameter?

timotheeguerin: yeah missed the `@operationGroup` on the namespaces, will add it.

timotheeguerin: @chunyu3 that is a good point but I think it is a different scenario. Different input doesn't require negotiation for the output, it could just be represented as a union of input.

However I'll maybe try to rename this to make sure to leave room for the input scenarios

weidongxu-microsoft:
> I think the combination of those to treat it as "overload" is an hortogonal feature. Having it as I wrote it now might not be what you want in some cases(merge into getAvatar) but it could also be. So I think we can just add the "overload" scenario later a separate entity.

Agree. This is more an orthogonal issue in SDK (and likely going to take time anyway). I've confirmed that Java would generate fine with your current typespec.

weidongxu-microsoft (approved):

iscai-msft (approved):

srnagar:
> TypeSpec looks good to me (haven't read the mockapi yet).
> 
> SDK mostly call it shared-route case.
> 
> Below discussion relates to the shared-route SDK design.
> 
> In this particular case of content negotiation, there could be a question to .NET/Java on protocol API: whether or not should there be a single shared `getAvatar` protocol API (instead of 2 `getAvatarAsPng` and `getAvatarAsJpeg`) as REST API level it is a single API. If so, how to get the name.
> 
> We may use a "client.tsp" to specify a single protocol API (or have emitter try infer such)
> 
> ```ts
> @client({
>   name: "SameBodyClient",
>   service: ContentNegotiation,
> })
> interface SameBodyClient {
>   // need to disable on python/javascript
>   @convenientAPI(false)
>   @route("same-body")
>   @sharedRoute
>   getAvatar(@header accept: "image/png" | "image/jpeg"): bytes;
> 
>   @protocolAPI(false, "java")
>   getAvatarAsPng is SameBody.getAvatarAsPng;
> 
>   @protocolAPI(false, "java")
>   getAvatarAsJpeg is SameBody.getAvatarAsJpeg;
> }
> ```
> 
> (note "scope" is not supported yet)
> 
> Or we may need an overload of `@protocolAPI` which can specify a method name e.g. `@@protocolAPI(SameBody.getAvatarAsPng, true, "getAvatar")`
> 
> Or we wait for project TOM to solve that... @srnagar
> 
> @chunyu3 for awareness.

If there is no `client.tsp` that sets `protocolAPI` to `false` for `getAvatarAsPng` and `getAvatarAsJpg` (like @weidongxu-microsoft has in the example above), then by default, the generator should generate two separate protocol methods for each of the `sharedRoute` methods.

lirenhe:
> 
cc @m-nash.   So, I believe dotnet is following the same to generate two protocol methods for shared route.

PR #402, opened 2023-08-09
Add test for Head operation

chunyu3:


packages/cadl-ranch-specs/http/http-method/head/main.tsp
```diff
@@ -0,0 +1,31 @@
...
+
+@scenario
+@scenarioDoc("""
+Test 404 response.
+
+Expected path parameter `id=invalid`
+""")
+@route("/404")
+@head
+op head404(@path id: string): NoContentResponse;
```
weidongxu-microsoft, line 31: It the test is for `Exists`, Java would like response be `NoContentResponse | NotFoundResponse`, as 404 does not mean an exception.

If it is only for 204 as OK, 404 as exception, I am fine with current operation.

chunyu3: Only 204 is OK response, 404 will be an exception.

weidongxu-microsoft: Ok. Then we probably don't need `@path id`, right?

The 2 route should already tell the difference.

chunyu3: Thanks @weidongxu-microsoft . Updated

weidongxu-microsoft:
@msyyc @qiaozha Any one have other opinion?

I would wonder why need a specific test for HEAD, as I assume emitter naturally supports that. Though I am fine to have it.

chunyu3:
> @msyyc @qiaozha Any one have other opinion?
> 
> I would wonder why need a specific test for HEAD, as I assume emitter naturally supports that. Though I am fine to have it.

We will implement `HeadAsBoolean`. we need add a test project for it.

weidongxu-microsoft:
> We will implement `HeadAsBoolean`. we need add a test project for it.

I didn't get it. If `HeadAsBoolean`, why 404 still be exception? If 404 be exception, how do you get `false` in response?

chunyu3:
> exception? If 404 be exception, how do you get `false` in response?

If no `head-as-boolean` set, 404 will be an exception. If `head-as-boolean` is set, 404 will not be regarded as Error response and return a response.  .NET implemented in this way.

weidongxu-microsoft:
> If no `head-as-boolean` set, 404 will be an exception. If `head-as-boolean` is set, 404 will not be regarded as Error response and return a response. .NET implemented in this way.

OK. Mind you that Batch may want some of their HEAD return boolean, some not.

chunyu3:
> Java is not fond of this global opti

Thanks @weidongxu-microsoft. That's another topic, shall we support global `HeadAsBoolean` (which is current feature requirement, and it is consistent with swagger) or we need to define operation-level flag/decorator for `HeadAsBoolean`.

timotheeguerin:
Yeah we do have plan as Weidong pointed out above(https://github.com/Azure/typespec-azure/issues/3044)  to add in azure core distinction between head meant to check if entity exists and other head. I would prefer we do not add this emitter option and then have different ways to do things and confuse users.

msyyc:
(1) If service wants all HEAD operation result to be boolean, the best way is to support global `head-as-boolean`;
(2) If service only wants some HEAD operation result as boolean, the best way is tsp add specific feature to support it like:
```
@headAsBoolean
op operation(...)
```
Now python has supported [global `head-as-boolean`](https://github.com/Azure/autorest.python/blob/f3ed94c465cb988ea1432d193916ef2902a00ab0/packages/typespec-python/src/lib.ts#L10), and I think we may need to discuss about how to support it in operation level.
As for the testcase, Python keeps open mind since we test it with [different confiuration](https://github.com/Azure/autorest.python/blob/f3ed94c465cb988ea1432d193916ef2902a00ab0/packages/typespec-python/tasks.py#L78-L79)

timotheeguerin:
> (1) If service wants all HEAD operation result to be boolean, the best way is to support global `head-as-boolean`; (2) If service only wants some HEAD operation result as boolean, the best way is tsp add specific feature to support it like:

I don't know if I agree with this either. The fact that they want to return a boolean feels more than just an emitter option. It represent the meaning of this operation. This option means every emitter needs to have it and it affect the logical representation of the SDK which is not what emitter options should really be about. If they want all `head` operations to be represented like that then they should use that pattern everywhere instead of delegating that to the emitter config.

iscai-msft:
I agree with @timotheeguerin , esp if when `ResourceExists` is added and we also have a flag to toggle this behavior, it will lead to confusing outcomes. For example, if I have `head-as-boolean=False`, but then a `ResourceExists`, what should I do?

chunyu3:
> (1) If service wants all HEAD operation result to be boolean, the best way is to support global `head-as-boolean`; (2) If service only wants some HEAD operation result as boolean, the best way is tsp add specific feature to support it like:
> 
> ```
> @headAsBoolean
> op operation(...)
> ```
> 
> Now python has supported [global `head-as-boolean`](https://github.com/Azure/autorest.python/blob/f3ed94c465cb988ea1432d193916ef2902a00ab0/packages/typespec-python/src/lib.ts#L10), and I think we may need to discuss about how to support it in operation level. As for the testcase, Python keeps open mind since there is HEAD in [existing testcase about HEAD](https://github.com/Azure/cadl-ranch/blob/30ae8294f26b090fda612cb441b98b6f9fc700e6/packages/cadl-ranch-specs/http/type/model/visibility/main.tsp#L63C1-L64) and we test it with [different confiuration](https://github.com/Azure/autorest.python/blob/f3ed94c465cb988ea1432d193916ef2902a00ab0/packages/typespec-python/tasks.py#L78-L79)

Yes, there are some head operations, but it tend to test some special functionality. I think it is better to provide separate test case for test Head operation, not mixed them together. We use this separated test cases to test the Head operation related scenarios which is more clear.

weidongxu-microsoft:
Above seems not yet to arrive the agreement on how HEAD for true/false going to be written in TypeSpec.

If you only need to test HEAD (not the true/false case), would this kind of non-boolean case be better?
https://github.com/Azure/azure-sdk-for-java/blob/nikoudsi/batch-typespec/sdk/batch/batch-typespec/routes.tsp#L1812-L1819

PR #403, opened 2023-08-15
add union in response cases

ArcturusZhang:


packages/cadl-ranch-specs/http/type/union/main.tsp
```diff
@@ -79,3 +91,75 @@ This test is testing sending the second union value in named union property.
...
+op receiveString(@body input: Input): ModelWithSimpleUnionPropertyInResponse;
+
+@scenario
+@scenarioDoc("""
+This test is tesing receiving an int array value in simple union property.
+
+Expect request:
+```json
+{ "kind": "int-array" }
+```
```
weidongxu-microsoft, line 120: string or int-array is already in route, why need to have it in request body?

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -3567,6 +3567,78 @@ Expected input body:
 { "property": "hello" }
 ```
 
+### Type_Union_receiveFirstNamedUnionValue
```
archerzz, line 3570:
```suggestion
### Type_Union_receiveFirstTypeOfNamedUnionValue
```

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -3567,6 +3567,78 @@ Expected input body:
...
+{ "kind": "int-array" }
+```
+
+Expect response:
+
+```json
+{ "simpleUnion": [1, 2] }
+```
+
+### Type_Union_receiveSecondNamedUnionValue
```
archerzz, line 3606:
```suggestion
### Type_Union_receiveSecondTypeOfNamedUnionValue
```

packages/cadl-ranch-specs/http/type/union/main.tsp
```diff
@@ -79,3 +91,75 @@ This test is testing sending the second union value in named union property.
...
+
+Expect response:
+```json
+{ "simpleUnion": [1, 2] }
+```
+""")
+@route("/receive/int-array")
+@post
```
MaryGao, line 128: If this is only testing the response could we change `post` to `get` and then remove the request body part?

packages/cadl-ranch-specs/http/type/union/main.tsp
```diff
@@ -79,3 +91,75 @@ This test is testing sending the second union value in named union property.
...
+{ "kind": "model1" }
+```
+
+Expect response:
+```json
+{ "namedUnion": { "name": "model1", "prop1": 1 } }
+```
+""")
+@route("/receive/model1")
+@post
```
MaryGao, line 146: +1 as above

packages/cadl-ranch-specs/http/type/union/main.tsp
ArcturusZhang: Is there a difference?
Get could have a response, Post could also have a response body.
I could remove the request body, that is fine.
I am fine with changing the verb to get but I wonder if there is a difference

MaryGao: I understand both the GET and POST could test the response body but I prefer GET without any request body with following reasons:
- the type info is already in route we don't need to duplicate them in request body
- downstream could focus the existing changing part to do verification in our case that is the response body and no need to care others

I think this is our convertion in cadl-ranch with minimal code to make it work, but feel free to correct me if not!
/cc @weidongxu-microsoft @tadelesh

weidongxu-microsoft: I'd agree with Mary.

My main concern was the request body. Since you are going to remove the request body, and you do not test a POST, a GET is a better solution (IMHO).

There may not be significant difference between POST and GET. But since you are going to modify it anyway, GET is the most simple one when you don't have reason to use other verb.

packages/cadl-ranch-specs/http/type/union/main.tsp
ArcturusZhang: OK I could remove. And how to make two scenarios which one of them send the first type of union, the other send the other type of the union? Should I add another `@scenario` to this operation?

packages/cadl-ranch-specs/http/type/union/main.tsp
ArcturusZhang: body removed and changed to get

packages/cadl-ranch-specs/cadl-ranch-summary.md
ArcturusZhang: I would like the name to keep alignment to the existing cases: https://github.com/Azure/cadl-ranch/blob/a71bd94f358a3ca8e276e2345a87b023b05eda85/packages/cadl-ranch-specs/http/type/union/mockapi.ts#L20
Unless I could change them both otherwise I rather keep these names.

packages/cadl-ranch-specs/http/type/union/main.tsp
```diff
@@ -79,3 +91,55 @@ This test is testing sending the second union value in named union property.
...
+@scenarioDoc("""
+This test is tesing receiving an int array value in simple union property.
+
+Expect response:
+```json
+{ "simpleUnion": [1, 2] }
+```
+""")
+@route("/receive/int-array")
+@post
```
weidongxu-microsoft, line 118: get?

packages/cadl-ranch-specs/http/type/union/main.tsp
```diff
@@ -36,6 +36,18 @@ model ModelWithNamedUnionProperty {
   namedUnion: MyNamedUnion;
 }
 
+model Input {
+  kind: string;
+}
+
```
weidongxu-microsoft, line 42: no longer needed?

packages/cadl-ranch-specs/http/type/union/main.tsp
```diff
@@ -79,3 +91,55 @@ This test is testing sending the second union value in named union property.
...
+@scenarioDoc("""
+This test is tesing receiving the first union value in named union property.
+
+Expect response:
+```json
+{ "namedUnion": { "name": "model1", "prop1": 1 } }
+```
+""")
+@route("/receive/model1")
+@post
```
weidongxu-microsoft, line 131: get?

packages/cadl-ranch-specs/http/type/union/main.tsp
```diff
@@ -79,3 +91,55 @@ This test is testing sending the second union value in named union property.
...
+@scenarioDoc("""
+This test is tesing receiving the second union value in named union property.
+
+Expect response:
+```json
+{ "namedUnion": { "name": "model2", "prop2": 2 } }
+```
+""")
+@route("/receive/model2")
+@post
```
weidongxu-microsoft, line 144: get?

packages/cadl-ranch-specs/http/type/union/mockapi.ts
```diff
@@ -30,3 +30,51 @@ Scenarios.Type_Union_sendSecondNamedUnionValue = passOnSuccess(
     return { status: 200 };
   }),
 );
+
+Scenarios.Type_Union_receiveString = passOnSuccess(
+  mockapi.post("/type/union/receive/string", (req) => {
```
weidongxu-microsoft, line 35: same, "get" here in all

packages/cadl-ranch-specs/http/type/union/mockapi.ts
```diff
@@ -30,3 +30,51 @@ Scenarios.Type_Union_sendSecondNamedUnionValue = passOnSuccess(
...
+    req.expect.bodyEquals({});
```
weidongxu-microsoft, line 36: I don't think you need to "expect body"

ArcturusZhang: should I write `req.expect.bodyEquals(undefined)`?
Now I removed this

packages/cadl-ranch-specs/http/type/union/mockapi.ts
ArcturusZhang: fixed

packages/cadl-ranch-specs/http/type/union/main.tsp
ArcturusZhang: fixed

packages/cadl-ranch-specs/http/type/union/main.tsp
ArcturusZhang: fixed

packages/cadl-ranch-specs/http/type/union/main.tsp
ArcturusZhang: Yes, removed

packages/cadl-ranch-specs/http/type/union/main.tsp
ArcturusZhang: fixed

weidongxu-microsoft:
LGTM for Java.

Still need an approval from interpreted language.

CI need to pass before merge.

ArcturusZhang:
> LGTM for Java.
> 
> Still need an approval from interpreted language.
> 
> CI need to pass before merge.

Thank you! CI should be fixed now

MaryGao (approved):
LGTM for JS.

weidongxu-microsoft (approved):

PR #404, opened 2023-08-17
add test cases to project model name

chunyu3:


packages/cadl-ranch-specs/http/projection/projected-name/main.tsp
```diff
@@ -128,3 +128,51 @@ op parameter(
...
+@operationGroup
+namespace Model {
+  @projectedName("client", "_DontUseMeAsAName_") // Should be ignored
+  @projectedName("csharp", "CSModel")
+  @projectedName("java", "JavaModel")
+  @projectedName("javascript", "TSModel")
+  @projectedName("python", "PythonModel")
+  model ModelWithLanguageProjectedName {
+    @doc("Pass in true")
+    defaultName: boolean;
```
weidongxu-microsoft, line 142: Why do we need this? Any use in test?

packages/cadl-ranch-specs/http/projection/projected-name/main.tsp
```diff
@@ -128,3 +128,51 @@ op parameter(
...
 ): NoContentResponse;
+
+@route("/model")
+@operationGroup
+namespace Model {
+  @projectedName("client", "_DontUseMeAsAName_") // Should be ignored
+  @projectedName("csharp", "CSModel")
+  @projectedName("java", "JavaModel")
+  @projectedName("javascript", "TSModel")
+  @projectedName("python", "PythonModel")
```
qiaozha, line 139: should we add projectedName json for it. though it might not impact any of the dpg codegen, as model name doesn't show up in payload.

chunyu3: Hello @qiaozha It will affect the model name. Codegen will generate the model with the name specified in `projectedName`. e.g csharp will generate a model named `CSModel` instead of `ModelWithLanguageProjectedName`. It will not affect the payload, but it affects the code generated and the signature in the convenience method.

packages/cadl-ranch-specs/http/projection/projected-name/main.tsp
chunyu3: This is not used in test. But spread an empty model in operation, compiler will consider no body parameter. So the model will not be referenced in operation.

packages/cadl-ranch-specs/http/projection/projected-name/main.tsp
qiaozha: Interesting, Are you suggesting csharp will respect projectedName json. I thought it will just ignore json name ?

weidongxu-microsoft (approved):

tadelesh (approved):

PR #409, opened 2023-08-21
Add tests for access and usage.

pshao25:


packages/cadl-ranch-specs/http/azure/client-generator-core/access/main.tsp
```diff
@@ -0,0 +1,191 @@
...
+This scenario contains public operations. It should be generated and exported.
+Expected query parameter: name=<any string>
+Expected response body:
+```json
+{
+  "name": <any string>
+}
+```
+""")
+namespace Public {
```
weidongxu-microsoft, line 26: Why namespace, not interface? May need `@operationGroup`

Also you may want to avoid using possible reserved keyword like `public`/`private`

packages/cadl-ranch-specs/http/azure/client-generator-core/access/main.tsp
```diff
@@ -0,0 +1,191 @@
...
+This scenario contains two operations, one public, another internal. The public one should be generated and exported while the internal one should be generated but not exposed.
+Expected query parameter: name=<any string>
+Expected response body:
+```json
+{
+  "name": <any string>
+}
+```
+""")
+namespace Shared {
```
weidongxu-microsoft, line 109: SharedModel?

packages/cadl-ranch-specs/http/azure/client-generator-core/access/main.tsp
```diff
@@ -0,0 +1,191 @@
...
+  op internal(@query name: string): SharedModel;
+}
+
+@route("/relative")
+@global.Azure.ClientGenerator.Core.operationGroup
+@scenario
+@scenarioDoc("""
+This scenario contains internal operations. All should be generated but not exposed.
+""")
+namespace Relative {
```
weidongxu-microsoft, line 132: What is Relative? It is just all internal models. NestedModel?

packages/cadl-ranch-specs/http/azure/client-generator-core/usage/main.tsp
```diff
@@ -0,0 +1,67 @@
...
+@doc("Test for internal decorator.")
+@supportedBy("dpg")
+@scenarioService("/azure/client-generator-core/usage")
+namespace _Specs_.Azure.ClientGenerator.Core.Usage;
+
+@scenario
+@scenarioDoc("""
+This scenario contains two public operations. Both should be generated and exported.
+The models are override to roundtrip, so they should be generated and exported as well.
+""")
```
weidongxu-microsoft, line 17: It has nothing to do about roundtrip.

A model of `input + output` could be just an input to op1, and output to op2. It does not need to be roundtrip at all (of course, roundtrip means `intput + output`, but not reverse).

packages/cadl-ranch-specs/http/azure/client-generator-core/usage/main.tsp
```diff
@@ -0,0 +1,67 @@
...
+    "name": <any string>
+  }
+  ```
+  """)
+  @route("/outputToRoundTrip")
+  @get
+  op outputToRoundTrip(): OutputModel;
+}
+
+namespace Orphan {
```
weidongxu-microsoft, line 60: No need for namespace.

packages/cadl-ranch-specs/http/azure/client-generator-core/access/main.tsp
pshao25: I just want to group relative model and operation together. Any problem for the namespace?

packages/cadl-ranch-specs/http/azure/client-generator-core/access/main.tsp
pshao25: It just means the models here are related, like A is B's property, A is B's base model, etc. I didn't figure out a better name.

packages/cadl-ranch-specs/http/azure/client-generator-core/usage/main.tsp
pshao25: But if we add `@usage(Usage.input | Usage.output)` we will generate the model the same as a roundtrip model, right?

weidongxu-microsoft: No, that depends on whether emitter has additional process for roundtrip model.

If you test usage, not roundtrip, just don't mention roundtrip,

packages/cadl-ranch-specs/http/azure/client-generator-core/access/main.tsp
weidongxu-microsoft: Likely won't affect Java (but all operations and models would get to a single client, without `@operationGroup`, so the test and test client won't be grouped).

Not sure about other languages.

pshao25: I believe it is valid spec, right? If so, other languages should support.

weidongxu-microsoft: Sure, we will require all language approve then.

packages/cadl-ranch-specs/http/azure/client-generator-core/access/main.tsp
pshao25: SharedModel is already a model name below. Maybe "SharedModelInOperation"?

chunyu3 (approved):

weidongxu-microsoft (approved):
Approved by Isabella and Crystal.

merge it.

PR #410, opened 2023-08-22
spec, add test case for pageable, focus on maxpagesize and nextLink

weidongxu-microsoft:
I currently put the test case in payload folder, as part of it tests the `nextLink`.

Though another part of the test is on `maxpagesize` query parameter (and emitter's handling of it).

---

PS: later we may add another case here, as pageable but without `nextLink` in response. We saw some backend write this (if new service, API would likely advise them to add `nextLink` just not send it; not able to do this to old service).
```ts
@pagedResult
model Page {
  @items
  value: Resource[];
}
```

### DIFF
--- a/packages/cadl-ranch-specs/cadl-ranch-summary.md
+++ b/packages/cadl-ranch-specs/cadl-ranch-summary.md
@@ -1412,6 +1412,49 @@ Scenario that returns a different file encoding depending on the accept header.
 - image/png return a png image
 - image/jpeg return a jpeg image
 
+### Payload_Pageable_list
+
+- Endpoint: `get /payload/pageable`
+
+List users.
+
+Expected query parameter:
+skip=5
+maxpagesize=3
+
+Expected response body:
+
+```json
+{
+  "value": [
+    {
+      "name": "user5"
+    },
+    {
+      "name": "user6"
+    },
+    {
+      "name": "user7"
+    }
+  ],
+  "nextLink": "{endpoint}//payload/pageable?skip=8&maxpagesize=3"
+}
+```
+
+Expected query parameter:
+skip=8
+maxpagesize=3
+
+```json
+{
+  "value": [
+    {
+      "name": "user8"
+    }
+  ]
+}
+```
+
 ### Projection_ProjectedName_operation
 
 - Endpoint: `post /projection/projected-name/operation`
--- a/packages/cadl-ranch-specs/cadl-ranch-summary.md
+++ b/packages/cadl-ranch-specs/cadl-ranch-summary.md
@@ -1418,6 +1418,8 @@ Scenario that returns a different file encoding depending on the accept header.
 
 List users.
 
+SDK may hide the "maxpagesize" from API signature. The functionality of "maxpagesize" could be in related language Page model.
+
 Expected query parameter:
 skip=5
 maxpagesize=3
--- a/packages/cadl-ranch-specs/cadl-ranch-summary.md
+++ b/packages/cadl-ranch-specs/cadl-ranch-summary.md
@@ -1421,7 +1421,6 @@ List users.
 SDK may hide the "maxpagesize" from API signature. The functionality of "maxpagesize" could be in related language Page model.
 
 Expected query parameter:
-skip=5
 maxpagesize=3
 
 Expected response body:
@@ -1439,12 +1438,12 @@ Expected response body:
       "name": "user7"
     }
   ],
-  "nextLink": "{endpoint}//payload/pageable?skip=8&maxpagesize=3"
+  "nextLink": "{endpoint}//payload/pageable?skipToken=name-user7&maxpagesize=3"
 }
 ```
 
 Expected query parameter:
-skip=8
+skipToken=name-user7
 maxpagesize=3
 
 ```json
--- a/packages/cadl-ranch-specs/http/payload/pageable/main.tsp
+++ b/packages/cadl-ranch-specs/http/payload/pageable/main.tsp
@@ -1,6 +1,6 @@
 import "@azure-tools/cadl-ranch-expect";
 import "@azure-tools/typespec-azure-core";
-import "@typespec/versioning"
+import "@typespec/versioning";
 
 using TypeSpec.Versioning;
 using Azure.Core;
@@ -12,7 +12,7 @@ using Azure.Core;
 namespace Payload.Pageable;
 
 model User {
-  name: string
+  name: string;
 }
 
 @scenario
--- a/packages/cadl-ranch-specs/http/payload/pageable/main.tsp
+++ b/packages/cadl-ranch-specs/http/payload/pageable/main.tsp
@@ -22,6 +22,8 @@ model User {
 @scenarioDoc("""
 List users.
 
+SDK may hide the "maxpagesize" from API signature. The functionality of "maxpagesize" could be in related language Page model.
+
 Expected query parameter:
 skip=5
 maxpagesize=3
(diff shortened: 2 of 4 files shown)


packages/cadl-ranch-specs/http/payload/pageable/main.tsp
```diff
@@ -0,0 +1,64 @@
...
+  "value":[
+    {
+      "name":"user8"
+    }
+  ]
+}
+```
+""")
+@doc("List users")
+op list(...SkipQueryParameter, ...MaxPageSizeQueryParameter): Page<User>;
```
weidongxu-microsoft, line 64: I didn't use Azure.Core operation template.

I tried to write this case without model from Azure.Core, but Page there has further ref of `ResourceLocation` template.
https://github.com/Azure/typespec-azure/blob/ac3dc1566e3dea2e245f2e0034e42fcccbf6d2f9/packages/typespec-azure-core/lib/models.tsp#L14-L25

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -1412,6 +1412,49 @@ Scenario that returns a different file encoding depending on the accept header.
...
+      "name": "user7"
+    }
+  ],
+  "nextLink": "{endpoint}//payload/pageable?skip=8&maxpagesize=3"
+}
+```
+
+Expected query parameter:
+skip=8
+maxpagesize=3
```
weidongxu-microsoft, line 1446: SDK may need to be careful not to send `maxpagesize` twice here, as `nextLink` already have `maxpagesize` parameter

iscai-msft: are languages generating the `maxpagesize` parameter for DPG? I know Python isn't because we want to expose the ability to toggle page size through our paging model. Wondering where other languages stand cc @joheredi @srnagar @m-nash

iscai-msft (approved):

packages/cadl-ranch-specs/cadl-ranch-summary.md
weidongxu-microsoft: Java and .NET also hide it from parameter. The page mode would handle them (for Java it is `PagedIterable.iteratorByPage(pageSize)`. I think JS is under discussion as well.

That is the reason I write this in doc
```
SDK may hide the "maxpagesize" from API signature. The functionality of "maxpagesize" could be in related language Page model.
```

I use "may" so the case here won't dictate what SDK should behavior, but at least give a hint here.

chunyu3 (approved):

qiaozha (commented):
Not sure what's the behavior should look like if both top and maxpagesize are specified ? And is it usually the case that top will be used with skip, I kind of don't see maxpagesize being used with skip a lot ?

weidongxu-microsoft:
> Not sure what's the behavior should look like if both top and maxpagesize are specified ? And is it usually the case that top will be used with skip, I kind of don't see maxpagesize being used with skip a lot ?

My personal understanding of "top" is that is be the max number of resources returned by all these paginations. E.g. if service has 1000, top be 150, maxpagesize be 100, it gets 2 pages, 100+50.

I skip "top" here because there seems pending discussion, and it is not the focus of my test anyway. (and yes, it could be ambiguous with "maxpagesize", from other sources)

"skip" is actually a common one, if services done this via "client side driven" pagination.
https://learn.microsoft.com/en-us/odata/webapi-8/fundamentals/client-driven-paging

I can use "skipToken" instead, though I assume "skip" does not affect our test. For 2nd page, just take nextLink for it.

qiaozha (approved):

weidongxu-microsoft:
I've switched to use `skipToken` for paging, as focus here is not on `skip` parameter.

Personal understanding of the origin of `maxpagesize` is from odata preference header (it is query now)
https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#_Toc31358876

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -1412,6 +1412,50 @@ Scenario that returns a different file encoding depending on the accept header.
...
 - image/jpeg return a jpeg image
 
+### Payload_Pageable_list
+
+- Endpoint: `get /payload/pageable`
+
+List users.
+
+SDK may hide the "maxpagesize" from API signature. The functionality of "maxpagesize" could be in related language Page model.
+
```
MaryGao, line 1422: @mikekistler Could you help review this test case? And I have below two questions.

- Server side may not support the `maxpagesize` so even we passed the value for this param the returned page size may or may not follow that, right?
- Regarding the client side if the tsp defined this query param, should we take this param as a regular one or hide it per language's implementation?

mikekistler: If the service does not support `maxpagesize` then it should not be defined in the REST API. And if a client sends a query parameter that is not recognized/supported by the service, the service should fail the request with a 400 response code.

If the service does support `maxpagesize` and defines in the REST API, I don't know why we would hide it in the client library.

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -1412,6 +1412,50 @@ Scenario that returns a different file encoding depending on the accept header.
...
+    {
+      "name": "user6"
+    },
+    {
+      "name": "user7"
+    }
+  ],
+  "nextLink": "{endpoint}//payload/pageable?skipToken=name-user7&maxpagesize=3"
+}
+```
```
mikekistler, line 1443: Note that it would not be wrong to return only two elements with a nextLink -- `maxpagesize` is a _maximum_ and the service is allowed use a smaller page size.

mikekistler (commented):
pageable test looks good. 👍

packages/cadl-ranch-specs/cadl-ranch-summary.md
MaryGao: @weidongxu-microsoft  Not sure the summary `SDK may hide the "maxpagesize" from API signature.` is proper?

weidongxu-microsoft: Mary, here is what Java does

API signature
```java
PagedIterable<User> list()
```

Hence `hide the "maxpagesize" from API signature` in summary. I use "may" because it may be language dependent.

Here the impl
```java
        return new PagedIterable<>(
                (pageSize) -> {
                    RequestOptions requestOptionsLocal = requestOptions == null ? new RequestOptions() : requestOptions;
                    if (pageSize != null) {
                        requestOptionsLocal.addRequestCallback(
                                requestLocal -> {
                                    UrlBuilder urlBuilder = UrlBuilder.parse(requestLocal.getUrl());
                                    urlBuilder.setQueryParameter("maxpagesize", String.valueOf(pageSize));
                                    requestLocal.setUrl(urlBuilder.toString());
                                });
                    }
                    return listSinglePage(requestOptionsLocal);
                },
                (nextLink, pageSize) -> {
                    RequestOptions requestOptionsLocal = new RequestOptions();
                    requestOptionsLocal.setContext(requestOptionsForNextPage.getContext());
                    if (pageSize != null) {
                        requestOptionsLocal.addRequestCallback(
                                requestLocal -> {
                                    UrlBuilder urlBuilder = UrlBuilder.parse(requestLocal.getUrl());
                                    urlBuilder.setQueryParameter("maxpagesize", String.valueOf(pageSize));
                                    requestLocal.setUrl(urlBuilder.toString());
                                });
                    }
                    return listNextSinglePage(nextLink, requestOptionsLocal);
                });
```

It is pretty long, but the idea is that it sets the "maxpagesize" from `pageSize` of `PagedIterable` (when available).

If TypeSpec of Swagger of the API does not have "maxpagesize" query parameter, none of the `maxpagesize` code will be generated, hence SDK won't pass "maxpagesize" to server.

MaryGao: Got it! Thanks!

PR #413, opened 2023-09-04
[etag] add test case for scenario that only one kind of conditional request header is defined

msyyc:
some service only define `if-match` header instead of defining both of `if-match` and `if-none-match`.
Related issue https://github.com/Azure/autorest.python/pull/2086


weidongxu-microsoft (commented):
OK for Java

tadelesh (approved):

weidongxu-microsoft (approved):

PR #415, opened 2023-09-04
specs, LRO, use pollingOperation to customize non-standard LRO

weidongxu-microsoft:
`@pollingOperation(CreateResourcePollViaOperationLocation.getJob)` require release of https://github.com/Azure/typespec-azure/issues/3410 in 0.34.0 or +

Before that, this will not have effect, as parameter map is not provided.


packages/cadl-ranch-specs/http/azure/core/lro/rpc-legacy/main.tsp
```diff
@@ -138,10 +137,12 @@ Expected response body:
 """)
 @route("/create-resource-poll-via-operation-location")
 interface CreateResourcePollViaOperationLocation {
+  @doc("Poll a Job")
+  getJob is ResourceOperations.ResourceRead<JobResult>;
+
   #suppress "@azure-tools/typespec-azure-core/use-standard-operations" "Legacy LRO uses local template"
-  #suppress "@azure-tools/typespec-azure-core/long-running-polling-operation-required" "Operation link pollingOperation is not required for getLroMetadata"
+  @pollingOperation(CreateResourcePollViaOperationLocation.getJob)
```
weidongxu-microsoft, line 144: There is a bit strange of `getLroMetadata` API, that the `logicalResult` is same as `JobResult`, but it is an anomymous model.

emitter had to call `getEffectiveModelType` to get a named model (but we probably should call it anyway).

tadelesh (approved):

iscai-msft (approved):

PR #421, opened 2023-09-18
deprecate internal test

iscai-msft:


timotheeguerin (commented):
I feel like we need to just remove those. Don’t think there is a point keeping deprecated features as scenarios to support

weidongxu-microsoft (approved):
Or should we just remove the test?

PR #422, opened 2023-09-21
Bytes in body

timotheeguerin:


iscai-msft (approved):

packages/cadl-ranch-specs/http/encode/bytes/main.tsp
```diff
@@ -226,3 +226,139 @@ namespace Header {
...
+  };
+
+  @route("/base64url")
+  @scenario
+  @scenarioDoc("""
+  Test base64url encode for bytes header.
+  Expected header:
+  value=dGVzdA (base64url encode of test)
+  """)
+  op base64url(): base64urlBytes;
```
weidongxu-microsoft, line 363: I am not sure whether this is a valid case as HTTP (not only SDK).

What is the response headers? `content-type: application/octet-stream`? But where is the encoding specified? `content-transfer-encoding` seems not have a base64url.
https://www.w3.org/Protocols/rfc1341/5_Content-Transfer-Encoding.html

In HTTP context, TypeSpec operation is not only for SDK. The Internet need to understand the data.

timotheeguerin: This is returning json. By default content type for http library is always json

timotheeguerin: It should just be a JSON string(wrapper in quotes) that is the bytes64url encoded. If it isn’t I might have made a mistake in the implementation

weidongxu-microsoft: I see. Yeah, this makes sense. Will double check in codegen. Thanks!

weidongxu-microsoft: So I assume if user write
```ts
  op octetStream(): bytes
```
It is also Json string?

User must write this for octet-stream
```ts
  op octetStream(): {
    @header
    contentType: "application/octet-stream";

    @body
    value: bytes;
  };
```

timotheeguerin: Indeed, we did go back and forth with this and if we should infer the content type or not. But being always json by default ends up being much cleaner. A lot of times you’ll want a specific content type anyway if you pass a file.

weidongxu-microsoft:
attempting an improvement/fix
https://github.com/Azure/cadl-ranch/pull/424

pretty unsure about the quality, as Java emitter is in bad shape around this, at present :-)

PR #429, opened 2023-09-29
Unions redesign

### LINKED ISSUE #446: Add test for support of union with consistent type as extensible enum
In particular this:
```typescript
union KnownErrorCode {
  // This is not found
  "NotFound",
  InvalidName: "Invalid",
  string
} 
```

should generate an extensible enum once https://github.com/Azure/typespec-azure/issues/3813 is done.

Was discussed with @johanste and @bterlson 

[Full picture on Loop](https://microsoft.sharepoint.com/:fl:/s/d40daadc-c0b7-46ef-8088-d7b3a5bad5f8/EYXTfObxOKJCodhiChGfJ-ABHlnWX1a4fBuwK2YkXa2KIA?e=8mVnB5&nav=cz0lMkZzaXRlcyUyRmQ0MGRhYWRjLWMwYjctNDZlZi04MDg4LWQ3YjNhNWJhZDVmOCZkPWIlMjE1LW9vVDlZZjJrZU1qNE8xT1JQdmJ2NTBieVRvNjd0SWdHSC1xOHA0dmFISkRTcUR1eHVyU2F0WU9IT3pfT1VJJmY9MDFLS09FQ1k0RjJONk9ONEpZVUpCS0RXRENCSUlaNko3QSZjPSUyRiZhPUxvb3BBcHAmcD0lNDBmbHVpZHglMkZsb29wLXBhZ2UtY29udGFpbmVyJng9JTdCJTIydyUyMiUzQSUyMlQwUlRVSHh0YVdOeWIzTnZablF1YzJoaGNtVndiMmx1ZEM1amIyMThZaUUxTFc5dlZEbFpaakpyWlUxcU5FOHhUMUpRZG1KMk5UQmllVlJ2TmpkMFNXZEhTQzF4T0hBMGRtRklTa1JUY1VSMWVIVnlVMkYwV1U5SVQzcGZUMVZKZkRBeFMwdFBSVU5aTjFoYU5FRTNRVFZHV1ZKR1FqTkJWRlUzVUVOSlVqVTJURVElM0QlMjIlMkMlMjJpJTIyJTNBJTIyYmQ3ZDFhNzMtOGI2Ni00YTg5LWEyMTMtOTI5MDQxODIxMzFlJTIyJTdE)

timotheeguerin:
fix  #446
As part of the openai initiative unions have proven problematic for some SDK, this redesign the union scernarios to cover an aggressive set of options. Most of those scenarios have been taken directly from something found in openai spec. With some additional coverage

## Old scenarios
The existing scenarios are a little hard to see what union cases they actually test for 
image

## New scenarios

each scenario has a get and send
- StringsOnly: union of string literal
- StringsExensible: union of string literal and open string
- StringsExtensibleNamed: Named union with named variant
- IntsOnly: union of ints
- EnumsOnly: union of 2 different enums
- FloatsOnly: union of floats
- ModelsOnly: union of models
- MixedLiterals: string, int, float, boolean literals
- Mixed: union of literal, string, models, array


packages/cadl-ranch-specs/http/type/union/main.tsp
ArcturusZhang, line 1: These changes look good to me, but these cases are all unions in model properties. We could have unions in a lot of other places - could we take this opportunity to make a hierarchy to the union cases like we have in others?
For instance, this could be `http/type/union/property`

timotheeguerin: yes this is a good point, my idea in general was to have the following:
1. one spec per "type" which test all the case of that type and usually we have included that type directly as the body
2. then 1 or 2 test for can I use that type in that context like in the ModelProperties/Types it cover all the types you can use. But I am not sure it is worth to duplicate all the cases in 1. for all the potential context.

packages/cadl-ranch-specs/http/type/union/main.tsp
```diff
@@ -3,139 +3,199 @@ import "@azure-tools/cadl-ranch-expect";
...
+      """
+      {
+        "string": "test",
+        "array": ["test1", "test2"]
+      }
+      """,
+      StringAndArrayCases
+    > {}
+
+alias MixedTypesUnion = Cat | "a" | 2 | 3.3 | true;
```
weidongxu-microsoft, line 118: Emm, so this is a mix of model and literal (not model and scala).

packages/cadl-ranch-specs/http/type/union/main.tsp
```diff
@@ -3,139 +3,199 @@ import "@azure-tools/cadl-ranch-expect";
...
+ * Describe union of string "a" | "b" | "c"
+ */
+@route("/strings-only")
+interface StringsOnly extends GetAndSend<"a" | "b" | "c", "\"b\""> {}
+
+/**
+ * Describe union of string string | "b" | "c"
+ */
+@route("/string-extensible")
+interface StringExtensible extends GetAndSend<string | "b" | "c", "\"custom\""> {}
```
weidongxu-microsoft, line 40: so this could be the new extensible enum :-)

weidongxu-microsoft (approved):

packages/cadl-ranch-specs/http/type/union/main.tsp
timotheeguerin: Right maybe can split into mixed literals and mixed types

packages/cadl-ranch-specs/http/type/union/main.tsp
timotheeguerin: Indeed, that’s what Laurent just filed as an issue today. If you use the named version as below you basically have an extensible enum being explicit

packages/cadl-ranch-specs/http/type/union/main.tsp
weidongxu-microsoft: Yeah, maybe we can just insert a scala to this MixedTypesUnion as well :-)

weidongxu-microsoft: Though Java likely going to have a problem with NumericLiteral. It is hard to tell whether the literal is an integer or float. <-- that is to be solved within Java, maybe always use float.

timotheeguerin: yeah though I can see emitters being able to do a better job if its a union of literals vs a union of basically anything(which approach just putting `unknown`)

iscai-msft (approved):

PR #432, opened 2023-10-02
Special words expansion

timotheeguerin:
Redesign and expansion of the reserved words spec.

Taken the reserved works collection for the python extension. Feel free to add more to the list.

```
and
as
assert
async
await
break
class
constructor
continue
def
del
elif
else
except
exec
finally
for
from
global
if
import
in
is
lambda
not
or
pass
raise
return
try
while
with
yield
```

As well as for parameters:
```
cancellationToken
```

And for model properties

```
[property with same name as model]
````

### DIFF
--- a/packages/cadl-ranch-specs/http/special-words/main.tsp
+++ b/packages/cadl-ranch-specs/http/special-words/main.tsp
@@ -1,84 +1,237 @@
 import "@typespec/http";
 import "@azure-tools/cadl-ranch-expect";
 import "@azure-tools/typespec-client-generator-core";
+import "./dec.js";
 
 using TypeSpec.Http;
 using Azure.ClientGenerator.Core;
 
-@doc("Illustrates operation, parameter and property have name of special words or characters.")
+/**
+ * Scenarios to verify that reserved words can be used in service and generators will handle it appropriately.
+ *
+ * Current list of special words
+ * ```txt
+ * and
+ * as
+ * assert
+ * async
+ * await
+ * break
+ * class
+ * constructor
+ * continue
+ * def
+ * del
+ * elif
+ * else
+ * except
+ * exec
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: packages/cadl-ranch-specs/http/special-words/main.tsp [hunk 1]
```diff
@@ -1,84 +1,237 @@
...
+@route("/operations")
 @operationGroup
-interface Operation {
-  @scenario
-  @scenarioDoc("A operation name of `for` should work.")
-  @get
-  @route("/for")
-  for(): void;
+interface Operations {
+  @opNameScenario("and") and(): void;
```
iscai-msft, line 58: qq: what is `@opNameScenario` vs `@scenario`?

iscai-msft (approved):

packages/cadl-ranch-specs/http/special-words/main.tsp
timotheeguerin: yeah just made some simple decorators so this spec doesn't look like a big mess as everything is about the same x100

packages/cadl-ranch-specs/http/special-words/mockapi.ts
```diff
@@ -1,61 +1,164 @@
...
-      status: 204,
-    };
-  }),
-);
+// ------------------------------------------------------------------------
+// Operation name scenarios
+// ------------------------------------------------------------------------
+function opNameScenario(name: string) {
+  return passOnSuccess(
+    mockapi.get(`/special-words/operation/${name}`, (req) => {
```
weidongxu-microsoft, line 10: @timotheeguerin 

the path here does not match path in tsp (it is `operations`)

similar for other ones.

PR #445, opened 2023-11-01
Resource operation with request and response type

chunyu3:

### DIFF
--- a/packages/cadl-ranch-specs/cadl-ranch-summary.md
+++ b/packages/cadl-ranch-specs/cadl-ranch-summary.md
@@ -190,6 +190,30 @@ Expected response body:
 }
 ```
 
+### Azure_Core_Basic_createUser
+
+- Endpoint: `get /azure/core/basic`
+
+Should only generate CreateUserRequest Model as input and CreateUserResponse as response body.
+
+Expected input body:
+
+```json
+{
+  "name": "Madge",
+  "nickname": "nickname"
+}
+```
+
+Expected response body:
+
+````json
+{
+  "id": 1,
+  "name": "Madge",
+  "nickname": "nickname"
+}
+
 ### Azure_Core_Basic_delete
 
 - Endpoint: `get /azure/core/basic`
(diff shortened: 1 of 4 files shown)


packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
```diff
@@ -354,3 +354,39 @@ Expected response body:
...
+```json
+{
+  "id": 1,
+  "name": "Madge",
+  "nickname": "nickname"
+}
+""")
+@createsResource(User)
+@action("create")
+op createUser is global.Azure.Core.Foundations.ResourceCollectionOperation<User, CreateUserRequest, CreateUserResponse>;
```
weidongxu-microsoft, line 392: service does not suppose to use operation template from `Azure.Core.Foundations`?

weidongxu-microsoft:
I won't block this, but I also won't approve this.

https://github.com/Han-msft/azure-rest-api-specs-pr/pull/1/files#r1378296263

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -190,6 +190,30 @@ Expected response body:
 }
 ```
 
+### Azure_Core_Basic_createUser
```
timotheeguerin, line 193: what is the goal of this test? what does `createUser` mean as a scenario

timotheeguerin (changes requested):

packages/cadl-ranch-specs/cadl-ranch-summary.md
chunyu3: hello @timotheeguerin This test is to cover the senario: the request body and the response body is different from the resource type.

timotheeguerin: So what specific things from the azure core library does this scenario test. Seems to me it’s basically just adding wrapper on top of what could just be a pure typespec operation

PR #448, opened 2023-11-03
add test case for recursive inheritance scenario

msyyc:
with reference to https://github.com/Azure/azure-rest-api-specs-pr/pull/13119/files#diff-7b8b0c147d451c2784b2ea195d6d80753ede7d2e77579d16a1777c3a9f86ddff


weidongxu-microsoft (commented):
Is this about recursive reference, or about inheritance?

msyyc:
> Is this about recursive reference, or about inheritance?

The scenario contains inheritance and nested property but it is legal since it is not circular inheritance. Typespec-Python met an issue in this scenario so I want to add it in cadl-ranch

weidongxu-microsoft:
Any one have second opinion?

Personally I'd prefer having it `inheritance/recursive-reference`

msyyc:
> Any one have second opinion?
> 
> Personally I'd prefer having it `inheritance/recursive-reference` (or just `inheritance/recursive`?)

`inheritance/recursive-reference` is better. @tadelesh @chunyu3 @pshao25 @qiaozha Any thoughts?

iscai-msft:
> Any one have second opinion?
> 
> Personally I'd prefer having it `inheritance/recursive-reference` (or just `inheritance/recursive`?)

I like `inheritance/recursive`, I think that's all that's needed to describe the scenario

weidongxu-microsoft (approved):

msyyc:
> > Any one have second opinion?
> > Personally I'd prefer having it `inheritance/recursive-reference` (or just `inheritance/recursive`?)
> 
> I like `inheritance/recursive`, I think that's all that's needed to describe the scenario

Updated

chunyu3:
I think it is not a inheritance circle( model A extend B, B extend A), it just reference/property circle. It may not related to the inheritance. In that case, we may put it in `type/model/recursive`.

msyyc:
> I think it is not a inheritance circle( model A extend B, B extend A), it just reference/property circle. It may not related to the inheritance. In that case, we may put it in `type/model/recursive`. And i think if it is want to test property type recursive, it may not have difference with the circle e.g. `A: {p1: string; p2: A}` or `A{p1: string; p2: B} B: {pb: A}`. If so, we need to remove inheritance in the test, so that the test case is more focus and concise and only test what we want and does not bring in any other.

Python could pass `A: {p1: string; p2: A}` but fail for `A{p1: string; p2: B} B extends A`. That's why I add this scenario. It involves inheritance and recursion so I think it is fine to put it here.

chunyu3:
> > I think it is not a inheritance circle( model A extend B, B extend A), it just reference/property circle. It may not related to the inheritance. In that case, we may put it in `type/model/recursive`. And i think if it is want to test property type recursive, it may not have difference with the circle e.g. `A: {p1: string; p2: A}` or `A{p1: string; p2: B} B: {pb: A}`. If so, we need to remove inheritance in the test, so that the test case is more focus and concise and only test what we want and does not bring in any other.
> 
> Python could pass `A: {p1: string; p2: A}` but fail for `A{p1: string; p2: B} B extends A`. That's why I add this scenario. It involves inheritance and recursion so I think it is fine to put it here.

Thanks @msyyc But I still think it belong to property recursive, and this property recursive + inheritance is a special case of recursive in my opinions.  Anyway, It is ok if you want to put there. And  I suggest to add a normal property recursive test case.

chunyu3 (approved):

PR #455, opened 2023-11-23
fix for encode/bytes test case

msyyc:

### DIFF
--- a/packages/cadl-ranch-specs/http/encode/bytes/mockapi.ts
+++ b/packages/cadl-ranch-specs/http/encode/bytes/mockapi.ts
@@ -41,20 +41,11 @@ function createHeaderMockApis(route: string, value: any): MockApi {
   });
 }
 
-function createRequestBodyMockApis(
-  route: string,
-  value: any,
-  contentType: string = "application/json",
-  isBinary: boolean = false,
-): MockApi {
+function createRequestBodyMockApis(route: string, value: any, contentType: string = "application/json"): MockApi {
   const url = `/encode/bytes/body/request/${route}`;
   return mockapi.post(url, (req) => {
     req.expect.containsHeader("content-type", contentType);
-    if (isBinary) {
-      req.expect.bodyEquals(value);
-    } else {
-      req.expect.rawBodyEquals(value);
-    }
+    req.expect.rawBodyEquals(value);
     return {
       status: 204,
     };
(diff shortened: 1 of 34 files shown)


packages/cadl-ranch-specs/http/encode/bytes/mockapi.ts
```diff
@@ -41,11 +41,20 @@ function createHeaderMockApis(route: string, value: any): MockApi {
...
+  value: any,
+  contentType: string = "application/json",
+  isBinary: boolean = false,
+): MockApi {
   const url = `/encode/bytes/body/request/${route}`;
   return mockapi.post(url, (req) => {
     req.expect.containsHeader("content-type", contentType);
-    req.expect.rawBodyEquals(value);
+    if (isBinary) {
+      req.expect.bodyEquals(value);
```
weidongxu-microsoft, line 54: What's the difference of `bodyEquals` that makes the binary compare works?

weidongxu-microsoft (approved):

packages/cadl-ranch-specs/http/encode/bytes/mockapi.ts
msyyc: `req.expect.rawBodyEquals` compares request.rawBody which may not exist in this scenario while `req.expect.bodyEquals` compares request.body

pshao25: It's better to understand why request.rawBody is empty. I think it is not expected.

msyyc: I'm OK if you want to find the root cause. But I think it doesn't block the PR merge since it won't influence typespec-emitter.

pshao25: I don't know the root cause. My concern is the fix is too superficial. @timotheeguerin do you know why the  request.rawBody is empty while request.body actually has value.

timotheeguerin: Yeah I agree with @pshao25 this is a bit hacky here, I understand why it is working with body but I think the problem is quite simple, in server.ts you are missing the `verify: rawBodySaver` in the `image/png` parser you added

timotheeguerin: also as there is a lot of binary content types maybe its easier if we fix this one 
```
    this.app.use(bodyParser.raw({ type: "application/octet-stream", limit: "10mb" }));
```
aand change `type` to be an array of all the binary content types

timotheeguerin (approved):

PR #463, opened 2023-12-05
Add tests for api-version

pshao25:
There are two testing points
1. The version information in a service should be in the `@versioned`, anything in the `@service` should not be taken as versions.
2. Whether there is version appended in the uri for each operation depends on whether there is an `api-version` parameter in this operation. Version information just decides the default value of this `api-version` parameter, not decides the existence of this `api-version` parameter.

According to these, this PR divides the scenarios into 1) versioned service and 2) not versioned service. For each of these services there are three sub scenarios 1) apiversion not declared in the operation, 2) apiversion declared as query parameter in the operation, and 3) apiversion declared as path parameter in the operation.


packages/cadl-ranch-specs/http/server/versions/non-operations/main.tsp
```diff
@@ -0,0 +1,22 @@
...
+    endpoint: url,
+  }
+)
+namespace Server.Versions.NonOperations;
+
+@scenario
+@scenarioDoc("A simple operation without api-version. Expected url: '/myOp', it should not contain any api-version.")
+@route("/myOp")
+@head
+op myOp(): OkResponse;
```
msyyc, line 22: same with up comment

packages/cadl-ranch-specs/http/server/versions/all-operations/main.tsp
```diff
@@ -0,0 +1,37 @@
...
+  @doc("The version 2022-12-01-preview.")
+  @useDependency(global.Azure.Core.Versions.v1_0_Preview_2)
+  v2022_12_01_preview: "2022-12-01-preview",
+}
+
+@scenario
+@scenarioDoc("A simple operation with api-version. Expected url: '/myOp?api-version=2022-12-01-preview'.")
+@route("/myOp")
+@head
+op myOp(@query apiVersion: string): OkResponse;
```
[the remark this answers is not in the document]
pshao25: I changed to `withApiVersion`.

packages/cadl-ranch-specs/http/server/versions/non-operations/main.tsp
pshao25: I changed to `withoutApiVersion`.

packages/cadl-ranch-specs/http/server/versions/all-operations/main.tsp
```diff
@@ -0,0 +1,37 @@
+import "@azure-tools/typespec-azure-core";
+import "@typespec/http";
+import "@azure-tools/cadl-ranch-expect";
+import "@typespec/versioning";
+
+using TypeSpec.Http;
+using TypeSpec.Versioning;
+
+@doc("Illustrates server with all operations defined with api-versions.")
```
timotheeguerin, line 9: Can we start using doc comment instead of `@doc` 
See this PR for example https://github.com/Azure/cadl-ranch/pull/429/files#diff-0c3650d9ce1977c4f60dc4f5231c2b6be51d87d1c28b15548d74623ce89581e7

packages/cadl-ranch-specs/http/server/versions/non-operations/main.tsp
```diff
@@ -0,0 +1,22 @@
...
+  "{endpoint}",
+  "Testserver endpoint",
+  {
+    @doc("Need to be set as 'http://localhost:3000' in client.")
+    endpoint: url,
+  }
+)
+namespace Server.Versions.NonOperations;
+
+@scenario
```
timotheeguerin, line 18: For the not versioned and every operation versioned as the scenario is basically the whole service can you move the `@secenario` and `@Sscenario Doc` decorator directly on the namespace instead of the operation. That way it won't have this extra nesting in the dashboard

packages/cadl-ranch-specs/http/server/versions/all-operations/main.tsp
```diff
@@ -0,0 +1,37 @@
...
+)
+@server(
+  "{endpoint}",
+  "Testserver endpoint",
+  {
+    @doc("Need to be set as 'http://localhost:3000' in client.")
+    endpoint: url,
+  }
+)
+namespace Server.Versions.AllOperations;
```
timotheeguerin, line 24:
```suggestion
namespace Server.Versions.FullyVersioned;
```

packages/cadl-ranch-specs/http/server/versions/versioned/mockapi.ts
```diff
@@ -0,0 +1,37 @@
+import { passOnSuccess, mockapi, ValidationError } from "@azure-tools/cadl-ranch-api";
+import { ScenarioMockApi } from "@azure-tools/cadl-ranch-api";
+
+export const Scenarios: Record<string, ScenarioMockApi> = {};
+
+Scenarios.Server_Versions_Versioned_withoutApiVersion = passOnSuccess(
+  mockapi.head("/server/versions/versioned/withoutApiVersion", (req) => {
+    if (Object.keys(req.originalRequest.query).length > 0) {
```
msyyc, line 8: Why don't use `req.query` directly?

timotheeguerin (approved):

packages/cadl-ranch-specs/http/server/versions/not-versioned/main.tsp
```diff
@@ -0,0 +1,38 @@
+import "@typespec/http";
+import "@azure-tools/cadl-ranch-expect";
+
+using TypeSpec.Http;
+
+/**
+ * Illustrates not-versioned server.
+ */
+@scenarioService("/server/versions/notVersioned")
```
weidongxu-microsoft, line 9: nit, kebab-casing
https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#http-url-casing

same for `@route`

weidongxu-microsoft, line 9: `scenarioService` would sets a 1.0.0 to service version
https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-expect/src/decorators.ts#L228-L231

I am not sure whether emitter would/should take this as api-version.

We currently seems having no way in typespec to tell us whether this is Azure or non-Azure.

packages/cadl-ranch-specs/http/server/versions/versioned/main.tsp
```diff
@@ -0,0 +1,57 @@
...
+@scenarioDoc("A simple operation with query api-version, whose default value is defined as '2022-12-01-preview'. Expected url: '/withQueryApiVersion?api-version=2022-12-01-preview'.")
+@route("/withQueryApiVersion")
+@head
+op withQueryApiVersion(@query apiVersion: string): OkResponse;
+
+@scenario
+@scenarioDoc("A simple operation with path api-version, whose default value is defined as '2022-12-01-preview'. Expected url: '/withPathApiVersion/2022-12-01-preview'.")
+@route("/withPathApiVersion")
+@head
+op withPathApiVersion(@path apiVersion: string): OkResponse;
```
weidongxu-microsoft, line 57: api-version in path is usually specified via `@server`
https://github.com/Azure/cadl-ranch/blob/08312adaeaf041bd6446824476927a85a3e4d0a9/packages/cadl-ranch-specs/http/server/path/multiple/main.tsp#L13-L23

Unsure whether emitter supports this pattern directly in operation.

pshao25: Yes, this is another case I haven't got chance to add. I want these two query scenarios go first because our customer is waiting. I'll add path scenarios later.

packages/cadl-ranch-specs/http/server/versions/not-versioned/main.tsp
pshao25: This link you provided is setting service version, not related to versioned. So we are still not versioned.

> I am not sure whether emitter would/should take this as api-version.

Definitely it should not, and that is exactly what this test is testing, i.e., anything in `@service` should not be taken as version.

weidongxu-microsoft: I do not object to it. But please do have other languages give their opinion about whether they are on same page on this `@server(version=..).

I agree that non-Azure should definitely not take this. I am not exactly sure whether Azure should take this or not (but inclined to say not).

pshao25: I talked with @timotheeguerin long time ago about this. Everything in `@server(version=..)` is just metadata and that's it. All the versions should go in `@versioned`. Now @m-nash agrees with it too. Further, it applies to both azure and non-azure.

> Michael's position is that we should do at little as possible on this if (azure) {} else {}

A flag doesn't mean we will have such code. In .net we will have default non-azure behavior, and if the flag set we just override the default code to generate azure code (if the flag says azure). But how other languages do this I've no ideas.

packages/cadl-ranch-specs/http/server/versions/not-versioned/main.tsp
```diff
@@ -0,0 +1,38 @@
...
+namespace Server.Versions.NotVersioned;
+
+@scenario
+@scenarioDoc("A simple operation without api-version. Expected url: '/without-api-version', it should not contain any api-version.")
+@route("/without-api-version")
+@head
+op withoutApiVersion(): OkResponse;
+
+@scenario
+@scenarioDoc("A simple operation with query api-version, which doesn't have any default value. Expected url: '/with-query-api-version?api-version=v1.0.")
```
tadelesh, line 29:
```suggestion
@scenarioDoc("A simple operation with query api-version, which doesn't have any default value. Expected url: '/with-query-api-version?api-version=v1.0'.")
```

packages/cadl-ranch-specs/http/server/versions/not-versioned/main.tsp
```diff
@@ -0,0 +1,38 @@
+import "@typespec/http";
+import "@azure-tools/cadl-ranch-expect";
+
+using TypeSpec.Http;
+
+/**
+ * Illustrates not-versioned server.
```
tadelesh, line 7: What is the scenario for these two cases? A service that accept api version param, but not defined the potential value? How should emitter deal with the api version param?

packages/cadl-ranch-specs/http/server/versions/versioned/main.tsp
```diff
@@ -0,0 +1,57 @@
...
+@scenarioDoc("A simple operation without api-version. Expected url: '/without-api-version', it should not contain any api-version.")
+@route("/without-api-version")
+@head
+op withoutApiVersion(): OkResponse;
+
+@scenario
+@scenarioDoc("A simple operation with query api-version, whose default value is defined as '2022-12-01-preview'. Expected url: '/with-query-api-version?api-version=2022-12-01-preview'.")
+@route("/with-query-api-version")
+@head
+op withQueryApiVersion(@query apiVersion: string): OkResponse;
```
tadelesh, line 51: Why this api version param not use the enum type?

packages/cadl-ranch-specs/http/server/versions/not-versioned/main.tsp
pshao25: See the description.

packages/cadl-ranch-specs/http/server/versions/versioned/main.tsp
pshao25: You mean `op withQueryApiVersion(@query apiVersion: SomeEnum): OkResponse;`? I'm not testing the type of apiversion. I'm testing whether the apiversion should be in the url.

tadelesh (approved):

PR #465, opened 2023-12-06
Add more cases for default client hierarchy

tadelesh:


packages/cadl-ranch-specs/http/client/structure/default/client.tsp
```diff
@@ -6,7 +6,7 @@ import "@azure-tools/typespec-client-generator-core";
   """
   This is to show that if we don't do any customization. The client side should be able to call the api like
   ```ts
-  const client = new MultiClient("default");
+  const client = new ServiceClient("default");
```
qiaozha, line 9: should we update the call for seven, eight and nine ? I might get a little confused, what's the default behavior for hierarchy client is ?  do we treat namespaces as hierarchies or not ?

qiaozha: if yes, should it be client.foo.three() and client.baz.foo.seven() ?

tadelesh: yes. it should be `client.baz.foo.seven()`

weidongxu-microsoft (approved):

qiaozha (approved):

PR #469, opened 2023-12-07
disable api-version check in lro polling for now

msyyc:
It still needs some time for Python to fix api-version check logic for lro polling, however, there are already some issues introduced into python emitter because of lack of lro test case. I hope to disable the api-version check logic for now since it won't influence other language. Once python fixed, I will return the logic back.


weidongxu-microsoft (commented):
OK for Java, for a temporary solution.

Please get approve from all affected SDKs.

PR #473, opened 2023-12-14
Add tests for payload under different content types and body types

pshao25:
1. Add tests for payload under different content types and body types
2. Fix a bug in packages\cadl-ranch-specs\http\payload\content-negotiation

### DIFF
--- a/packages/cadl-ranch-specs/cadl-ranch-summary.md
+++ b/packages/cadl-ranch-specs/cadl-ranch-summary.md
@@ -1581,10 +1581,10 @@ Scenario that returns a different file encoding depending on the accept header.
 ### Payload_MediaType_StringBody
 
 - Endpoints:
-  - `post /payload/media-type/sendStringAsText`
-  - `post /payload/media-type/getStringAsText`
-  - `post /payload/media-type/sendStringAsJson`
-  - `post /payload/media-type/getStringAsJson`
+  - `post /payload/media-type/sendAsText`
+  - `post /payload/media-type/getAsText`
+  - `post /payload/media-type/sendAsJson`
+  - `post /payload/media-type/getAsJson`
 
 When the request body and response body are both strings, the different media types would lead to different payloads.
 
--- a/packages/cadl-ranch-specs/cadl-ranch-summary.md
+++ b/packages/cadl-ranch-specs/cadl-ranch-summary.md
@@ -1578,15 +1578,29 @@ Scenario that returns a different file encoding depending on the accept header.
 - image/png return a png image
 - image/jpeg return a jpeg image
 
-### Payload_MediaType_StringBody
+### Payload_MediaType_StringBody_getAsJson
 
-- Endpoints:
-  - `post /payload/media-type/stringBody/sendAsText`
-  - `post /payload/media-type/stringBody/getAsText`
-  - `post /payload/media-type/stringBody/sendAsJson`
-  - `post /payload/media-type/stringBody/getAsJson`
+- Endpoint: `get /payload/media-type/string-body/getAsJson`
 
-When the request body and response body are both strings, the different media types would lead to different payloads.
+Expected response body is "foo".
+
+### Payload_MediaType_StringBody_getAsText
+
+- Endpoint: `get /payload/media-type/string-body/getAsText`
+
+Expected response body is a string '{cat}'.
+
+### Payload_MediaType_StringBody_sendAsJson
+
+- Endpoint: `post /payload/media-type/string-body/sendAsJson`
+
+Expected request body is "foo".
+
+### Payload_MediaType_StringBody_sendAsText
+
+- Endpoint: `post /payload/media-type/string-body/sendAsText`
+
+Expected request body is a string '{cat}'.
 
 ### Payload_MultiPart_FormData_basic
 
--- a/packages/cadl-ranch-specs/cadl-ranch-summary.md
+++ b/packages/cadl-ranch-specs/cadl-ranch-summary.md
@@ -1581,10 +1581,10 @@ Scenario that returns a different file encoding depending on the accept header.
 ### Payload_MediaType_StringBody
 
 - Endpoints:
-  - `post /payload/media-type/sendAsText`
-  - `post /payload/media-type/getAsText`
-  - `post /payload/media-type/sendAsJson`
-  - `post /payload/media-type/getAsJson`
+  - `post /payload/media-type/stringBody/sendAsText`
+  - `post /payload/media-type/stringBody/getAsText`
+  - `post /payload/media-type/stringBody/sendAsJson`
+  - `post /payload/media-type/stringBody/getAsJson`
 
 When the request body and response body are both strings, the different media types would lead to different payloads.
 
--- a/packages/cadl-ranch-specs/cadl-ranch-summary.md
+++ b/packages/cadl-ranch-specs/cadl-ranch-summary.md
@@ -1578,6 +1578,16 @@ Scenario that returns a different file encoding depending on the accept header.
 - image/png return a png image
 - image/jpeg return a jpeg image
 
+### Payload_MediaType_StringBody
+
+- Endpoints:
+  - `post /payload/media-type/sendStringAsText`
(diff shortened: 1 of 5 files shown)


packages/cadl-ranch-specs/http/payload/content-negotiation/main.tsp
```diff
@@ -53,7 +53,7 @@ namespace DifferentBody {
   }
 
   model PngImageAsJson {
-    @header contentType: "image/jpeg";
+    @header contentType: "application/json";
```
iscai-msft, line 56: @timotheeguerin to confirm that this fix is correct

packages/cadl-ranch-specs/http/payload/media-type/main.tsp
```diff
@@ -0,0 +1,56 @@
...
+using Azure.ClientGenerator.Core;
+
+/**
+ * Test the payload with different media types and different types of the payload itself.
+ */
+@supportedBy("dpg")
+@scenarioService("/payload/media-type")
+namespace Payload.MediaType;
+
+// TODO: this file currently only contains string as body, we will add bytes body, model body soon.
```
iscai-msft, line 15: nit: can you make an issue for this instead of publishing with a `TODO`

packages/cadl-ranch-specs/http/payload/media-type/main.tsp
```diff
@@ -0,0 +1,56 @@
...
+  When the request body and response body are both strings, the different media types would lead to different payloads.
+  """)
+@operationGroup
+namespace StringBody {
+  /**
+   * Expected request body is a string '{"Cat": "Meow"}'.
+   */
+  @post
+  @route("/sendStringAsText")
+  op sendStringAsText(@header contentType: "text/plain", @body text: string): OkResponse;
```
iscai-msft, line 28: the name can just be `sendAsText` because `StringBody` will already be part of the title

packages/cadl-ranch-specs/http/payload/media-type/main.tsp
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Test the payload with different media types and different types of the payload itself.
+ */
+@supportedBy("dpg")
+@scenarioService("/payload/media-type")
+namespace Payload.MediaType;
+
+@scenario
+@scenarioDoc("""
+  When the request body and response body are both strings, the different media types would lead to different payloads.
```
tadelesh, line 17: What do you mean different media types lead to different payloads? It seems all the payload is just string, the difference is only the content type.

packages/cadl-ranch-specs/http/payload/media-type/main.tsp
```diff
@@ -0,0 +1,55 @@
...
+  op sendAsJson(@header contentType: "application/json", @body text: string): OkResponse;
+
+  /**
+   * Expected response body is "foo".
+   */
+  @get
+  @route("/getAsJson")
+  op getAsJson(): {
+    @header contentType: "application/json";
+    @body text: string;
```
haolingdong-msft, line 53: Sorry I don't quite get the intention of the test case. `send` and `get` as Json return string-like object, `send` and `get` as string return object-like string?

packages/cadl-ranch-specs/http/payload/media-type/main.tsp
pshao25: You could see details from this customer request https://github.com/microsoft/typespec/discussions/2557

You could think the media type as how we encode/decode the content. For example, when the media type is "text/plain", we just keep what the data is, not do anything like adding quotes or deleting quotes as what "application/json" does.

packages/cadl-ranch-specs/http/payload/media-type/main.tsp
pshao25: Yes, this is customer request https://github.com/microsoft/typespec/discussions/2557

packages/cadl-ranch-specs/http/payload/media-type/main.tsp
tadelesh: I see. Seems content negotiation is for response, this is for request.

tadelesh (approved):

packages/cadl-ranch-specs/http/payload/media-type/main.tsp
pshao25: Yes, customer request is for response, but here in tests I want to make sure we test request as well.

packages/cadl-ranch-specs/http/payload/content-negotiation/main.tsp
msyyc: The fix shall be right since accept of the op is `application/json`: https://github.com/Azure/cadl-ranch/blob/6fafb9a5bdced283985dfa8ed960791efb84d5fd/packages/cadl-ranch-specs/http/payload/content-negotiation/main.tsp#L64

packages/cadl-ranch-specs/http/payload/media-type/main.tsp
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * Expected request body is a string '{"Cat": "Meow"}'.
+   */
+  @post
+  @route("/sendAsText")
+  op sendAsText(@header contentType: "text/plain", @body text: string): OkResponse;
+
+  /**
+   * Expected response body is a string '{"Cat": "Meow"}'.
+   */
```
weidongxu-microsoft, line 31: Use a text that would be invalid if interpretated as JSON, e.g. `cat`.

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -1578,6 +1578,16 @@ Scenario that returns a different file encoding depending on the accept header.
...
 
+### Payload_MediaType_StringBody
+
+- Endpoints:
+  - `post /payload/media-type/stringBody/sendAsText`
+  - `post /payload/media-type/stringBody/getAsText`
+  - `post /payload/media-type/stringBody/sendAsJson`
+  - `post /payload/media-type/stringBody/getAsJson`
+
+When the request body and response body are both strings, the different media types would lead to different payloads.
```
weidongxu-microsoft, line 1589: Personally prefer to use "content type" in all places.

pshao25: I use "media type" because we not only contain "content type" but also "accept".

packages/cadl-ranch-specs/http/payload/media-type/main.tsp
```diff
@@ -0,0 +1,55 @@
...
+@scenario
+@scenarioDoc("""
+  When the request body and response body are both strings, the different media types would lead to different payloads.
+  """)
+@operationGroup
+@route("/stringBody")
+namespace StringBody {
+  /**
+   * Expected request body is a string '{cat}'.
+   */
```
weidongxu-microsoft, line 24: Should you do `@scenario` and `@scenarioDoc` on each API?

I guess the comment here won't show up in dashboard. And it be hard for dev to figure out what is the expected request/response.

packages/cadl-ranch-specs/http/payload/media-type/main.tsp
```diff
@@ -0,0 +1,55 @@
...
+@supportedBy("dpg")
+@scenarioService("/payload/media-type")
+namespace Payload.MediaType;
+
+@scenario
+@scenarioDoc("""
+  When the request body and response body are both strings, the different media types would lead to different payloads.
+  """)
+@operationGroup
+@route("/stringBody")
```
weidongxu-microsoft, line 20: nit, `/string-body`, or `/body-string`?

packages/cadl-ranch-specs/http/payload/content-negotiation/main.tsp
timotheeguerin: Yeah  that looks correct

timotheeguerin (approved):

weidongxu-microsoft (approved):
Make sure you've tested the case via .NET client or curl.

PR #484, opened 2023-12-28
add non-string type for multipart/form-data basic/complex scenario

msyyc:
With reference of openai: https://github.com/bterlson/openai-in-typespec/blob/aa3891ccb29a1c8377b6ef3e8f5b5f4825630979/audio/operations.tsp#L19

### DIFF
--- a/packages/cadl-ranch-specs/cadl-ranch-summary.md
+++ b/packages/cadl-ranch-specs/cadl-ranch-summary.md
@@ -285,6 +285,34 @@ Expected response body:
 }
 ```
 
+### Azure_Core_Basic_exportAllUsers
+
+- Endpoint: `post /azure/core/basic`
+
+Should generate a model named User.
+
+Expected query parameter: format=json
+Expected query parameter: api-version=2022-12-01-preview
+
+Expected response body:
+
+```json
+{
+  "users": [
+    {
+      "id": 1,
+      "name": "Madge",
+      "etag": "11bdc430-65e8-45ad-81d9-8ffa60d55b59"
+    },
+    {
+      "id": 2,
+      "name": "John",
+      "etag": "22bdc430-65e8-45ad-81d9-8ffa60d55b59"
+    }
+  ]
+}
+```
+
 ### Azure_Core_Basic_get
 
 - Endpoint: `get /azure/core/basic`
--- a/packages/cadl-ranch-specs/cadl-ranch-summary.md
+++ b/packages/cadl-ranch-specs/cadl-ranch-summary.md
@@ -1056,6 +1056,152 @@ Expected response body:
 }
 ```
 
+### Azure_ResourceManager_Models_Resources_SingletonTrackedResources_createOrUpdate
+
+- Endpoint: `put https://management.azure.com`
+
+Resource PUT operation.
+Expected path: /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/test-rg/providers/Azure.ResourceManager.Models.Resources/singletonTrackedResources/default
+Expected query parameter: api-version=2023-12-01-preview
+Expected request body:
+
+```json
+{
+  "location": "eastus",
+  "properties": {
+    "description": "valid"
+  }
+}
+```
+
+Expected response body:
+
+```json
+{
+  "id": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/test-rg/providers/Azure.ResourceManager.Models.Resources/singletonTrackedResources/default",
+  "name": "default",
+  "type": "Azure.ResourceManager.Models.Resources/singletonTrackedResources",
+  "location": "eastus",
+  "properties": {
+    "description": "valid",
+    "provisioningState": "Succeeded"
+  },
+  "systemData": {
+    "createdBy": "AzureSDK",
+    "createdByType": "User",
+    "createdAt": <any date>,
+    "lastModifiedBy": "AzureSDK",
+    "lastModifiedAt": <any date>,
+    "lastModifiedByType": "User",
+  }
+}
+```
+
+### Azure_ResourceManager_Models_Resources_SingletonTrackedResources_getByResourceGroup
+
+- Endpoint: `get https://management.azure.com`
+
+Resource GET operation.
+Expected path: /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/test-rg/providers/Azure.ResourceManager.Models.Resources/singletonTrackedResources/default
+Expected query parameter: api-version=2023-12-01-preview
+
+Expected response body:
+
+```json
+{
+  "id": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/test-rg/providers/Azure.ResourceManager.Models.Resources/singletonTrackedResources/default",
+  "name": "default",
+  "type": "Azure.ResourceManager.Models.Resources/singletonTrackedResources",
+  "location": "eastus",
+  "properties":{
+    "description": "valid",
+    "provisioningState": "Succeeded"
+  },
+  "systemData": {
+    "createdBy": "AzureSDK",
+    "createdByType": "User",
+    "createdAt": <any date>,
+    "lastModifiedBy": "AzureSDK",
+    "lastModifiedAt": <any date>,
+    "lastModifiedByType": "User",
+  }
+}
+```
+
+### Azure_ResourceManager_Models_Resources_SingletonTrackedResources_listByResourceGroup
+
+- Endpoint: `get https://management.azure.com`
+
+Resource LIST by resource group operation.
+Expected path: /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/test-rg/providers/Azure.ResourceManager.Models.Resources/singletonTrackedResources
+Expected query parameter: api-version=2023-12-01-preview
+
+Expected response body:
+
+```json
+{
+  "value": [{
+    "id": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/test-rg/providers/Azure.ResourceManager.Models.Resources/singletonTrackedResources/default",
+    "name": "default",
+    "type": "Azure.ResourceManager.Models.Resources/singletonTrackedResources",
+    "location": "eastus",
+    "properties":{
+      "description": "valid",
+      "provisioningState": "Succeeded"
+    },
+    "systemData": {
+      "createdBy": "AzureSDK",
+      "createdByType": "User",
+      "createdAt": <any date>,
+      "lastModifiedBy": "AzureSDK",
+      "lastModifiedAt": <any date>,
+      "lastModifiedByType": "User",
+    }
+  }]
+}
+```
+
+### Azure_ResourceManager_Models_Resources_SingletonTrackedResources_update
+
+- Endpoint: `patch https://management.azure.com`
+
+Resource PATCH operation.
+Expected path: /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/test-rg/providers/Azure.ResourceManager.Models.Resources/singletonTrackedResources/default
+Expected query parameter: api-version=2023-12-01-preview
+Expected request body:
+
+```json
+{
+  "location": "eastus2",
+  "properties": {
+    "description": "valid2"
+  }
+}
+```
+
+Expected response body:
+
+```json
+{
+  "id": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/test-rg/providers/Azure.ResourceManager.Models.Resources/singletonTrackedResources/default",
+  "name": "default",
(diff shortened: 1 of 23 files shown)


packages/cadl-ranch-specs/http/payload/multipart/main.tsp
```diff
@@ -9,6 +9,7 @@ namespace Payload.MultiPart;
 
 model MultiPartRequest {
   id: string;
+  temperature: float64;
```
MaryGao, line 12: Do we have a real case here? Do we need to consider other primitive types e.g null, boolean?

packages/cadl-ranch-specs/http/payload/multipart/main.tsp
```diff
@@ -65,6 +65,11 @@ Content-Type: text/plain
 
 123
 --abcde12345
+Content-Disposition: form-data; name="temperature"
+Content-Type: text/plain
```
MaryGao, line 69: Does that mean following settings?
- for primitive types the default content type is `text/plain`
- for objects(object, array, record etc) the default content type is `application/json`

I am asking because for `string` if it is json content, it would be `"raw string"`; if it is text content, then would be `raw string`.

packages/cadl-ranch-specs/http/payload/multipart/main.tsp
msyyc: Yeah, openai: https://github.com/bterlson/openai-in-typespec/blob/aa3891ccb29a1c8377b6ef3e8f5b5f4825630979/audio/operations.tsp#L19

packages/cadl-ranch-specs/http/payload/multipart/main.tsp
msyyc: I understand what you mean. cadl-ranch uses `multer` to parse and it has limitation: different with `files`, `data` part lose content-type info so please pass it as plain string instead of json string. 
image
 In the future, maybe I can find another package to get the info then the case can handle it with different content types.

weidongxu-microsoft: I think it still `text/plain`. Admit this be the same problem of ambiguous on content-type.

AFAIK, for current openai impl, the string and float would result to a text without quote.
```java
        if (audioTranscriptionOptions.getPrompt() != null) {
            fields.add(new MultipartField("prompt",
                audioTranscriptionOptions.getPrompt()));
        }
        if (audioTranscriptionOptions.getTemperature() != null) {
            fields.add(new MultipartField("temperature",
                String.valueOf(audioTranscriptionOptions.getTemperature())));
        }
```
(float temperature is converted to string)

weidongxu-microsoft (approved):

packages/cadl-ranch-specs/http/payload/multipart/main.tsp
msyyc: @MaryGao, @weidongxu-microsoft I update this case with `HttpPart` to declare contentType with `text/plain` so there shall be no ambiguity now.

weidongxu-microsoft (approved):
Not sure about float prevcision

msyyc:
> Not sure about float prevcision

Update to `float64`

packages/cadl-ranch-specs/http/payload/multipart/main.tsp
```diff
@@ -457,4 +457,33 @@ namespace FormData {
...
+    Content-Type: text/plain
+    
+    0.5
+    --abcde12345
+    ```
+    """)
+  @doc("Test content-type: multipart/form-data for non string")
+  @post
+  @route("/non-string")
+  op nonString(
```
iscai-msft, line 480: i wonder if it would be better to call it `float` instead of `nonString`, incase there are other non-string primitives we want to test. I'm ok either way though

msyyc: updated

packages/cadl-ranch-specs/http/payload/multipart/main.tsp
```diff
@@ -457,4 +457,33 @@ namespace FormData {
...
+    Content-Type: text/plain
+    
+    0.5
+    --abcde12345
+    ```
+    """)
+  @doc("Test content-type: multipart/form-data for non string")
+  @post
+  @route("/non-string-float")
+  op floatWithHttpPart(
```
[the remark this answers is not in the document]
msyyc: I am OK if other language developers agree with the breakings. @weidongxu-microsoft @MaryGao @chunyu3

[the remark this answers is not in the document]
weidongxu-microsoft: I am OK if you are to move HttpPart test to a operation group or a client.

If possible please keep non-HttpPart same.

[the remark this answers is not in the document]

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -2954,7 +2954,7 @@ Content-Type: application/octet-stream
 --abcde12345--
 ```
 
-### Payload_MultiPart_FormData_complexWithHttpPart
+### Payload_MultiPart_FormData_HttpPartTest_complexWithHttpPart
```
iscai-msft, line 2957: nit: `Payload_MultiPart_FormData_HttpParts_x`. Also `complex` isn't a great name: what makes it complex? It appears to be that includes an array of http parts, so I would call it `Payload_MultiPart_FormData_HttpParts_Array`

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -3005,7 +3005,7 @@ Content-Type: application/octet-stream
 --abcde12345--
 ```
 
-### Payload_MultiPart_FormData_fileWithHttpPartOptionalContentType
+### Payload_MultiPart_FormData_HttpPartTest_fileWithHttpPartOptionalContentType
```
iscai-msft, line 3008: `Payload_MultiPart_FormData_HttpParts_optionalContentType`

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -3062,6 +3062,25 @@ Content-Type: image/jpg
 --abcde12345--
 ```
 
+### Payload_MultiPart_FormData_HttpPartTest_NonString_floatMethod
```
iscai-msft, line 3065: `Payload_MultiPart_FormData_HttpParts_NonString_float`

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -3024,7 +3024,7 @@ Content-Type: application/octet-stream
 --abcde12345
 ```
 
-### Payload_MultiPart_FormData_fileWithHttpPartRequiredContentType
+### Payload_MultiPart_FormData_HttpPartTest_fileWithHttpPartRequiredContentType
```
iscai-msft, line 3027: `Payload_MultiPart_FormData_HttpParts_requiredContentType`

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -3043,7 +3043,7 @@ Content-Type: application/octet-stream
 --abcde12345--
 ```
 
-### Payload_MultiPart_FormData_fileWithHttpPartSpecificContentType
+### Payload_MultiPart_FormData_HttpPartTest_fileWithHttpPartSpecificContentType
```
iscai-msft, line 3046: I also wouldn't use "specific", it's image jpeg, so it can be `Payload_MultiPart_FormData_HttpParts_ImageJpegContentType`.

We can also have a content types section, which has sub-ops `optional`, `required`, `image_jpeg`. But I'm not sure if all languages can handle nesting that far

packages/cadl-ranch-specs/cadl-ranch-summary.md
msyyc: - `HttpPartTest` udpated to `HttpParts`
- "Complex" means combination of single scenarios. Array may cause confusion since it contains scenario for array of json and array of file.

packages/cadl-ranch-specs/cadl-ranch-summary.md
msyyc: I group these 3 content type cases like non-string.

packages/cadl-ranch-specs/cadl-ranch-summary.md
iscai-msft: We've purposefully not used the word `complex` in `cadl-ranch` though, and this is the only case of a test with `complex`. I would prefer to have it explicitly state `JsonArrayAndFileArray` than have `complex`, because we want to be able to immediately tell exactly what we're testing from the dashboard. 

Since this is testing multiple scenarios at once, it might also mean we should split this into two tests: `_Array_Json` and `_Array_File`, unless we're explicitly testing the interplay of these 2 concepts

msyyc: We already have cases to test array of json and array of file and `complex` case is to testing the interplay of these 2 concepts. Now that name `complex` is not recommended, I update it to `jsonArrayAndFileArrayAndOthers`

iscai-msft: where are the json array tests? I see a model defined called `JsonArrayPartsRequest` but no test for it. I'm ok with the name, my only nit is we don't need `andOthers`, but let me go ahead and approve this pr. thanks for your work @msyyc

iscai-msft (approved):

packages/cadl-ranch-specs/cadl-ranch-summary.md
msyyc: Thanks for the approval.
- FileArray is only tested in `ComplexHttpPartsModelRequest` since common model can't define the valid scenario for cadl-ranch to test and only `HttpPart` can
- Removed `AndOthers` from the method name.

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -3172,60 +3173,78 @@ Content-Type: application/octet-stream
...
-{…file content of .jpg file…}
+0.5
 --abcde12345
 ```
 
-### Payload_MultiPart_FormData_fileWithHttpPartRequiredContentType
+### Payload_MultiPart_FormData_jsonArrayAndFileArrayAndOthers
 
-- Endpoint: `post /multipart/form-data/check-filename-and-required-content-type-with-httppart`
+- Endpoint: `post /multipart/form-data/complex-parts`
```
MaryGao, line 3197: nit: change sth like: form-data/json-array-and-file-array-and-others

packages/cadl-ranch-specs/http/payload/multipart/main.tsp
```diff
@@ -330,131 +333,165 @@ namespace FormData {
...
+        ```
+        """)
+      @doc("Test content-type: multipart/form-data for non string")
+      @post
+      @route("/non-string-float")
+      op float(
+        @header contentType: "multipart/form-data",
+        @multipartBody body: {
+          temperature: HttpPart<{
+            @body body: float64;
```
MaryGao, line 490: just curious, do we have a RP for this?

MaryGao (approved):
Approved!

packages/cadl-ranch-specs/http/payload/multipart/main.tsp
```diff
@@ -164,7 +164,10 @@ namespace FormData {
   @doc("Test content-type: multipart/form-data for mixed scenarios")
   @post
   @route("/complex-parts")
-  op complex(@header contentType: "multipart/form-data", @body body: ComplexPartsRequest): NoContentResponse;
+  op jsonArrayAndFileArrayAndOthers(
+    @header contentType: "multipart/form-data",
+    @body body: ComplexPartsRequest,
```
chunyu3, line 169: There is no `jsonArray` in the model `ComplexPartsRequest`. Do you use the wrong model as body?

chunyu3, line 169: There is no `json array` in the model `ComplexPartsRequest`. Do you use the wrong model as body?
```suggestion
    @body body: ComplexHttpPartsModelRequest,
```

packages/cadl-ranch-specs/cadl-ranch-summary.md
msyyc: It doesn't matter so will keep it

packages/cadl-ranch-specs/http/payload/multipart/main.tsp
msyyc: https://github.com/Azure/cadl-ranch/pull/484#issue-2057926153

packages/cadl-ranch-specs/http/payload/multipart/main.tsp

PR #485, opened 2023-12-29
cadl-ranch-spec, add test case for json-merge-patch

haolingdong-msft:
Add test cases for json-merge-patch.
I'm not using Azure core template [ResourceCreateOrUpdate](https://github.com/Azure/typespec-azure/blob/86d926a3ce78328e9bab146f15f46873cb0e1461/packages/typespec-azure-core/lib/operations.tsp#L161) currently because I would like to make the case reusable for both azure and non-azure. Template underline set the content type to `merge-patch+json`, and it is transparent to emitters as we only get the content type regardless it's set by template or by user. Let me know if you feel it's necessary to add another test file under `Azure` folder.

Test serialization for json merge patch on below cases:
1. request body required and optional.
1. property is `Record`(map): this is the same case as loadtesting, loadtesting is using map property to do json merge patch.  https://github.com/Azure/azure-rest-api-specs/blob/main/specification/loadtestservice/LoadTestService/models.tsp#L418-L431
2. property is string: this is the same case as content safety.  https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/ContentSafety/models.tsp#L136-L149
3. property is inner model
4. property is array
5. property is int
6. property is float


packages/cadl-ranch-specs/http/payload/json-merge-patch/main.tsp
```diff
@@ -0,0 +1,219 @@
...
+  ],
+  "intValue": 1,
+  "floatValue": 1.1
+}
+```
+""")
+@doc("Test content-type: application/merge-patch+json with required body")
+@route("/create/resource")
+@put
+op createResource(@body body: Resource): Resource;
```
pshao25, line 76: Do you have any guidance saying a type could be in both put and patch?

packages/cadl-ranch-specs/http/payload/json-merge-patch/main.tsp
```diff
@@ -0,0 +1,219 @@
...
+import "@azure-tools/cadl-ranch-expect";
+
+using TypeSpec.Http;
+
+@doc("Test for merge-patch+json content-type")
+@scenarioService("/json-merge-patch")
+namespace Payload.JsonMergePatch;
+
+@doc("Details about a resource.")
+model Resource {
```
pshao25, line 11: https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#rest-response-body-is-resource-schema

The PATCH request schema should contain all the same fields with no required fields

haolingdong-msft: Yes. This is Azure guideline, I think in Azure world, we may recommend customer to use the core template.

packages/cadl-ranch-specs/http/payload/json-merge-patch/main.tsp
haolingdong-msft: Azure may use the core template. But this would be a common case in non-azure. First create the resource, then update. I did not find official guidelines. But there is a discussion in Java PR: https://github.com/Azure/autorest.java/pull/2424#discussion_r1432216605

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -1578,6 +1578,189 @@ Scenario that returns a different file encoding depending on the accept header.
...
+
+- Endpoint: `patch /json-merge-patch/update/resource/optional`
+
+Should serialize null values with merge-patch+json enabled.
+
+Expected input body:
+
+```json
+{
+  "name": "Madge",
```
weidongxu-microsoft, line 1691: For a patch, we'd expect no value for property with no change? So no `name` needed?

At least, your verification code should not verify this `name=Madge` in payload.

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -1578,6 +1578,189 @@ Scenario that returns a different file encoding depending on the accept header.
...
+  "description": null,
+  "map": {
+    "key": {
+      "name": "InnerMadge",
+      "description": null
+    }
+  },
+  "array": null,
+  "intValue": null,
+  "floatValue": null
```
weidongxu-microsoft, line 1719: Typical response would just having these optional property (of no value) not present.

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -1578,6 +1578,189 @@ Scenario that returns a different file encoding depending on the accept header.
...
+```
```
weidongxu-microsoft, line 1744: typo?

packages/cadl-ranch-specs/http/payload/json-merge-patch/main.tsp
pshao25: > I think in Azure world, we may recommend customer to use the core template [ResourceCreateOrUpdate](https://github.com/Azure/typespec-azure/blob/86d926a3ce78328e9bab146f15f46873cb0e1461/packages/typespec-azure-core/lib/operations.tsp#L161)

I don't think so. See https://github.com/Azure/typespec-azure-pr/issues/3688

> But in the core template, the json-merge-patch operation will have the same request and response model. -> this could be different to the guideline you pasted, so I did not add the case using Azure template before there is agreement on this.

What I pasted is restapi guideline.  No matter how you define your template, the payload should follow it. To be more specific, the raw operation you defined should follow it.

packages/cadl-ranch-specs/http/payload/json-merge-patch/main.tsp
pshao25: Yeah I think for Azure, we shouldn't have both put and patch with the same model as stated in https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#rest-response-body-is-resource-schema. But for non-azure, I'm not sure.

packages/cadl-ranch-specs/cadl-ranch-summary.md
haolingdong-msft: I changed response to `NoContentResponse`

packages/cadl-ranch-specs/http/payload/json-merge-patch/main.tsp
haolingdong-msft: `loadtesting` is already using required property to do patch now, `components: Record<AppComponent>;` -> `components` is required.

https://github.com/Azure/azure-rest-api-specs/blob/main/specification/loadtestservice/LoadTestService/models.tsp#L418-L431

packages/cadl-ranch-specs/cadl-ranch-summary.md
weidongxu-microsoft: `NoContentResponse` may not be a realistic response... I assume typical PATCH would return a resource.

Let's wait for opinion from other reviewers before further change.

packages/cadl-ranch-specs/http/payload/json-merge-patch/main.tsp
haolingdong-msft: If it's round trip model, I think both create(put) and update(patch) will use the same model, which means patch operation will use the model with optional property. This may violate the guideline @pshao25 pasted: [The PATCH request schema should contain all the same fields with no required fields]https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#rest-response-body-is-resource-schema

I can add `@visibility('create')` on required properties to avoid required properties to be used on patch operation. But I think we also have many unfinalized discussions on `@visibility` decorator, and `@visibility('create')` is not supported by our languages currently. Also now `loadtesting` is already using model with required property for json-merge-patch operation. (See above example). 

**So I'm wondering should json-merge-patch operation use a model that does not contain any required property?**

Add archs and typespec crew: @srnagar  @m-nash @markcowl

haolingdong-msft: To make json-merge-patch test case less ambiguous. I created another model named `ResourcePatch` which includes all the optional properties.

pshao25: They are using template because they don't have any other choices, not because it is correct usage.

And from https://learn.microsoft.com/en-us/rest/api/loadtesting/dataplane/load-test-administration/create-or-update-app-components?view=rest-loadtesting-dataplane-2022-11-01&tabs=HTTP, they have a `AppComponentUpdate` for patch operation. 

I think we still have a lot to discuss for required properties. I suggest we remove them for now until we have an aggreement.

packages/cadl-ranch-specs/cadl-ranch-summary.md
haolingdong-msft: Updated. Make response as `Resource` and remove the `null` fields in response now.

packages/cadl-ranch-specs/http/payload/json-merge-patch/main.tsp
```diff
@@ -0,0 +1,222 @@
...
+    }
+  ],
+  "intValue": 1,
+  "floatValue": 1.1
+}
+```
+""")
+@doc("Test content-type: application/merge-patch+json with required body")
+@route("/create/resource")
+@put
```
pshao25, line 84: What is the purpose of adding put operation?

packages/cadl-ranch-specs/http/payload/json-merge-patch/main.tsp
```diff
@@ -0,0 +1,222 @@
...
+  array?: InnerModel[];
+  intValue?: int32;
+  floatValue?: float32;
+}
+
+@doc("Details about a resource for patch operation.")
+model ResourcePatch {
+  description?: string;
+  map?: Record<InnerModel>;
+  array?: InnerModel[];
```
pshao25, line 24: It may be `InnerModelPatch`?

packages/cadl-ranch-specs/http/payload/json-merge-patch/main.tsp
haolingdong-msft: Simulate create operation, normally a resource needs to be created first, then update.

packages/cadl-ranch-specs/http/payload/json-merge-patch/main.tsp
pshao25: and seems it lacks a testing point that a property is a model?

weidongxu-microsoft: @pshao25 you are asking that every model that referred by a PATCH op have all properties optional.

In reality this probably not going to happen. It is EVERY models referred by the PATCH op.

In the cadl-ranch, I think Haoling can make the change to have the `name` there optional as well. But in reality, you will need a lint to prevent anyone doing this (provided that we all agree that required property should not appear).

pshao25: and I think we need an array of primitive type.

packages/cadl-ranch-specs/http/payload/json-merge-patch/mockapi.ts
```diff
@@ -0,0 +1,76 @@
+import { passOnSuccess, ScenarioMockApi, mockapi, json, MockApi } from "@azure-tools/cadl-ranch-api";
+
+export const Scenarios: Record<string, ScenarioMockApi> = {};
+
+function createMockApis(route: string, isUpdateRequest: boolean): MockApi {
+  const url = `/json-merge-patch/${route}`;
+  const expectedCreateBody = {
+    name: "Madge",
```
pshao25, line 8: What you expect user to send if they want to clear the dictionary?

pshao25: And a similar question is what you expect user to send if they want to add an item to the dictionary?

packages/cadl-ranch-specs/http/payload/json-merge-patch/main.tsp
haolingdong-msft: Currently I'm adding cases that service uses in priority, i.e. Record<InnerModel> and string. Also add some typical cases.

Generally if you covered `InnerModel[]` and `Record<InnerModel>`, `InnerModel` property should be covered fine. Typespec does not have primitive types, Java/.Net have, but if it's optional property, it's not primitive type as well. 

If .Net wants to have these two test cases specifically, I can add in this pr. But in the future, I think we can increamentally add test cases to json-merge-patch scenario when we encounter more use cases.

packages/cadl-ranch-specs/http/payload/json-merge-patch/mockapi.ts
haolingdong-msft: No difference with setting other properties to `null`.

packages/cadl-ranch-specs/http/payload/json-merge-patch/mockapi.ts
```diff
@@ -0,0 +1,84 @@
...
+    intArray: null,
+  };
+  if (isUpdateRequest) {
+    return mockapi.patch(url, (req) => {
+      req.expect.deepEqual(req.body.description, expectedUpdateBody.description);
+      req.expect.deepEqual(req.body.map, expectedUpdateBody.map);
+      req.expect.deepEqual(req.body.array, expectedUpdateBody.array);
+      req.expect.deepEqual(req.body.intValue, expectedUpdateBody.intValue);
+      req.expect.deepEqual(req.body.floatValue, expectedUpdateBody.floatValue);
+      req.expect.deepEqual(req.body.innerModel, expectedUpdateBody.innerModel);
```
weidongxu-microsoft, line 52: Same here, verification should only about the `description`, not `name` which is not modified.

packages/cadl-ranch-specs/http/payload/json-merge-patch/main.tsp
```diff
@@ -0,0 +1,248 @@
...
+Expected input body:
+```json
+{
+  "description": null,
+  "map": {
+    "key": {
+      "name": "InnerMadge",
+      "description": null
+    }
+  },
```
weidongxu-microsoft, line 221: You may want to include here a test to remove a key, e.g.

```json
  "map": {
    "key": {
      "name": "InnerMadge",
      "description": null
    },
    "key2": null
  },
```

packages/cadl-ranch-specs/http/payload/json-merge-patch/main.tsp
haolingdong-msft: Add the model `ResourcePatch` to include all the optional properties.

packages/cadl-ranch-specs/http/payload/json-merge-patch/main.tsp
```diff
@@ -0,0 +1,248 @@
...
+}
+```
+""")
+@doc("Test content-type: application/merge-patch+json with optional body")
+@route("/create/resource/optional")
+@put
+op createOptionalResource(
+  @header("content-type") contentType: "application/merge-patch+json",
+  @body body?: Resource,
+): Resource;
```
weidongxu-microsoft, line 206: Do we need this test?

`@put` may be useful for shared model. `body?` is useful for json-merge-patch. But `@put op create(body?)` seems not necessary?

weidongxu-microsoft (commented):
LGTM

My approval probably does not count, as I am Java as well...

packages/cadl-ranch-specs/http/payload/json-merge-patch/main.tsp
haolingdong-msft: I just add it to align with the required case (create+update).  I can remove it.

haolingdong-msft: Removed 'create optional resource' case

packages/cadl-ranch-specs/http/payload/json-merge-patch/main.tsp
```diff
@@ -0,0 +1,183 @@
...
+  "intArray": [1, 2, 3]
+}
+```
+
+Expected response body:
+```json
+{
+  "name": "Madge",
+  "description": "desc",
+  "map": {
```
lirenhe, line 74: Can we split this case into (1) a Patch model only contains basic types.   (2) The Patch model with complex types, e.g. array, records, inner models etc. 
The complexity of handling them are quite different.

haolingdong-msft: I think the priority regarding to these cases are the same, as there is service team using patch on `Map<InnerModel>` which is a complex case. I listed the case in the description of the pr. 
So I think those are all mandatory..

weidongxu-microsoft:
Another existing tsp source (already have SDK generated -- every createOrUpdate there is a PATCH)
op https://github.com/Azure/azure-rest-api-specs/blob/main/specification/communication/Communication.JobRouter/routes.tsp#L104
resource https://github.com/Azure/azure-rest-api-specs/blob/main/specification/communication/Communication.JobRouter/models.tsp#L457

packages/cadl-ranch-specs/http/payload/json-merge-patch/main.tsp
pshao25: I think we are good having all of these in the same model

weidongxu-microsoft (approved):

packages/cadl-ranch-specs/http/payload/json-merge-patch/main.tsp
haolingdong-msft: Thanks @pshao25 . I will resolve the comment, @lirenhe please let me know if you have futrhur concern.

packages/cadl-ranch-specs/http/payload/json-merge-patch/mockapi.ts
```diff
@@ -0,0 +1,83 @@
...
+    ],
+    intValue: 1,
+    floatValue: 1.1,
+    innerModel: {
+      name: "InnerMadge",
+      description: "innerDesc",
+    },
+    intArray: [1, 2, 3],
+  };
+  const expectedUpdateBody = {
```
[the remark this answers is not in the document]
haolingdong-msft: We only assert for updated properties, if the property is not changed, we will not verify, we will not throw error if you passed in the property that is not changed.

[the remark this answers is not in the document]
pshao25: I mean we should test the ability user could not send properties. For example, you expect user only sets description, so the expected payload only contains description and should not contain any other properties.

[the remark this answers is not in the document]
haolingdong-msft: I don't get it, you mean assert a particular field does not exist? But this depends on each language's design, from cadl-ranch, we allow client sends properties that are not updated together with properties that are updated.

pshao25 (approved):

PR #486, opened 2023-12-29
Support dashboard with multiple types

pshao25:
We are going to support not only azure SDK but also non-azure SDK. After that, each emitter would have the ability to generate both azure SDK and non-azure SDK. We hope all the mock apis defined in the cadl-ranch could test both azure and non-azure.

**Current storage structure:**
coverages (container)
|- manifest.json // scenario metadata
|- csharp
|    |- index.json // latest version
|    |- reports
|    |    |- 1.0.1.json
|    |    |- 1.0.2.json
|    |    |- // other versions
|- java
|    |- index.json // latest version
|    |- reports
|    |    |- 1.0.1.json
|    |    |- 1.0.2.json
|    |    |- // other versions
|- Other languages

**Changed storage structure:**
coverages (container)
|- manifest.json // scenario metadata with types (azure, non-azure)
|- csharp
|    |- index.json // latest version with types
|    |- reports
|    |    |- 1.0.1
|    |    |    |- azure.json
|    |    |    |- non-azure.json
|    |    |- 1.0.2
|    |    |    |- azure.json
|    |    |    |- non-azure.json
|    |    |- // other versions
|- java
|    |- index.json // latest version with types
|    |- reports
|    |    |- 1.0.1
|    |    |    |- azure.json
|    |    |    |- non-azure.json
|    |    |- 1.0.2
|    |    |    |- azure.json
|    |    |    |- non-azure.json
|    |    |- // other versions
|- Other languages

**Current user story:**
Only one report with all the scenario case. Run `cadl-ranch upload-coverage` to upload the report to above storage.

**Changed user story:**
One report for each type, therefore we will have two reports now, one for azure, the other for non-azure. Run `cadl-ranch upload-coverage` twice to upload these two reports, example `cadl-ranch upload-coverage --coverageFile xxx.json --generatorName python --storageAccountName azuresdkcadlranch --generatorVersion 3.2.1 --reportType unbranded` The command has one more argument `reportType`.
On the dashboard, it will have two tables with type marked in the table header. `python` and `typescript/modular` are examples which uploads both reports. `typescript/rlc` is an example only reports azure, while `java` is an example only reports non-azure. `csharp` is an example using currenct schema.
image

### DIFF
--- a/packages/cadl-ranch-coverage-sdk/src/client.ts
+++ b/packages/cadl-ranch-coverage-sdk/src/client.ts
@@ -152,14 +152,3 @@ async function readJsonBlob<T>(blobClient: BlockBlobClient): Promise<T> {
   const content = await body?.text();
   return content ? JSON.parse(content) : undefined;
 }
-
-async function readJsonBlobFromStream<T>(blobClient: BlockBlobClient): Promise<T> {
-  const blob = await blobClient.download();
-  let result = "";
-  if (blob.readableStreamBody) {
-    for await (const chunk of blob.readableStreamBody) {
-      result += chunk;
-    }
-  }
-  return result ? JSON.parse(result) : undefined;
-}
--- a/packages/cadl-ranch-coverage-sdk/src/client.ts
+++ b/packages/cadl-ranch-coverage-sdk/src/client.ts
@@ -61,7 +61,7 @@ export class CadlRanchCoverageOperations {
 
   public async upload(generatorMetadata: GeneratorMetadata, report: CoverageReport): Promise<void> {
     const blob = this.#container.getBlockBlobClient(
-      `${generatorMetadata.name}/reports/${generatorMetadata.version}/${generatorMetadata.type}.json`,
+      `${generatorMetadata.name}/reports/${generatorMetadata.version}/${generatorMetadata.mode}.json`,
     );
     const resolvedReport: ResolvedCoverageReport = { ...report, generatorMetadata };
     const content = JSON.stringify(resolvedReport, null, 2);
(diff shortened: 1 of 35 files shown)


timotheeguerin:
Are the 2 emitters not going to be just different emitter name. Can we just not have that as part of the emitter name like tupespexript did for rlc va modular.

branded vs unbranded just adds another parameter which might be confusing at some point.

packages/cadl-ranch-coverage-sdk/src/types.ts
```diff
@@ -33,6 +34,12 @@ export type ScenarioStatus =
   // Scenario is not applicable in current test.
   | "not-applicable";
 
+export type ScenarioType =
+  // Report for branded scenarios
+  | "branded"
```
msyyc, line 39: Compared with "branded", I prefer "azure".

timotheeguerin: yeah I kinda agree, I don't really like that name for someone not in this team this is very confusing. I think we should just show all the different emitters by name instead and have a way to filter so you can just show the azure emitters or omit them

pshao25: This type is actually not used. Removed and changed all the places related to type to azure and non-azure.

packages/cadl-ranch/src/coverage/scenario-manifest.ts
```diff
@@ -6,6 +6,8 @@ import { ScenarioLocation, ScenarioManifest } from "@azure-tools/cadl-ranch-cove
 import { getSourceLocation, normalizePath } from "@typespec/compiler";
 import { relative } from "path";
 
+const scenarioTypes = ["azure", "non-azure"];
```
timotheeguerin, line 9: I think  `non-azure` is not a great name for this. My understanding is that this should have 0 mention of azure.

Maybe just `standard`

packages/cadl-ranch/src/cli/cli.ts
```diff
@@ -267,6 +272,7 @@ async function main() {
           generatorName: args.generatorName,
           generatorVersion: args.generatorVersion,
           generatorCommit: args.generatorCommit ?? getCommit(process.cwd()),
+          reportType: args.reportType,
```
timotheeguerin, line 275: `reportType` doesn't make a lot of sense to me, it is not the type of the report but more the mode of the emitter you are using. Maybe just `--emitterMode` or `--mode`

packages/cadl-ranch-dashboard/src/components/dashboard.tsx
```diff
@@ -1,23 +1,44 @@
...
 import { DashboardTable } from "./dashboard-table.js";
+import { ResolvedCoverageReport } from "@azure-tools/cadl-ranch-coverage-sdk";
 export interface DashboardProps {
-  coverageSummary: CoverageSummary;
+  coverageSummaryAllTypes: CoverageSummaryAllTypes;
 }
 
-export const Dashboard: FunctionComponent<DashboardProps> = ({ coverageSummary }) => {
+export const Dashboard: FunctionComponent<DashboardProps> = ({ coverageSummaryAllTypes }) => {
+  const coverageSummaries: CoverageSummary[] = [];
```
timotheeguerin, line 10: All of this should be wrapped in a `useMemo` otherwise it will do the grouping everything it rerender

packages/cadl-ranch-coverage-sdk/src/client.ts
```diff
@@ -126,3 +160,14 @@ async function readJsonBlob<T>(blobClient: BlockBlobClient): Promise<T> {
   const content = await body?.text();
   return content ? JSON.parse(content) : undefined;
 }
+
+async function readJsonBlobFromStream<T>(blobClient: BlockBlobClient): Promise<T> {
```
timotheeguerin, line 164: Why is this needed? Isn't the method above already doing that? If it is better why not changing it above

packages/cadl-ranch-coverage-sdk/src/client.ts
```diff
@@ -61,49 +61,83 @@ export class CadlRanchCoverageOperations {
...
-  private async updateIndex(generatorName: string, version: string) {
-    const blob = this.#container.getBlockBlobClient(`${generatorName}/index.json`);
+  private async updateIndex(generatorName: string, version: string, type: string) {
     const data = {
       version,
+      types: [type],
     };
+
+    const blobClient = this.#container.getBlockBlobClient(`${generatorName}/index.json`);
+    if (await blobClient.exists()) {
```
timotheeguerin, line 132: Here if you want to stick with this approach you hsould be using the eTag and if match to make sure nothing else changed that file in the meantime or we might get a race condition that will override it. Which feels likely if you run the 2 set of test in parallel

timotheeguerin: an alternative would be to not contain the type but just use the blob list feature.

packages/cadl-ranch-dashboard/src/app.tsx
```diff
@@ -1,22 +1,22 @@
 import { FunctionComponent, useState } from "react";
-import { CoverageSummary, getCoverageSummary } from "./apis.js";
+import { CoverageSummaryAllTypes, getCoverageSummary } from "./apis.js";
 import { Dashboard } from "./components/dashboard.js";
 import { useEffectAsync } from "./utils.js";
 import { FluentProvider, webLightTheme } from "@fluentui/react-components";
 
 export const App: FunctionComponent = () => {
-  const [coverageSummary, setCoverageSummary] = useState<CoverageSummary | undefined>(undefined);
+  const [coverageSummaryAllTypes, setCoverageSummaryAllTypes] = useState<CoverageSummaryAllTypes | undefined>(undefined);
```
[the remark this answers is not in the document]
timotheeguerin: when you pass it to the DashboardTable you can just keep creating a `CoverageSummary` but with the right report filtered and instead add a new props that lets you label that table.

That will be better to be able to provide a more dynamic view in the future, as I feel like the double table will get hard to compare.

packages/cadl-ranch-coverage-sdk/src/client.ts
pshao25:
```js
/** Contains response data for the download operation. */
export declare type BlobDownloadResponseModel = BlobDownloadHeaders & {
    /**
     * BROWSER ONLY
     *
     * The response body as a browser Blob.
     * Always `undefined` in node.js.
     */
    blobBody?: Promise<Blob>;
    /**
     * NODEJS ONLY
     *
     * The response body as a node.js Readable stream.
     * Always `undefined` in the browser.
     */
    readableStreamBody?: NodeJS.ReadableStream;
    /** The underlying HTTP response. */
    _response: coreHttp.HttpResponse & {
        /** The parsed HTTP response headers. */
        parsedHeaders: BlobDownloadHeaders;
    };
};```

Above method is calling blobBody, which is browser only, so it could be used when we download from website. This method is calling readableStreamBody, which is nodejs only, so it is used when we upload index from command.

File: packages/cadl-ranch-dashboard/src/app.tsx (hunk 7 above)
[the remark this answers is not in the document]
Comment (pshao25) [reply packages/cadl-ranch-dashboard/src/app.tsx:8]: > I feel like the double table will get hard to compare

What are you thinking to compare? I think we are not expecting to compare the azure version and standard version.

File: packages/cadl-ranch-coverage-sdk/src/client.ts (hunk 6 above)
Comment (pshao25) [reply packages/cadl-ranch-coverage-sdk/src/client.ts:132]: I don't like etag solution because I don't want upload to fail.

The alternative solution still has race condition problem:
1. Version1.azure upload
2. Version2.azure upload
3. Version2.index update
4. Version1.index update

So I'm thinking sending the coverage files in a single command by making arg `coverageFile` an array, i.e.,  `cadl-ranch upload-coverage --coverageFile xxx1.json xxx2.json --generatorName python --storageAccountName azuresdkcadlranch --generatorVersion 3.2.1` and the mode of the coverage would be stored in the xxx1.json itself.

Comment (timotheeguerin) [reply packages/cadl-ranch-coverage-sdk/src/client.ts:132]: Well you keep retrying that’s the point of etag

File: packages/cadl-ranch-dashboard/src/apis.ts [hunk 8]
```diff
@@ -2,9 +2,10 @@ import {
   CadlRanchCoverageClient,
   ResolvedCoverageReport,
   ScenarioManifest,
+  GeneratorMode,
 } from "@azure-tools/cadl-ranch-coverage-sdk";
 
-const storageAccountName = "azuresdkcadlranch";
+const storageAccountName = "cadlranchstorage";
```
pshao25, line 8: Remember to revert before merge

packages/cadl-ranch-dashboard/src/components/dashboard.tsx
pshao25: I put all of these into `useEffectAsync` of `app.tsx`.

packages/cadl-ranch-coverage-sdk/src/client.ts
pshao25: But in my latest code, I don't need this anymore.

packages/cadl-ranch-coverage-sdk/src/client.ts
pshao25: I have some latest change seems don't have the race problem.

PR #491, opened 2024-01-09
test(model): add flattening cases

### LINKED ISSUE #471: Add flattenProperty tests
Add test for flattening of model in property. At least a basic scenario, a multi-depth scenario if we feel like it.

Need https://github.com/Azure/typespec-azure/issues/13

archerzz:
- add two test cases regarding model flattening:
  - single level flatten
  - two levels flatten
- add mock api

resolve #471


archerzz:
This adds test cases for https://github.com/Azure/typespec-azure/pull/94

Since there is no official release of `typespec-client-generator-core`, I'm using a dev build. Thus, this PR is marked as draft. Need to update to official release version later.

archerzz:
Per discussion, the naming conflict case will be added later. We probably need to add support in TCGC models, like moduler-four.

packages/cadl-ranch-specs/http/azure/client-generator-core/flatten/main.tsp
```diff
@@ -0,0 +1,107 @@
...
+@scenarioService("/type/model/flatten")
+namespace Type.Model.Flatten;
+
+@doc("This is the model with one level of flattening.")
+model FlattenModel {
+  name: string;
+
+  #suppress "deprecated" "@flattenProperty decorator is not recommended to use."
+  @flattenProperty
+  child: ChildModel;
```
msyyc, line 18: It is better if use `properties` and `typeProperties` of real cases.

archerzz: rename `child` to `properties`

packages/cadl-ranch-specs/http/azure/client-generator-core/flatten/main.tsp
```diff
@@ -0,0 +1,107 @@
...
+
+Expected response body:
+```json
+{
+  "name": "test",
+  "properties": {
+    "description": "test",
+    "age": 1
+  }
+}
```
weidongxu-microsoft, line 68: nit, there seems no reason to have response json be different from request json?

archerzz: yes. I intentionally make it different so that people cannot be lazy when composing test cases in each language...

msyyc (approved):

weidongxu-microsoft (approved):
Not sure if you need to re-run changeset, as other packages probably also be a minor version update, given the change to these package.json

PR #493, opened 2024-01-12
introduce cadl ranch test cases for format on properties

ArcturusZhang:

### DIFF
--- a/packages/cadl-ranch-specs/http/azure/client-generator-core/format/main.tsp
+++ b/packages/cadl-ranch-specs/http/azure/client-generator-core/format/main.tsp
@@ -1,0 +1,126 @@
+import "@typespec/http";
+import "@azure-tools/cadl-ranch-expect";
+import "@azure-tools/typespec-client-generator-core";
+
+using TypeSpec.Http;
+using Azure.ClientGenerator.Core;
+
+/**
+ * Describe scenarios for various combinations of unions.
+ */
+@supportedBy("dpg")
+@scenarioService("/azure/client-generator-core/format")
+namespace _Specs_.Azure.ClientGenerator.Core.Format;
+
+model Model {
+  @format("guid")
+  guidProperty?: string;
+
+  @format("url")
+  urlProperty?: string;
+
+  @format("uuid")
+  uuidProperty?: string;
+
+  @format("armId")
+  armIdProperty?: string;
+
+  @format("ipAddress")
+  ipAddressProperty?: string;
+
+  @format("azureLocation")
+  azureLocation?: string;
+
+  @format("etag")
+  etagProperty?: string;
+}
+
+model InputModel is Model;
+
+model OutputModel is Model;
+
+model RoundTripModel is Model;
+
+/**
+ * Consumes an InputModel with properties of various formats
+ */
+@route("/input")
+@post
+@scenario
+@scenarioDoc("""
+  Verify properties with various of formats can be processed in a request.
+
+  Expected request body:
+  ```json
+  {
+    "guidProperty": "3f2504e0-4f89-11d3-9a0c-0305e82c3301",
+    "urlProperty": "https://www.example.com/path/to/resource",
+    "uuidProperty": "3f2504e0-4f89-11d3-9a0c-0305e82c3301",
+    "armIdProperty": "/subscriptions/3f2504e0-4f89-11d3-9a0c-0305e82c3301/resourceGroups/test/providers/Microsoft.Compute/virtualMachines/name",
+    "ipAddressProperty": "192.168.1.1",
+    "azureLocation": "eastus",
+    "etagProperty": "33a64df551425fcc55e4d42a148795d9f25f89d4"
+  }
+  ```
+  """)
+op inputWithFormat(@body body: InputModel): void;
+
+/**
+ * Produces an OutputModel with properties of various formats
+ */
+@route("/output")
+@post
+@scenario
+@scenarioDoc("""
+  Verify properties with various of formats can be processed in a response.
+
+  Expected response body:
+  ```json
+  {
+    "guidProperty": "550e8400-e29b-41d4-a716-446655440000",
+    "urlProperty": "https://www.example.com/path/to/another/resource",
+    "uuidProperty": "550e8400-e29b-41d4-a716-446655440000",
+    "armIdProperty": "/subscriptions/550e8400-e29b-41d4-a716-446655440000/resourceGroups/test/providers/Microsoft.Compute/virtualMachineScaleSets/name",
+    "ipAddressProperty": "172.16.254.1",
+    "azureLocation": "westus",
+    "etagProperty": "aabbccddeeff00112233445566778899"
+  }
+  ```
+  """)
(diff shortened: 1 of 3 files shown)


packages/cadl-ranch-specs/http/azure/client-generator-core/format/main.tsp
```diff
@@ -0,0 +1,126 @@
...
+  armIdProperty?: string;
+
+  @format("ipAddress")
+  ipAddressProperty?: string;
+
+  @format("azureLocation")
+  azureLocation?: string;
+
+  @format("etag")
+  etagProperty?: string;
```
msyyc, line 35: How many valid stuff for `@format`? I can't find any doc describes it.

weidongxu-microsoft: `url` should be a scalar. https://microsoft.github.io/typespec/language-basics/built-in-types#string-types

`@format` seems just the "format" in swagger, which can be anything.

For Java, these are just all `String`. So no opinion from me...

ArcturusZhang: > `url` should be a scalar. [microsoft.github.io/typespec/language-basics/built-in-types#string-types](https://microsoft.github.io/typespec/language-basics/built-in-types#string-types)
> 
> `@format` seems just the "format" in swagger, which can be anything.
> 
> For Java, these are just all `String`. So no opinion from me...

it is kind of required for dotnet MPG, therefore I introduced these cases

packages/cadl-ranch-specs/http/azure/client-generator-core/format/main.tsp
```diff
@@ -0,0 +1,126 @@
...
+    "urlProperty": "https://www.example.com/path/to/resource",
+    "uuidProperty": "3f2504e0-4f89-11d3-9a0c-0305e82c3301",
+    "armIdProperty": "/subscriptions/3f2504e0-4f89-11d3-9a0c-0305e82c3301/resourceGroups/test/providers/Microsoft.Compute/virtualMachines/name",
+    "ipAddressProperty": "192.168.1.1",
+    "azureLocation": "eastus",
+    "etagProperty": "33a64df551425fcc55e4d42a148795d9f25f89d4"
+  }
+  ```
+  """)
+op inputWithFormat(@body body: InputModel): void;
```
weidongxu-microsoft, line 66: nit, I don't think you need to repeat "WithFormat" in a test where everything is about format.

packages/cadl-ranch-specs/http/azure/client-generator-core/format/main.tsp
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ * Produces an OutputModel with properties of various formats
+ */
+@route("/output")
+@post
```
weidongxu-microsoft, line 72: `@get`?

weidongxu-microsoft (commented):
LGTM

Make sure you've tested it.

packages/cadl-ranch-specs/http/azure/client-generator-core/format/main.tsp
ArcturusZhang: sure. I will just change those to `input`, `output` and `roundtrip`

packages/cadl-ranch-specs/http/azure/client-generator-core/format/main.tsp
ArcturusZhang: well since post could also accepts nothing and returns you something, considering the other two are post, I choose post here.

packages/cadl-ranch-specs/http/azure/client-generator-core/format/main.tsp
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ * Describe scenarios for various combinations of unions.
+ */
+@supportedBy("dpg")
+@scenarioService("/azure/client-generator-core/format")
+namespace _Specs_.Azure.ClientGenerator.Core.Format;
+
+model Model {
+  @format("guid")
```
timotheeguerin, line 16: What is the goal of this test? `@format` is only to be used as a validation tool which our SDK don't do, I worry that if we even add this test it will lead to believe you should create types for those known formats

ArcturusZhang: the goal of this test is kind of creating a common platform for all the formats that every language should properly handle. From each language's perspective, they could have their own implementation for these formats but the wire content must remain the same.

ArcturusZhang: Oh and most importantly, MPG tsp will need these format to work properly without breaking changes, therefore I believe it benefits if we have the format we supported in a centralized place.
And also this is the reason we put these test cases under `azure` instead of `type/property`

timotheeguerin: Again those formats are just an indication of the pattern of the string. The type requested here is a string not what the format said. If a different format should be used a known scalar should be created.
please do not do this

ArcturusZhang: But during the generation of MPG libraries, we never want undeterministic results, such as if the tsp owner defines its own scalar `azureLocation`, we will not generate that type as `Azure.Core.AzureLocation`.
We want only one specific type defined somewhere in our resourcemanager library to be parsed differently from a random scalar our customer defined.

I heard TCGC now just trims off scalar and returns us its underlying type, if this is true, it is not acceptable for MPG

timotheeguerin: Tcgc should be aware of all the established built in scalar we have and reduce to those not use format.

weidongxu-microsoft: There were one in typespec-azure-core of `eTag`
https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-core/lib/models.tsp#L343-L345

I do not mean this is the expected definition. Just a discussion about whether we should do e.g. `@format(eTag)` or something like it. If we do, should we prefer it is done as `scalar guid is ...` in typespec-azure-resource-manager lib (instead of in service ts files)

weidongxu-microsoft: typespec-azure-resource-manager got `arm-id`
https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-resource-manager/lib/models.tsp#L100-L108

timotheeguerin: there is a linter rule telling you not to use `@format` in azure core. Those `@format` on those types are to match to some known openapi/autorest formats but they are not meant to be consumed outside. We found out that we shouldn't even have added `@format` for `eTag` because it isn't a known format to autorest so caused all kind of lint issues.

PR #495, opened 2024-01-16
specs, update multipart complex case, make "pictures" optional

weidongxu-microsoft:
Translate service uses optional -- `glossary?: bytes[]`
https://github.com/Azure/azure-rest-api-specs/pull/26667/files#r1451932014

So update the complex case for this.


msyyc (approved):

lirenhe (approved):

packages/cadl-ranch-specs/http/payload/multipart/mockapi.ts
```diff
@@ -83,8 +83,34 @@ function createMockApis(route: string, checkList: ((param: MockRequest) => void)
 
 Scenarios.Payload_MultiPart_FormData_basic = passOnSuccess(createMockApis("mixed-parts", [checkId, checkProfileImage]));
 
-Scenarios.Payload_MultiPart_FormData_complex = passOnSuccess(
-  createMockApis("complex-parts", [checkId, checkAddress, checkPreviousAddresses, checkAllFiles]),
+Scenarios.Payload_MultiPart_FormData_complex = withKeys(["two-pictures", "no-pictures"]).pass(
+  mockapi.post("/multipart/form-data/complex-parts", (req) => {
```
timotheeguerin, line 87: Wouldn't it be better to do that check in the checkPictures instead of hte uber test that contains everything(think we shouldn;t even have that one or just be a smoke test)

Also I wouldn't do the call twice comment people will not read it. Make 2 differernt endpoint and scenarios

weidongxu-microsoft: The test suite already contains all sorts of smaller pieces of the multipart case (simple scalar, model as JSON, 1 file, 2 files, etc.). Here I only modify the most complex one.

Sure, let me see if I break it, or just test the "two-pictures" case as the origin case (personally I only want the `pictures?`).

timotheeguerin: that is my point the smaller ones should be the ones doing the test, the complex one otherwise ends up being this enormous mix where we need to test a million thing.
- small ones test each individual part types
- complex one is just a smoke test to make sure you can have all the parts together

lmazuel: @weidongxu-microsoft agree with @timotheeguerin , let's make `pictures` optional in BinaryArrayPartsRequest instead

weidongxu-microsoft: Got it. Just add a new case with 
```ts
model OptionalBinaryArrayPartsRequest {
  profileImage: bytes;
  pictures?: bytes[];
}
```

Existing `BinaryArrayPartsRequest` could still test the "required" scenario.

PR #504, opened 2024-01-25
add test case which contains anonymous model for multipart/form-data

msyyc:
real service: https://github.com/Azure/azure-rest-api-specs/blob/b414505c35526e7f6c64c9351d2bf97ee95e7ff5/specification/purview/Azure.Analytics.Purview.DataMap/routes.tsp#L763


packages/cadl-ranch-specs/http/payload/multipart/main.tsp
```diff
@@ -303,4 +308,35 @@ Content-Type: image/jpg
...
+Content-Type: application/octet-stream;
+
+{…file content…}
+--abcde12345--
+```
+""")
+  @doc("Test content-type: multipart/form-data")
+  @post
+  @route("/anonymous-model")
+  op anonymousModel(@header contentType: "multipart/form-data", ...MultiPartAnonymousRequest): NoContentResponse;
```
weidongxu-microsoft, line 341: I still don't see why `id` needs to get in there.

Can we do this?
```suggestion
  op anonymousModel(@header contentType: "multipart/form-data", profileImage: MultiPartRequest.profileImage): NoContentResponse;
```

Fine to wait for opinion from others. Not a strong opinion here.

lmazuel: +1 with @weidongxu-microsoft

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -1602,6 +1602,36 @@ Expected request body is "foo".
...
+Expect request (
+
+- according to https://datatracker.ietf.org/doc/html/rfc7578#section-4.4, content-type of file part shall be labeled with
+  appropriate media type, cadl-ranch will check it; content-type of other parts is optional, cadl-ranch will ignore it.
+- according to https://datatracker.ietf.org/doc/html/rfc7578#section-4.2, filename of file part SHOULD be supplied.
+  If there are duplicated filename in same filedName, cadl-ranch can't parse them all.
+  ):
+
+```
+POST /upload HTTP/1.1
```
weidongxu-microsoft, line 1618: nit, path in desc is incorrect.

packages/cadl-ranch-specs/http/payload/multipart/main.tsp

weidongxu-microsoft (approved):

lmazuel (approved):
Note that in Azure, service teams are required to provide a name, so this is not blocking, but yes we should solve it for unbranding

weidongxu-microsoft:
> Note that in Azure, service teams are required to provide a name, so this is not blocking, but yes we should solve it for unbranding

Actually, OpenAI Assistants did exactly this :-(
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/ai/OpenAI.Assistants/files/routes.tsp#L43-L56

I remember DataMap doing this as well...

PR #516, opened 2024-02-08
Use unambiguous float32 values

### LINKED ISSUE #512: Update `42.42` float32 constant to something else
`42.42` float32 constant is used in several places in cadl-ranch specs: https://github.com/search?q=repo%3AAzure%2Fcadl-ranch%2042.42&type=code

The problem is that `42.42` is not exactly representable in IEEE754.
The closest float32 representation is `42.4199981689453125`.

I wrote a test that passes `42.42f` to cadl-ranch, and it does fail, because `42.42f` is `42.4199981689453125`, and I get HTTP 400 back, because I think you expect the JSON object you get, to have 42.42 exactly (or, more likely, likely a float64 representation for it, `42.4200000000000017`, which we can't exactly pass, unless we tell our implementation to serialize floats32 as floats64).
That is not possible. In C++, we pass `42.42f`, and it correctly gets serialized as `42.4199981689453125`. But cadl-ranch testserver doesn't like that. I think, the other option for it would be to accept that value as well.

But I think it is better to change `42.42` to something that is exactly representable in IEEE754, for example - `42.125`.
See https://www.h-schmidt.net/FloatConverter/IEEE754.html if you want to experiment with different values.

antkmsft:
Fixes #512.
I verified with C++ tests - works perfectly (https://github.com/Azure/autorest.cpp/pull/317).

I do not know what level of details you typically put in the changelog messages, feel free to correct my wording.


antkmsft:
changeset add --empty

timotheeguerin (approved):
Changelog is good, just there to give others a good idea of what changed when they pull the new specs

weidongxu-microsoft (approved):

PR #518, opened 2024-02-19
specs, use union for extensible enum

weidongxu-microsoft:
1. use `union` for current "extensible enum"

<details>
<summary>Dependency version</summary>
Need to use Azure.Core 0.53.1.

0.53.0 has such validation error

```
error: ✘ Union contains non-string value type Scalar.: C:/github/cadl-ranch/node_modules/.pnpm/@typespec+compiler@0.53.0/node_modules/@typespec/compiler/lib/lib.tsp:90:1
error: ✘ Union contains non-string value type Scalar.: C:/github/cadl-ranch/node_modules/.pnpm/@typespec+compiler@0.53.0/node_modules/@typespec/compiler/lib/lib.tsp:90:1
error: ✘ The operation linked in  @pollingOperation must return a valid status monitor.  The status monitor model must contain a 'status' property, or a property decorated with  '@lroStatus'.  The status field must be of Enum or Union type and contain terminal status values for success and failure.: C:/github/cadl-ranch/packages/cadl-ranch-specs/http/azure/core/lro/rpc-legacy/main.tsp:142:3
error: ✘ ✘ Scenario azure/core/lro/rpc-legacy is invalid.
C:\github\cadl-ranch\packages\cadl-ranch-specs:
 ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  @azure-tools/cadl-ranch-specs@0.30.0 validate-scenarios: `cadl-ranch validate-scenarios ./http`
```
</details>


packages/cadl-ranch-specs/http/type/enum/extensible/main.tsp
```diff
@@ -7,27 +7,29 @@ using TypeSpec.Http;
 namespace Type.Enum.Extensible;
 
 @doc("Days of the week")
-enum DaysOfWeekExtensibleEnum {
+union DaysOfWeekExtensibleEnum {
```
weidongxu-microsoft, line 10: `getKnownValue(): DaysOfWeekExtensibleEnum;` likely have problem even in TCGC.

tsp compiler would think this be multiple responses. (`HttpOperationResponse` which contains multiple `HttpOperationResponseContent` in `responses` property)

weidongxu-microsoft: Left a message in https://github.com/microsoft/typespec/issues/2848#issuecomment-1952088905

It is pretty hard for emitter to handle this case, when the Union is already taken apart to Type.

weidongxu-microsoft: Changed to use explicit `@body`
```ts
  getKnownValue(): {
    @body body: DaysOfWeekExtensibleEnum;
  };
```
on Timothee's advice. This should unblock emitter.

packages/cadl-ranch-specs/http/type/enum/fixed/main.tsp
```diff
@@ -8,8 +8,6 @@ using TypeSpec.Http;
 @scenarioService("/type/enum/fixed")
 namespace Type.Enum.Fixed;
 
-#suppress "@azure-tools/typespec-azure-core/use-extensible-enum" "For testing"
-@fixed
```
lmazuel, line 12: Too soon for this change, see derails in long email I'm writing right now

packages/cadl-ranch-specs/http/type/property/value-types/main.tsp
```diff
@@ -119,17 +119,17 @@ interface Duration extends ModelOperations<DurationProperty, "P123DT22H14M12.011
 
 // Test a model with an enum property
 @doc("Enum that will be used as a property for model EnumProperty. Non-extensible.")
-enum InnerEnum {
+union InnerEnum {
```
lmazuel, line 122: The doc says explicitly this is non-extensible, we should add `@fixed` instead

weidongxu-microsoft: There is a `FixedInnerEnum` after this model. So I tend to think the doc was wrong. I would update the doc.

msyyc (approved):

PR #511, opened 2024-01-30
test for `@clientName` of TCGC and `@encodedName` of compiler

msyyc:
fixes https://github.com/Azure/cadl-ranch/issues/508, https://github.com/Azure/cadl-ranch/issues/509

- These cases are almost same with existing cases for `@projectedName` so devs shall be able to add the test with few cost

PR #519, opened 2024-02-23
Split clientName and encodedName tests

timotheeguerin:
Previous [PR](https://github.com/Azure/cadl-ranch/pull/511) added those but no new version was released so this isn't breaking anybody 

Splitting the 2 so the naming stays in teh client category and the projected name moves out of the `projection` category that is different.

@msyyc FYI, thanks for the original PR we just noticed that today when looking at the dashboard that it would be better to have this new structure.

### DIFF
--- a/packages/cadl-ranch-specs/cadl-ranch-summary.md
+++ b/packages/cadl-ranch-specs/cadl-ranch-summary.md
@@ -670,7 +670,7 @@ Expected response body:
 }
 ```
 
-### Client_Naming_header
+### Client_Naming_Header_request
 
 - Endpoint: `post /client/naming/header`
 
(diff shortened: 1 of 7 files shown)


packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -670,6 +670,99 @@ Expected response body:
...
+Testing that we can project the client name and the wire name.
+Your generated SDK should generate ClientNameAndJsonEncodedNameModel with one property with client name `clientName` and wire name `wireName`.
+
+Expected request body:
+
+```json
+{ "wireName": true }
+```
+
+### Client_Naming_Property_language
```
iscai-msft, line 753: the endpoints for these new tests still show the projection/client-name-and-encoded-name url. change the urls, and grep "project" and change it to rename so we completely remove "projection"

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -670,6 +670,99 @@ Expected response body:
...
 ```
 
+### Client_Naming_header
+
+- Endpoint: `post /client/naming/header`
+
+Testing that we can project a header name.
+Your generated SDK should generate an operation header `parameter` with a single parameter called `clientName`.
+
+Expected header parameter: `default-name="true"`
```
iscai-msft, line 680: small nit: would this be better split into `client/naming/header/request` and `client/naming/header/response`

iscai-msft (approved):

PR #524, opened 2024-02-28
specs, add test case to type/property/value-types, property has type of enum value

weidongxu-microsoft:
Real case in OpenAI (currently it is value from Enum; it should be changed to value from Union, upon [PR](https://github.com/Azure/azure-rest-api-specs/pull/27687/files#diff-d40c5083c3e6eab3b0a599ba87fb01d60283f82fb428e0f5014a5a3654c16603))
https://github.com/Azure/azure-rest-api-specs/blob/ca6cb3595fb0d091691d09e76594546c01b074ea/specification/cognitiveservices/OpenAI.Inference/models/completions/extensions/oyd/azure_cognitive_search_options.tsp#L22


packages/cadl-ranch-specs/http/type/property/value-types/main.tsp
```diff
@@ -269,3 +269,19 @@ model UnionFloatLiteralProperty is ModelTemplate<43.125 | 46.875>;
...
+
+union BaseEnum {
+  string,
+  EnumValue1: "value1",
+}
+
+union ExtendedEnum {
+  BaseEnum,
+  EnumValue2: "value2",
+}
```
weidongxu-microsoft, line 281: I use the extended enum, as dev could write `property: BaseEnum.EnumValue1` as well.

Guess we first get the basic case covered.

packages/cadl-ranch-specs/http/type/property/value-types/main.tsp
```diff
@@ -144,13 +144,13 @@ enum FixedInnerEnum {
 model EnumProperty is ModelTemplate<FixedInnerEnum>;
 @route("/enum")
 @operationGroup
-interface Enum extends ModelOperations<EnumProperty, "ValueOne"> {}
+interface Enum extends ModelOperations<EnumProperty, "\"ValueOne\""> {}
```
msyyc, line 147: Does the change only influence summary.md?

weidongxu-microsoft: Yes, this part on `ValueOne` and `UnknownValue` is only for the fix on summary doc.

packages/cadl-ranch-specs/http/type/property/value-types/main.tsp
tadelesh: Seems weird. Why we don't use:
```
union BaseEnum {
  string,
  EnumValue1: "value1",
  EnumValue2: "value2",
}
```
directly?

weidongxu-microsoft: Dev may want to "extend" a Enum.

And it is already in use
https://github.com/Azure/azure-rest-api-specs/pull/27749#discussion_r1494047254

tadelesh: Checked the logic for `getUnionAsEnum` and it support such defintion.

weidongxu-microsoft: Yeah, that is the "flatten" on `flattenedMembers`

weidongxu-microsoft: I've removed the hierarchy in this case, as the expected behavior is in discussion.

tadelesh (approved):

MaryGao (approved):
LGTM

pshao25 (approved):

PR #525, opened 2024-02-28
specs, test client name on enum and its value

weidongxu-microsoft:
TCGC should have handled these in their common layer.


packages/cadl-ranch-specs/http/client/naming/main.tsp
```diff
@@ -187,3 +187,31 @@ namespace Model {
...
+    (The exact name may depend on language convention)
+
+    Expected request body:
+    ```json
+    "value1"
+    ```
+  """)
+  @route("/client")
+  @post
+  op client(@body body: ExtensibleEnum): NoContentResponse;
```
timotheeguerin, line 216: Maybe better to separate in 2 operation so it shows up as that in the dashboard.

```
op enumName

op enumMemberName
```

packages/cadl-ranch-specs/http/client/naming/main.tsp
```diff
@@ -187,3 +187,50 @@ namespace Model {
...
+    string,
+    EnumValue1: "value1",
+  }
+
+  union ExtensibleEnum {
+    string,
+
+    @clientName("ClientEnumValue1")
+    EnumValue1: "value1",
+  }
```
tadelesh, line 205:
```suggestion
  union ExtensibleEnum {
    string,

    @clientName("ClientEnumValue1")
    EnumValue1: "value1",
    
    @clientName("ClientEnumValue2")
    "value2",
  }
```
could you add this case?

weidongxu-microsoft: Added. Interesting, it is valid.

packages/cadl-ranch-specs/http/client/naming/main.tsp
weidongxu-microsoft: Sure. Separated to 2 scenarios.

tadelesh (approved):

packages/cadl-ranch-specs/http/client/naming/main.tsp
```diff
@@ -187,3 +187,53 @@ namespace Model {
...
+    (The exact name may depend on language convention)
+
+    Expected request body:
+    ```json
+    "value1"
+    ```
+  """)
+  @route("/enum-member-name")
+  @post
+  op enumMemberName(@body body: ExtensibleEnum): NoContentResponse;
```
MaryGao, line 238: i may not get the idea of enum-member-name here because i notice the body type is union(finally enum) not enum member?

MaryGao: oh this is for client name testing, i just image the case below:
```
op enumMemberName(@body body: ExtensibleEnum.EnumValue1): NoContentResponse;
```

weidongxu-microsoft: I could call it e.g. `union-enum-member-name` is that's sound better to you :-)

Personally I don't have good idea to say in the method name that "this type is union but it should be treated as enum..."

MaryGao (approved):
Thanks for this! LGTM and just confirm a small case.

packages/cadl-ranch-specs/http/client/naming/main.tsp
MaryGao: haha, interesting i am wondering if we can add @clientName for real tsp enum?

pshao25 (approved):

packages/cadl-ranch-specs/http/client/naming/main.tsp
weidongxu-microsoft: Sure. I've added prefix `union` to the case, so that we can use `enum` in future when do fixed enum.

PR #528, opened 2024-02-29
Remove deprecated service version

timotheeguerin:


packages/cadl-ranch-specs/http/client/structure/common/service.tsp
```diff
@@ -29,7 +29,6 @@ using Azure.ClientGenerator.Core;
 )
 @service({
   title: "MultiClient",
-  version: "1.0.0",
```
qiaozha, line 32: we should add `@versioned` to avoid breaking ?

timotheeguerin: this should not have been used and is the entire point why we deprecating `version:`

qiaozha: oh because there's no api version parameter in it? got it, thanks

qiaozha (approved):

weidongxu-microsoft (approved):

packages/cadl-ranch-specs/http/client/structure/common/service.tsp
timotheeguerin: yeah especially but either way

PR #451, opened 2023-11-14
Add test for decimal type

tadelesh:
# Cadl Ranch Contribution Checklist:

- [x] I have written a [scenario spec](../docs/writing-scenario-spec.md)
- [x] I have **meaningful** `@scenario` names. Someone can look at the list of scenarios and understand what I'm covering.
- [x] I have written a [mock API](../docs/writing-mock-apis.md)
- [x] I have used `@scenarioDoc`s for extra scenario description and to tell people **how to pass** my mock api check.

PR #535, opened 2024-03-12
remove-customer-scenario-assumption-for-decimal

qiaozha:
remove the decimal test scenario for 0.1 + 0.1 + 0.1 = 0.3 as it's about customer side scenario, which is out of the scope of cadl ranch testing. If service side expect their customer to use decimal for math calculation, we should suggestion them to use `@encode(string)`, which is pending https://github.com/microsoft/typespec/issues/2762

### DIFF
--- a/.changeset/tall-buses-run.md
+++ b/.changeset/tall-buses-run.md
@@ -1,0 +1,6 @@
+---
+"@azure-tools/cadl-ranch-specs": patch
+"@azure-tools/cadl-ranch": patch
+---
+
+remove-customer-scenario-assumption-for-decimal
--- a/packages/cadl-ranch-specs/cadl-ranch-summary.md
+++ b/packages/cadl-ranch-specs/cadl-ranch-summary.md
@@ -6151,23 +6151,6 @@ Expected response body:
 0.33333
 ```
 
-### Type_Scalar_Decimal128Verify_prepareVerify
-
-- Endpoint: `get /type/scalar/decimal128/prepare_verify`
-
-Get verify values:
-[0.1, 0.1, 0.1]
-
-### Type_Scalar_Decimal128Verify_verify
-
-- Endpoint: `post /type/scalar/decimal128/verify`
-
-Expected input body:
-
-```json
-0.3
-```
-
 ### Type_Scalar_DecimalType_requestBody
 
 - Endpoint: `put /type/scalar/decimal/resquest_body`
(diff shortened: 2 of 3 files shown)


weidongxu-microsoft:
I assume the intention was to make sure the `decimal` type works like a decimal, instead of float.

Do we know the owner of this 2 tests?

Java is kind of fine, we had ours in emitter https://github.com/Azure/autorest.java/blob/main/typespec-tests/src/test/java/com/type/scalar/DecimalTests.java#L34-L54

qiaozha:
I think @tadelesh added this https://github.com/Azure/cadl-ranch/pull/451

qiaozha:
> I assume the intention was to make sure the `decimal` type works like a decimal of infinite precision, instead of float.
> 
> Do we know the owner of this 2 tests?
> 
> Java is kind of fine, we had ours in emitter https://github.com/Azure/autorest.java/blob/main/typespec-tests/src/test/java/com/type/scalar/DecimalTests.java#L34-L54

I noticed that in java test, you are using string to create the BigDecimal too.

weidongxu-microsoft:
> I noticed that in java test, you are using string to create the BigDecimal too.

This is for Java type for that `decimal`. `BigDecimial` has ctor of `String`.
I can use other ctor e.g. https://docs.oracle.com/javase/8/docs/api/java/math/BigDecimal.html#valueOf-long-int-
Just choose the easy one.

archerzz (approved):

tadelesh:
The test is aimed to make sure client side generate the right precision for decimal. .NET has native decimal128 type, which means not all languages need to use string to do ser/deser.

tadelesh:
> > The test is aimed to make sure client side generate the right precision for decimal. .NET has native decimal128 type, which means not all languages need to use string to do ser/deser.
> 
> the math calculation is suggesting we are testing customer user scenarios which is out of the scope of cadl-ranch, however, what cadl ranch should be testing is the type expression and api interface.

Since we don't have a good way to test the edge number of decimal, we use this calculation to ensure client generate the right type. If removed, do you mean language generator should ensure it?

lmazuel:
We need test that check if the TypeSpec has Decimal as input or output, the SDK has a strategy to deal with it. The exact strategy for that, is up for SDK. Even if we're adding `encode(string)`, we still need "naked" tests with no decorators. So we should not remove those tests.

However, the test is designed to force the SDK person to sum 0.1+0.1+0.1 on the client side and get 0.3 in return, which is a math operation that requires Decimal support. I'm ok if TypeScript were to hardcode to return 0.3, with a comment that says that given JS limitation, it's not possible to do 0.1+0.1+0.1 in JS. I read all the JS issues, and @bterlson long comment about it, I don't think this is in opposition of the spirit of what we are trying to accomplish. IOW, I'd like language that can to have the possibility to show what they can do, without forcing languages that can't to be blocked. Removing the test, remove the ability to write anyything at all, which I think is a contradiction of the intent of cadl-ranch.

qiaozha:
I am fine to not remove it and work around it in JS by passing 0.3 for this case.

PR #556, opened 2024-04-11
additional-properties-add-test-cases-for-spread-of-record

qiaozha:
fix https://github.com/Azure/cadl-ranch/issues/546

### DIFF
--- a/packages/cadl-ranch-specs/http/type/property/additional-properties/main.tsp
+++ b/packages/cadl-ranch-specs/http/type/property/additional-properties/main.tsp
@@ -386,6 +386,61 @@ interface SpreadDifferentModelArray
       "{'knownProp': 'abc', 'prop': [{'state': 'ok'}, {'state': 'ok'}]}"
     > {}
 
+// ****************** extends model has spread Record<string> **************************
+@doc("The model extends from a model that spread Record<string> with the different known property type")
+model DifferentSpreadStringDerived extends DifferentSpreadStringRecord {
+  @doc("The index property")
+  derivedProp: string;
+}
+
+@route("/extendsDifferentSpreadString")
+@operationGroup
+interface ExtendsDifferentSpreadString
+  extends ModelOperations<DifferentSpreadStringDerived, "{'id': 43.125, 'prop': 'abc', 'derivedProp': 'abc'}"> {}
+
+// ****************** extends model has spread Record<float32> **************************
+@doc("The model extends from a model that spread Record<float32> with the different known property type")
+model DifferentSpreadFloatDerived extends DifferentSpreadFloatRecord {
+  @doc("The index property")
+  derivedProp: float32;
+}
+
+@route("/extendsDifferentSpreadFloat")
+@operationGroup
+interface ExtendsDifferentSpreadFloat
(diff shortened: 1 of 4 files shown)


packages/cadl-ranch-specs/http/type/property/additional-properties/main.tsp
```diff
@@ -231,35 +256,306 @@ model ModelForRecord {
...
 @operationGroup
 interface IsModelArray
-  extends ModelOperations<IsModelArrayAdditionalProperties, "{'prop': [{'state': 'ok'}, {'state': 'ok'}]}"> {}
+  extends ModelOperations<
+      IsModelArrayAdditionalProperties,
+      "{'knownProp': [{'state': 'ok'}, {'state': 'ok'}], 'prop': [{'state': 'ok'}, {'state': 'ok'}]}"
+    > {}
+
+model SpreadModelArrayRecord {
+  knownProp: ModelForRecord[];
```
weidongxu-microsoft, line 323: Why we use `knownProp: ModelForRecord[];` here? I assume we could use any type (not limited to `ModelForRecord[]`)?

qiaozha: we are, but I was trying to show that if the known properties type is the same as the additional properties type, we can have three ways to express that. 1, extends Record, 2. is Record, 3. spread of Record. 
I have another case here to show different types here https://github.com/Azure/cadl-ranch/pull/556/files#diff-9ab37e973a0f0d7e61ea487e765530d01f107dc0e5fec04b9dcbc829acd8a48aR375-R387

weidongxu-microsoft: Sure. Just wondering if we only want to show that "known properties type is the same as the additional properties type", we may use simpler type (instead of an array).

No strong opinion here. Just I don't know if array works for current Java emitter :-(

weidongxu-microsoft (approved):
I did a quick look and nothing seems wrong.

Count me as half of one approver.

@haolingdong-msft @XiaofeiCao from Java as well, if see anything apparently not good, please point out.

msyyc:
Python has no work for additional properties so it is better to get approval from JAVA/.NET

haolingdong-msft:
wonder do we have service using this spread of `Record`?

packages/cadl-ranch-specs/http/type/property/additional-properties/main.tsp
```diff
@@ -231,35 +256,306 @@ model ModelForRecord {
 
 #suppress "@azure-tools/typespec-azure-core/bad-record-type" "For testing"
 @doc("The model extends from Record<ModelForRecord> type.")
-model ExtendsModelAdditionalProperties extends Record<ModelForRecord> {}
+model ExtendsModelAdditionalProperties extends Record<ModelForRecord> {
+  knownProp: ModelForRecord;
+}
```
haolingdong-msft, line 261: why do we add the property?

qiaozha: because previous extends and is Record way only works when the known property type is the same as the additional properties type.

haolingdong-msft (approved):

PR #558, opened 2024-04-15
typespec-bump-dep, add typespec-autorest

weidongxu-microsoft:
typespec-azure-resource-manager has dependency on typespec-autorest
https://github.com/Azure/typespec-azure/blob/be1e2c332a991ed935dcd51db40bdefc7f83d86e/packages/typespec-azure-resource-manager/package.json#L59

So, if we include typespec-azure-resource-manager and use dev version of it, we also need to override for typespec-autorest


weidongxu-microsoft:
If we have https://github.com/Azure/typespec-azure/issues/697, we won't need this PR anymore.

tadelesh (approved):

msyyc (approved):

PR #559, opened 2024-04-19
add test for encode on union property

chunyu3:


packages/cadl-ranch-specs/http/encode/union/main.tsp
```diff
@@ -0,0 +1,36 @@
...
+
+using TypeSpec.Http;
+using Azure.ClientGenerator.Core;
+
+@doc("Test for encode decorator on union property.")
+@supportedBy("dpg")
+@scenarioService("/encode/union")
+namespace Encode.Union;
+
+model UnixTimestampDatetimeUnionProperty {
```
weidongxu-microsoft, line 13: I don't think this be a union property (as I don't think this is still valid if type be `utcDateTime | string`). Seems better to call it nullable property.

UnixTimestampDatetimeNullableProperty ?

weidongxu-microsoft (approved):

packages/cadl-ranch-specs/http/encode/union/main.tsp
weidongxu-microsoft: Should it just be moved into https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/encode/datetime/main.tsp ?

tadelesh: agree

timotheeguerin:
doing some cleanup as we are migrating scenarios to typespec and typespec-azure repo, please repopen if it is still relevant. If it is please make equivalent change in microsoft/typespec

PR #567, opened 2024-04-26
specs, add explicit body to operation response, to keep behavior of 0.56.0-dev same as 0.55.0

weidongxu-microsoft:
behavior change issue https://github.com/microsoft/typespec/issues/3225


packages/cadl-ranch-specs/http/type/model/visibility/main.tsp
```diff
@@ -48,7 +48,10 @@ Expected response body:
 ```
 """)
 @get
-op getModel(@body input: VisibilityModel): VisibilityModel;
+op getModel(@body input: VisibilityModel): {
+  @body
```
iscai-msft, line 52: what is the change you're seeing without this change?

tadelesh: for tsp 0.56.0, it will do visibility filtering for body if not explicitly `@body`.

tadelesh (approved):

packages/cadl-ranch-specs/http/type/model/visibility/main.tsp
weidongxu-microsoft: > what is the change you're seeing without this change?

There is behavior change in compiler (or http/rest) lib for the response model. 0.55 would give `VisibilityModel`, but 0.56 would give an anonymous model with non-"read" properties filtered out.
https://github.com/microsoft/typespec/issues/3225

iscai-msft (approved):

PR #568, opened 2024-04-26
add test for versioning

tadelesh:

fix: https://github.com/Azure/cadl-ranch/issues/564

current only test for latest version.


packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -6962,3 +6962,117 @@ Expected request to send body:
...
+This operation should be generated with latest version's signature.
+
+Expected request body:
+
+```json
+{ "prop": "foo", "changedProp": "bar" }
+
+Expected query param:
+param="baz"
+```
```
weidongxu-microsoft, line 7078: should not included in json block

packages/cadl-ranch-specs/http/versioning/madeOptional/main.tsp
```diff
@@ -0,0 +1,64 @@
...
+  "{endpoint}/versioning/made-optional/api-version:{version}",
+  "Testserver endpoint",
+  {
+    /**
+     * Need to be set as 'http://localhost:3000' in client.
+     */
+    endpoint: url,
+
+    /**
+     * Need to be set as 'v1' or 'v2' in client.
```
weidongxu-microsoft, line 23: Any of the test require we set `v1` in client?

packages/cadl-ranch-specs/http/versioning/added/main.tsp
```diff
@@ -0,0 +1,135 @@
...
+header-v2=bar
+
+""")
+@route("/v1")
+@post
+op v1(@body body: ModelV1, @added(Versions.v2) @header headerV2: string): ModelV1;
+
+@scenario
+@scenarioDoc("""
+This operation should only be generated with latest version.
```
weidongxu-microsoft, line 107: nit,
I would want to say the scenario doc better capture which model/enum would be generated, as this is more a "check the code" test, instead of "call the API" test.

But it could be verbose as there is lots of `added`.

packages/cadl-ranch-specs/http/versioning/madeOptional/main.tsp
tadelesh: nope. i just want to add test for latest signature after versioning projection for now.

weidongxu-microsoft (approved):
If possible test it in python before merge.

packages/cadl-ranch-specs/http/versioning/added/main.tsp
```diff
@@ -0,0 +1,135 @@
...
+   * The version v2.
+   */
+  v2: "v2",
+}
+
+model ModelV1 {
+  prop: string;
+  enumProp: EnumV1;
+
+  @added(Versions.v2)
```
MaryGao, line 49: Do we consider more complex cases e.g @removed in v1, @added back in v2?

packages/cadl-ranch-specs/http/versioning/added/main.tsp
```diff
@@ -0,0 +1,135 @@
...
+}
+
+enum EnumV1 {
+  enumMemberV1,
+
+  @added(Versions.v2)
+  enumMemberV2,
+}
+
+@added(Versions.v2)
```
MaryGao, line 60: Could we also add cases for query/header/path param?

iscai-msft (approved):

PR #587, opened 2024-05-24
specs, model used in read-only property should be output

weidongxu-microsoft:
real service
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/ai/HealthInsights/HealthInsights.RadiologyInsights/model.radiologyinsights.tsp#L61-L73

`RadiologyInsightsJob` used in https://github.com/Azure/azure-rest-api-specs/blob/main/specification/ai/HealthInsights/HealthInsights.RadiologyInsights/route.radiologyinsights.tsp#L25, and hence input+output

`RadiologyInsightsInferenceResult` should be usage=output, as the property is read-only

### DIFF
--- a/packages/cadl-ranch-specs/http/azure/client-generator-core/usage/main.tsp
+++ b/packages/cadl-ranch-specs/http/azure/client-generator-core/usage/main.tsp
@@ -55,6 +55,37 @@ namespace ModelInOperation {
   @route("/outputToInputOutput")
   @get
   op outputToInputOutput(): OutputModel;
+
+  model ResultModel {
+    name: string;
+  }
+
+  model RoundTripModel {
+    @visibility("read")
+    result: ResultModel;
+  }
+
+  @doc("""
+  "ResultModel" should be usage=output, as it is read-only and does not exist in request body.
+
+  Expected body parameter: 
+  ```json
+  {
+  }
+  ```
+
+  Expected response body: 
+  ```json
+  {
+    "name": <any string>
+  }
+  ```
+  """)
+  @route("/modelInReadOnlyProperty")
+  @put
+  op modelInReadOnlyProperty(@body body: RoundTripModel): {
+    @body body: RoundTripModel;
+  };
 }
 
 @doc("Not used anywhere, but access is override to public so still need to be generated and exported with serialization.")
--- a/.changeset/wild-boats-float.md
+++ b/.changeset/wild-boats-float.md
@@ -1,0 +1,5 @@
+---
+"@azure-tools/cadl-ranch-specs": patch
+---
+
+Add usage test case for model used in read-only property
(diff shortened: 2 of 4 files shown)


haolingdong-msft (approved):

packages/cadl-ranch-specs/http/azure/client-generator-core/usage/main.tsp
```diff
@@ -55,6 +55,37 @@ namespace ModelInOperation {
...
+  Expected response body: 
+  ```json
+  {
+    "name": <any string>
+  }
+  ```
+  """)
+  @route("/modelInReadOnlyProperty")
+  @put
+  op modelInReadOnlyProperty(@body body: RoundTripModel): {
```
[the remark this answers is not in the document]
weidongxu-microsoft: It would, if the type of `readProp` is a model.

For now, it is a string, and there is no input/output for a system String.

packages/cadl-ranch-specs/http/azure/client-generator-core/usage/main.tsp
```diff
@@ -55,6 +55,37 @@ namespace ModelInOperation {
   @route("/outputToInputOutput")
   @get
   op outputToInputOutput(): OutputModel;
+
+  model ResultModel {
+    name: string;
+  }
```
weidongxu-microsoft, line 61: Java should generate the class without constructor and setter.

```java
@Immutable
public final class ResultModel implements JsonSerializable<ResultModel> {
    /*
     * The name property.
     */
    @Generated
    private final String name;

    /**
     * Creates an instance of ResultModel class.
     * 
     * @param name the name value to set.
     */
    @Generated
    private ResultModel(String name) {
        this.name = name;
    }

    /**
     * Get the name property: The name property.
     * 
     * @return the name value.
     */
    @Generated
    public String getName() {
        return this.name;
    }
```

qiaozha: I see, should we also consider other visibilities  on model type if that's the case? maybe no need? as we are not really support those visibilities for now?

weidongxu-microsoft: One case that occurs to me is the opposite

```ts
  @visibility("create", "update")
  prop: ModelOfInputUsage;
```

that `ModelOfInputUsage` has usage=input.
But at present, Java would produce same class of input and input+output.

tadelesh (approved):

packages/cadl-ranch-specs/http/azure/client-generator-core/usage/main.tsp
```diff
@@ -55,6 +55,37 @@ namespace ModelInOperation {
...
+  Expected body parameter: 
+  ```json
+  {
+  }
+  ```
+
+  Expected response body: 
+  ```json
+  {
+    "name": <any string>
```
qiaozha, line 80: not sure why the result property is being flattened? should it be like ?
```
  {
     "result": {"name": <any string>}
  }
```

PR #591, opened 2024-06-04
Add nullable-list test

mcgallan:
It is necessary to remove the content in CadlRanchMockapi in autorest.charp, as well as its related tests and code. This PR includes some tests from the [original Model-Typespec](https://github.com/Azure/autorest.csharp/blob/feature/v3/test/CadlRanchMockApis/src/Models-TypeSpec.ts). I have moved this part of the content to http/type/property/nullable, for specific changes, please refer to the .changeset folder.


packages/cadl-ranch-specs/http/type/property/nullable/main.tsp
```diff
@@ -147,3 +147,17 @@ model CollectionsModelProperty is ModelTemplate<InnerModel[]>;
...
+@route("/collections/int32")
+@operationGroup
+interface CollectionsInt32 extends OperationsTemplate<CollectionsInt32Property, "[123, 456]"> {}
+
+// Model with nullable collection string property
+@doc("Model with collection string properties")
+model CollectionsStringProperty is ModelTemplate<string[]>;
+@route("/collections/string")
+@operationGroup
+interface CollectionsString extends OperationsTemplate<CollectionsStringProperty, "['hello', 'world']"> {}
```
weidongxu-microsoft, line 163: The test here is whether user need to send an array or no array (null or absent). Why make 2 test case about content of the array? It should not matter at all.

mcgallan: fixed

weidongxu-microsoft: What I mean is that a test with `string[] | null` should already do. No need for anther `int32[] | null` (if you have `int32` then why not also `float32`? I think it is useless repetitions)

mcgallan: fixed

tadelesh (approved):

PR #425, opened 2023-09-22
use-endpoint-instead-of-localhost

qiaozha:
# Cadl Ranch Contribution Checklist:

- [ ] I have written a [scenario spec](../docs/writing-scenario-spec.md)
- [ ] I have **meaningful** `@scenario` names. Someone can look at the list of scenarios and understand what I'm covering.
- [ ] I have written a [mock API](../docs/writing-mock-apis.md)
- [ ] I have used `@scenarioDoc`s for extra scenario description and to tell people **how to pass** my mock api check.

PR #598, opened 2024-06-13
fix(decorator): replace hard code url with `{endpoint}` in `@scenarioService`

### LINKED ISSUE #597: bug(decorator): hard coded server url in `@scenarioService`
## Problem

See the codes: https://github.com/Azure/cadl-ranch/blob/2fb006d5b4609647138382bf5d248a2edbde7742/packages/cadl-ranch-expect/src/decorators.ts#L241

`@scenarioService` will set `@server("http://localhost:3000", "TestServer endpoint")`. We should change the literal value to a parameter, like https://github.com/Azure/cadl-ranch/pull/425

archerzz:
Hard coded value should be replaced by a parameter, like #425

fix #597


weidongxu-microsoft:
Any other language need this?

For Java, this change means we need to update all the client (probably on the handreds) to set localhost:3000

archerzz:
> Any other language need this?
> 
> For Java, this change means we need to update all the client in test (probably on the hundreds) to set endpoint=localhost:3000

I believe dotnet is not using `localhost:3000`: https://github.com/Azure/autorest.csharp/blob/9790814dd86eb1279e5c98fd123e4bcea7456553/test/CadlRanchProjects.Tests/type-enum-fixed.cs#L14-L18

But dotnet allows users to override the host, even if it's explicitly specified in spec. Things could be different in other languages. We can hold this PR for further discussion.

pshao25 (commented):
LGTM, can we make its default value as "http://localhost:3000" so that it won't break others.

qiaozha:
> LGTM, can we make its default value as "http://localhost:3000" so that it won't break others.

I don't think we can, because currently there's no way to set client side default value in typespec.

tadelesh (approved):

tadelesh:
> > LGTM, can we make its default value as "http://localhost:3000" so that it won't break others.
> 
> I don't think we can, because currently there's no way to set client side default value in typespec.

for endpoint, tcgc will treat service default as client default. this is the only exception.

weidongxu-microsoft:
> for endpoint, tcgc will treat service default as client default. this is the only exception. mpg template also has such pattern. @archerzz could you help to refine it?

I see :-)

Yeah, if we all take the service default of "host" as client default, I am good with this PR, as long as it have localhost as default of the endpoint.

archerzz:
Please hold this PR. After discussion, we decided to inject the following:
```typescript
@server(
  "{endpoint}",
  "Test server endpoint",
  {
    endpoint: string = "http://localhost:3000",
  }
)
```

---

blocked by https://github.com/Azure/typespec-azure/issues/1030

packages/cadl-ranch-expect/src/decorators.ts
```diff
@@ -238,6 +238,29 @@ export function $scenarioService(
...
+      type: {
+        kind: "Scalar",
+        name: "string",
+        derivedScalars: [],
+        projections: [],
+        projectionsByName: [],
+        isFinished: false,
+        decorators: [],
+        namespace: { kind: "Namespace", name: "TypeSpec", namespace: context.program.getGlobalNamespaceType() },
+      },
```
weidongxu-microsoft, line 255: I would suggest use the TypeSpec.string directly (instead of a mock one)
```
type: context.program.getGlobalNamespaceType().namespaces.get("TypeSpec").scalars.get("string"),
```

packages/cadl-ranch-expect/src/decorators.ts
```diff
@@ -238,6 +238,29 @@ export function $scenarioService(
...
+      default: { kind: "String", value: "http://localhost:3000" },
```
weidongxu-microsoft, line 256: `defaultValue` please. `default` was deprecated long time ago.

packages/cadl-ranch-expect/src/decorators.ts
```diff
@@ -238,6 +238,20 @@ export function $scenarioService(
...
     projectionsByName: [],
   } as any);
-  context.call($server, target, "http://localhost:3000", "TestServer endpoint");
+
+  context.call($server, target, "{endpoint}", "Test server endpoint", {
+    kind: "Model",
+    properties: new Map().set("endpoint", {
+      name: "endpoint",
+      type: context.program!.getGlobalNamespaceType().namespaces.get("TypeSpec")!.scalars.get("string"),
+      defaultValue: { valueKind: "StringValue", value: "http://localhost:3000" },
```
archerzz, line 247: `defaultValue` is a `Value` per definition: https://github.com/microsoft/typespec/blob/60d87bc77014a6756b255f3e81f931ea8b25e80f/packages/compiler/src/core/types.ts#L352

Here we use `StringValue`: https://github.com/microsoft/typespec/blob/60d87bc77014a6756b255f3e81f931ea8b25e80f/packages/compiler/src/core/types.ts#L411-L415

archerzz:
Note, this is blocked by https://github.com/Azure/typespec-azure/issues/1030

tadelesh (approved):

packages/cadl-ranch-expect/src/decorators.ts
weidongxu-microsoft: This may need a `entityKind: "Value"` property.

tadelesh:
@archerzz i believe you could merge this pr. current tcgc will do the same thing.

PR #599, opened 2024-06-13
Disable patch convenience method for csharp

pshao25:
To adopt the usage of TCGC, we have some gap from the unsupported patch convenience method. So disable for now to unblock TCGC integration.


weidongxu-microsoft (approved):

MaryGao (approved):
LGTM since only client.tsp change for .Net

PR #600, opened 2024-06-17
Add api version test

mcgallan:
Move the content related to [ApiVersion-TypeSpec](https://github.com/Azure/autorest.csharp/tree/feature/v3/test/TestProjects/ApiVersion-TypeSpec) from TestProject in AutoRest.CSharp to the http/versioning/anomaly folder.


weidongxu-microsoft (commented):
Should it be here https://github.com/Azure/cadl-ranch/tree/94f48ba4fb638b0b78b00572d1b837d05f9a624f/packages/cadl-ranch-specs/http/server/versions ?

And what this Anomaly mean? I don't see anything unexpected from the test.

PR #603, opened 2024-06-18
specs, explicit body to avoid spread

weidongxu-microsoft:
To avoid these model getting spread, after the "spread is spread" feature.

We should already had `@body` and `@bodyRoot` in test, for explicit body (non-spread).
I will add a `@bodyIgnore` case in a follow-up PR.

### DIFF
--- a/packages/cadl-ranch-specs/http/azure/core/lro/rpc/main.tsp
+++ b/packages/cadl-ranch-specs/http/azure/core/lro/rpc/main.tsp
@@ -96,9 +96,14 @@ Expected response body:
 """)
 @route("/generations:submit")
 op longRunningRpc is global.Azure.Core.LongRunningRpcOperation<
-  {
-    @bodyRoot body: GenerationOptions;
-  },
+  BodyParameter<GenerationOptions>,
   GenerationResponse,
   GenerationResult
 >;
+
+alias BodyParameter<T, TName extends valueof string = "body", TDoc extends valueof string = ""> = {
+  @doc(TDoc)
+  @friendlyName(TName)
+  @bodyRoot
+  body: T;
+};
--- a/packages/cadl-ranch-specs/http/azure/core/lro/rpc/main.tsp
+++ b/packages/cadl-ranch-specs/http/azure/core/lro/rpc/main.tsp
@@ -95,4 +95,10 @@ Expected response body:
 ```
 """)
 @route("/generations:submit")
-op longRunningRpc is global.Azure.Core.LongRunningRpcOperation<GenerationOptions, GenerationResponse, GenerationResult>;
+op longRunningRpc is global.Azure.Core.LongRunningRpcOperation<
+  {
+    @body body: GenerationOptions;
+  },
+  GenerationResponse,
+  GenerationResult
+>;
--- a/packages/cadl-ranch-specs/http/azure/core/lro/rpc/main.tsp
+++ b/packages/cadl-ranch-specs/http/azure/core/lro/rpc/main.tsp
@@ -97,7 +97,7 @@ Expected response body:
 @route("/generations:submit")
 op longRunningRpc is global.Azure.Core.LongRunningRpcOperation<
   {
-    @body body: GenerationOptions;
+    @bodyRoot body: GenerationOptions;
   },
   GenerationResponse,
   GenerationResult
--- a/packages/cadl-ranch-specs/http/azure/core/lro/rpc/main.tsp
+++ b/packages/cadl-ranch-specs/http/azure/core/lro/rpc/main.tsp
@@ -101,7 +101,7 @@ op longRunningRpc is global.Azure.Core.LongRunningRpcOperation<
   GenerationResult
 >;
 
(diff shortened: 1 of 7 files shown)


packages/cadl-ranch-specs/http/azure/core/lro/rpc/main.tsp
```diff
@@ -95,4 +95,10 @@ Expected response body:
 ```
 """)
 @route("/generations:submit")
-op longRunningRpc is global.Azure.Core.LongRunningRpcOperation<GenerationOptions, GenerationResponse, GenerationResult>;
+op longRunningRpc is global.Azure.Core.LongRunningRpcOperation<
+  {
+    @body body: GenerationOptions;
```
weidongxu-microsoft, line 100: Any ask on the naming of the request body property?

As `body`, there will be a name change in Java (but it does not break API)
```diff
-    public PollerFlux<PollOperationDetails, GenerationResult> beginLongRunningRpc(GenerationOptions generationOptions) {
+    public PollerFlux<PollOperationDetails, GenerationResult> beginLongRunningRpc(GenerationOptions body) {
```

weidongxu-microsoft: Here the model is all body properties, so `@body` and `@bodyRoot` should be equivalent.

tadelesh: in tcgc logic, if the body model has name, then use that name, else use `body`.

tadelesh:
please do not merge after we have the final decision to do this approach of changing spec.

packages/cadl-ranch-specs/http/azure/core/lro/rpc/main.tsp
weidongxu-microsoft: Note that Java prefer e.g. `requestContext` instead of `body`.

packages/cadl-ranch-specs/http/parameters/spread/main.tsp
```diff
@@ -22,10 +22,10 @@ namespace Model {
   @scenarioDoc("""
   Test case for spread named model. 
   
-  Should generate request body model named `BodyParameter`.
+  Should not generate request body model named `BodyParameter`.
   Should generate an operation like below:
   ```
-  spreadAsRequestBody(bodyParameter: BodyParameter)
+  spreadAsRequestBody(name: string)
```
weidongxu-microsoft, line 28: I've changed the scenarioDoc, as now `op spreadAsRequestBody(...BodyParameter): NoContentResponse;` would be a spread.

msyyc (approved):
LGTM

tadelesh (commented):
do you think we need to follow the patten from https://github.com/Azure/azure-rest-api-specs/pull/29483/files, use the template as follows?
```
alias BodyParameter<
  T,
  TName extends valueof string = "body",
  TDoc extends valueof string = ""
> = {
  @doc(TDoc)
  @friendlyName(TName)
  @bodyRoot
  body: T;
};
```

weidongxu-microsoft:
> do you think we need to follow the patten from https://github.com/Azure/azure-rest-api-specs/pull/29483/files, use the template as follows?
> 
> ```
> alias BodyParameter<
>   T,
>   TName extends valueof string = "body",
>   TDoc extends valueof string = ""
> > = {
>   @doc(TDoc)
>   @friendlyName(TName)
>   @bodyRoot
>   body: T;
> };
> ```

It is test and most of them is valid for generic/unbranded. So far, I do not expect unbranded would write this `BodyParameter`, therefore it seems no reason to write this in cadl-ranch.

For the few "azure" tests, I would use that when `BodyParameter` gets to Azure.Core.

packages/cadl-ranch-specs/http/azure/core/lro/rpc/main.tsp
```diff
@@ -95,4 +95,15 @@ Expected response body:
...
+  GenerationResponse,
+  GenerationResult
+>;
+
+alias BodyParameter<T, TName extends valueof string = "body", TDoc extends valueof string = ""> = {
+  @doc(TDoc)
+  @friendlyName(TName)
+  @bodyRoot
+  body: T;
+};
```
weidongxu-microsoft, line 109: switch to use local `BodyParameter` in azure.

will switch to `Azure.Core.BodyParameter` in future.

tadelesh (approved):

packages/cadl-ranch-specs/http/azure/core/lro/rpc/main.tsp
```diff
@@ -95,4 +95,15 @@ Expected response body:
...
+  GenerationResponse,
+  GenerationResult
+>;
+
+alias BodyParameter<T, TName extends valueof string = "body", TDoc extends valueof string = "The body parameter."> = {
+  @doc(TDoc)
+  @friendlyName(TName)
+  @bodyRoot
+  body: T;
+};
```
weidongxu-microsoft, line 109: `@doc` default as `TDoc`

PR #612, opened 2024-06-28
rename arm tests to resourcemanager

pshao25:
Arm is in the `.gitignore` of autorest.csharp repo. This is legacy code so that we don't want to touch it. Therefore we want to rename it to resourcemanager.

### DIFF
--- a/packages/cadl-ranch-specs/http/azure/resource-manager/models/resources/mockapi.ts
+++ b/packages/cadl-ranch-specs/http/azure/resource-manager/models/resources/mockapi.ts
@@ -6,9 +6,9 @@ export const Scenarios: Record<string, ScenarioMockApi> = {};
 const SUBSCRIPTION_ID_EXPECTED = "00000000-0000-0000-0000-000000000000";
 const RESOURCE_GROUP_EXPECTED = "test-rg";
 const validTopLevelResource = {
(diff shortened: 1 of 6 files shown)


packages/cadl-ranch-specs/http/azure/resource-manager/models/resources/mockapi.ts
```diff
@@ -84,6 +84,7 @@ Scenarios.Azure_Arm_Models_Resources_TopLevelTrackedResources_createOrReplace =
         properties: {
           description: "valid",
         },
+        tags: {}, // Workaround here because of https://github.com/Azure/autorest.csharp/issues/4876
       });
```
pshao25, line 88: I can remove this workaround if it impacts others.

weidongxu-microsoft (approved):
Next time, please review the namespace when it is still PR.

It takes time after dev already added the tests.

packages/cadl-ranch-specs/http/azure/resource-manager/models/resources/mockapi.ts
pshao25: A second thought, removed this.

msyyc (approved):
Python feel OK since we have not tested these cases yet.

ArthurMa1978 (approved):

lirenhe (approved):

Issue #548, opened 2024-04-05
Multipart Form Data Scenario tests


lmazuel:
See https://github.com/microsoft/typespec/issues/3046 for details

msyyc:
depends on https://github.com/Azure/typespec-azure/issues/960

msyyc:
It is a little hard to catch up with June so move to July.

Issue #570, opened 2024-04-28
MFD `jsonArrayParts` case is not aligned with typespec definition


MaryGao:
In typespec [def](https://github.com/timotheeguerin/typespec/blob/5a9ffa2434f5c1add5a9e373d06f438614c6004e/docs/libraries/http/multipart.md#examples) for `addresses` it should have N parts for N addresses, not 1 part for an array of addresess.

```
// Send multiple fields - In this scenario each address is sent as an individual part
model Address {
  street: string;
  city: string;
}
op create(@header contentType: "multipart/form-data", addresses: Address[]): void;
```
So the current case implementation is not aligned with typespec definition
https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/payload/multipart/main.tsp#L231-L242

msyyc:
There is already typespec design for this case https://github.com/microsoft/typespec/issues/3046 however it is not implemented yet. And for now there is no service for this scenario, so we will not update the case until there is new requirement. If language emitter has not passed the case, please hold it on.

lmazuel:
@MaryGao is right that the test is not implemented as it should. If should be N part of one object:

```
model Something {
   id: Obj1 // one part, one obj
   adress[]: // N parts, one obj
   adress[][]: // N parts, array of object
}
```

This is the default semantic. For more advanced scenario, we need #3046, but the default behavior is as described here.

PR #613, opened 2024-06-28
[Multipart] add test case with `HttpPart`

msyyc:
https://github.com/Azure/cadl-ranch/issues/548

Most languages depends on TCGC so this PR is blocked by https://github.com/Azure/typespec-azure/pull/1090


packages/cadl-ranch-specs/http/payload/multipart/main.tsp
timotheeguerin, line 1: I think all the existing test should be converted to http part as well. Complex one is just a smoke test and is probable the least good to keep

packages/cadl-ranch-specs/http/payload/multipart/mockapi.ts
```diff
@@ -164,3 +164,11 @@ Scenarios.Payload_MultiPart_FormData_checkFileNameAndContentType = passOnSuccess
 Scenarios.Payload_MultiPart_FormData_anonymousModel = passOnSuccess(
   createMockApis("anonymous-model", [checkProfileImage]),
 );
+
+Scenarios.Payload_MultiPart_FormData_checkFileNameAndContentTypeWithHttpPart = passOnSuccess(
```
timotheeguerin, line 168: That also be better as 2 separate test and ideally more
Required file name 
Optional filename 
Required content type
Optional content type

timotheeguerin: Specific content type

packages/cadl-ranch-specs/http/payload/multipart/main.tsp
msyyc: Laurent said we shall not remove support format of `@body` for multipart and Python/Java/JS already covers these existing scenarios so I think we shall keep the existing cases unchanged and then add new case with `HttpPart`

packages/cadl-ranch-specs/http/payload/multipart/mockapi.ts
timotheeguerin: my point for the 1st one is that you have `complex` as a test which is bad(it is not representative of any feature if you look at the dashboard), this should be split into the individual feature it is testing

msyyc: I see. But I think it is also reasonable for complex scenarios that combine all single scenarios since sometimes `1+1` is much much more complicated than `1`.
I will add separate cases for  `Required file name/Required content type/Specific content type`

timotheeguerin: I think it’s fair to have one multiple scenario to ensure multiple parts work together as well but it shouldn’t be used as the catch all

msyyc: Updated: 
- fileWithHttpPartSpecificContentType for specific content type
- fileWithHttpPartRequiredContentType for required content type
- fileWithHttpPartOptionalContentType for optional content type
- each of up cases cover required filename
- complexWithHttpPart combines up scenarios

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -2567,29 +2567,36 @@ Content-Type: application/octet-stream
 --abcde12345--
 ```
 
-### Payload_MultiPart_FormData_jsonArrayParts
+### Payload_MultiPart_FormData_complexWithHttpPart
```
msyyc, line 2570: Remove legacy test case `jsonArrayParts` since it doesn't follow new guideline of https://github.com/Azure/cadl-ranch/issues/570#issuecomment-2089011502

packages/cadl-ranch-specs/http/payload/multipart/main.tsp
```diff
@@ -332,4 +329,132 @@ Content-Type: application/octet-stream;
...
+[{
+  "city": "Y"
+},{
+  "city": "Z"
+}]
+--abcde12345
+Content-Disposition: form-data; name="pictures"; filename="<any-name-is-ok>"
+Content-Type: application/octet-stream
+
+{…file content…}
```
weidongxu-microsoft, line 444: Preferrable to also specify e.g. Jpeg file

packages/cadl-ranch-specs/http/payload/multipart/main.tsp
```diff
@@ -332,4 +329,132 @@ Content-Type: application/octet-stream;
...
+--abcde12345
+Content-Disposition: form-data; name="pictures"; filename="<any-name-is-ok>"
+Content-Type: application/octet-stream
+
+{…file content…}
```
weidongxu-microsoft, line 449: Preferrable to also specify e.g. 2 Png files

packages/cadl-ranch-specs/http/payload/multipart/main.tsp
```diff
@@ -16,11 +16,44 @@ model Address {
...
+model FileWithHttpPartOptionalContentTypeRequest {
+  profileImage: HttpPart<FileOptionalContentType>;
+}
+
+model ComplexHttpPartsModelRequest {
+  id: HttpPart<string>;
+  address: HttpPart<Address>;
+  profileImage: HttpPart<FileRequiredMetaData>;
+  previousAddresses: HttpPart<Address[]>;
+  pictures: HttpPart<FileRequiredMetaData>[];
```
weidongxu-microsoft, line 50: nit, may use a slightly different File model e.g. `FileOptionalContentType`, for some diversity in this test.

packages/cadl-ranch-specs/http/payload/multipart/main.tsp

packages/cadl-ranch-specs/http/payload/multipart/main.tsp
msyyc: updated

packages/cadl-ranch-specs/http/payload/multipart/main.tsp
msyyc: Optional content type is already covered by case `fileWithHttpPartOptionalContentType`

weidongxu-microsoft (approved):

haolingdong-msft (approved):

tadelesh (approved):

PR #619, opened 2024-07-02
Use unambiguous float values in additional-properties test

antkmsft:
Updated expected float value of 2.71828 to 2.71875, a value that can be exactly represented as IEEE754 (http/type/property/additional-properties).

Same rationale as for #611.

More detailed explanation: the spec says the type is `float32`: https://github.com/Azure/cadl-ranch/blob/56f07cf86f006cba5c2e3024eae8581721791be4/packages/cadl-ranch-specs/http/type/property/additional-properties/main.tsp#L73

In terms of 32-bit floating points, `2.71828` and `2.7182800769805908203125` are the same thing (second one is what you get after trying to represent 2.71828 as 32-bit IEEE754). C++ codegen uses 32-bit foats for `float32`. Not `float64`. So, C++ codegen does serialize `2.7182800769805908203125`, because it is what it sees. Now, the problem is that mockapi.ts is TypeScript, so it reads values as `numeric`. Practically, that means that it will deserialize the number it receives as 64-bit floating point number. So when it receives `2.7182800769805908203125` it can be reprsented as 64-bit floating point number, but when mockapi tries to compare it with expected `2.71828` value, that value was never converted to 32 bits. And as far as 64-bit representation for `2.71828`, it is `2.71828000000000002955857780762016773223876953125`. So, the test sees that `2.7182800769805908203125 !== 2.71828000000000002955857780762016773223876953125`, and fails.
I am updating the expected value to `2.71875`. `2.71875` can be exactly represented as 32-bit floating point number (which means, it also can be represented as exactly the same 64-bit floating number). And the comparison of `2.71875 === 2.71875` will succeed, no matter if different floating point precisions were used on client and server, or the same precision.


weidongxu-microsoft (commented):
@qiaozha Is the change OK for you?

qiaozha (approved):
LGTM, but I have seen this kind of changes many times, how can we prevent introduce this again?

weidongxu-microsoft (approved):

weidongxu-microsoft:
I cannot think of an approach to lint/verify. The numeric is in ts file.

antkmsft:
I've been fixing as much of them as I know of and can find. The problem is, such constants are not that much searchable.
As I am going through implementation, I am fixing the ones that I encounter, and eventually this way I should encounter them all.
I mean, to my knowledge, we have fixed all of them, but I've been thinking the same after a previous PR, and before that - there's been 3 or 4 similar PRs from me :)

As for prevention: this could be a good tool to test floating point constants: https://www.exploringbinary.com/floating-point-converter/ : if you type `2.71828`, there, and check both "single" and "double" precision, it will give you both representations: `2.7182800769805908203125` and `2.71828000000000002955857780762016773223876953125`.

Speaking of choosing a new value, it could be helpful to understand how the fractional part should be a number that can eventually be represented as sum of `1/(2^n)`s: the same way that `int`s are a sum of `2^n`s, IEEE754 floating point numbers continue that axis to the other side, below 1, and go into fractions; and the "exponent" part is basically another int number which tells what's the `n` of the leftmost bit - that is what makes them "floating". To better understand the internal composition of floats, you can play with https://www.h-schmidt.net/FloatConverter/IEEE754.html.

PR #624, opened 2024-07-04
Add ReadOnly Properties

mcgallan:
It is necessary to remove the content in CadlRanchMockapi in autorest.charp, as well as its related tests and code. This PR includes some tests from the [original Model-Typespec](https://github.com/Azure/autorest.csharp/blob/feature/v3/test/CadlRanchMockApis/src/Models-TypeSpec.ts). I have moved this part of the content to http/type/model/visibility, and added tests for plaindate and plaintime in type/property/optionality. for specific changes, please refer to the .changeset folder.

### DIFF
--- a/packages/cadl-ranch-specs/http/type/model/visibility/main.tsp
+++ b/packages/cadl-ranch-specs/http/type/model/visibility/main.tsp
@@ -41,33 +41,9 @@ model ReadOnlyModel {
   @visibility("read")
   requiredNullableIntList: int32[] | null;
 
-  /** Optional readonly nullable int list. */
-  @visibility("read")
-  optionalNullableIntList?: int32[] | null;
-
-  /** Required readonly nullable model list. */
-  @visibility("read")
-  requiredNullableModelList: InnerModel[] | null;
-
-  /** Optional readonly nullable model list. */
-  @visibility("read")
-  optionalNullableModelList?: InnerModel[] | null;
-
   /** Required readonly string dictionary. */
   @visibility("read")
   requiredStringRecord: Record<string>;
-
-  /** Optional readonly string dictionary. */
-  @visibility("read")
-  optionalStringRecord?: Record<string>;
-
-  /** Required readonly model dictionary. */
-  @visibility("read")
-  requiredModelRecord: Record<InnerModel>;
-
-  /** Optional readonly model dictionary. */
-  @visibility("read")
-  optionalModelRecord?: Record<InnerModel>;
 }
 
 @scenario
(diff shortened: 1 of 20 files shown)


packages/cadl-ranch-specs/http/type/model/visibility/main.tsp
```diff
@@ -30,6 +30,22 @@ model VisibilityModel {
   deleteProp: boolean;
 }
 
+/** Innermodel used as a readonly property for testing. */
```
pshao25, line 33: Why we need innermodel

pshao25 (approved):
There are two parts in the PR:
1. A readonly model with complex type.
2. Test more types of a model property: plaindate and plaintime

packages/cadl-ranch-specs/http/type/model/visibility/main.tsp
mcgallan: fixed

weidongxu-microsoft (approved):

packages/cadl-ranch-specs/http/type/model/visibility/main.tsp
```diff
@@ -30,6 +30,17 @@ model VisibilityModel {
   deleteProp: boolean;
 }
 
+/** RoundTrip model with readonly properties. */
+model ReadOnlyModel {
+  /** Required readonly nullable int list. */
+  @visibility("read")
+  requiredNullableIntList: int32[] | null;
```
qiaozha, line 37: we will not be able to pass the case for Modular with the current design.

pshao25: which you cannot pass? you mean nullable?

msyyc (approved):
Looks good to Python

packages/cadl-ranch-specs/http/type/model/visibility/main.tsp
qiaozha: no, required readOnly case in input. is there a real case for it?

pshao25: It's roundtrip. But we can change it to optional.

qiaozha: That would be nice to us, otherwise we probably won't be able to support it in a significant long time.

packages/cadl-ranch-specs/http/type/model/visibility/main.tsp
```diff
@@ -30,6 +30,17 @@ model VisibilityModel {
   deleteProp: boolean;
 }
 
+/** RoundTrip model with readonly properties. */
+model ReadOnlyModel {
+  /** Required readonly nullable int list. */
+  @visibility("read")
+  requiredNullableIntList?: int32[] | null;
```
pshao25, line 37: Change its name as well

mcgallan: fixed

packages/cadl-ranch-specs/http/type/model/visibility/main.tsp
mcgallan: fixed

qiaozha (approved):

PR #625, opened 2024-07-08
Add Custom Operation Test in azure/core/basic

mcgallan:
The purpose of this PR is to migrate [the tests](https://github.com/Azure/autorest.csharp/blob/aee28deecf34b25e50cfdc950d572be6fb9aa6a8/test/TestProjects/Authoring-TypeSpec/Authoring-TypeSpec.tsp#L290) from the test.project folder in autorest.csharp to the cadl branch, specifically referring to the tests in latestVersion.typeSpec as detailed in the .changeset.
related issue: https://github.com/Azure/autorest.csharp/issues/4848


packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
```diff
@@ -70,6 +70,20 @@ model UserExportParams {
...
 
+model SwapUsersOptions {
+  @doc("Represents the first user id.")
+  firstUserId: int32;
+
+  @doc("Represents the second user id.")
+  secondUserId: int32;
+}
+
+model SwapUsersParams {
```
pshao25, line 81: How is this test related to "swap user"?

mcgallan: I have made improvements to this method, instead of simply calling it for the user. Regarding the swapParameter, I think it serves as a credential indicating which two models to swap, and it cannot be modified for the time being.

packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
```diff
@@ -214,3 +238,16 @@ Expected response body:
...
+@scenario
+@scenarioDoc("""
+Should only generate models named User and UserOrder.
+Expected query parameter: api-version=2022-12-01-preview
+""")
+@collectionAction(Dog, "swapdogs")
+op swapdogs is ResourceOperations.ResourceCollectionAction<
+  Dog,
+  SwapDogsParams,
+  global.Azure.Core.Foundations.AcceptedResponse
```
weidongxu-microsoft, line 252: Why use Foundations? There are not recommended to be used by Azure lib.

packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
```diff
@@ -214,3 +238,16 @@ Expected response body:
...
 op export is ResourceOperations.ResourceAction<User, UserExportParams, User>;
+
+/** Swaps two existing dogs with each other. */
+@scenario
+@scenarioDoc("""
+Should only generate models named User and UserOrder.
+Expected query parameter: api-version=2022-12-01-preview
+""")
+@collectionAction(Dog, "swapdogs")
+op swapdogs is ResourceOperations.ResourceCollectionAction<
```
weidongxu-microsoft, line 249: swapdogs...

Could we use test with more meaningful API?

packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
mcgallan: fixed

packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
mcgallan: fixed

packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
```diff
@@ -214,3 +216,41 @@ Expected response body:
...
+  },
+  {
+    "id": 2,
+    "name": "John",
+    "etag": "22bdc430-65e8-45ad-81d9-8ffa60d55b59"
+  }
+]
+""")
+@collectionAction(User, "exportallusers")
+op exportAllUsers is ResourceOperations.ResourceCollectionAction<
```
weidongxu-microsoft, line 249: Why do you need `@collectionAction(User, "exportallusers")`? I think it should be included in `ResourceCollectionAction` operation template.

See if you can add a `@post` above the operation. This would be redundant, but the summary seems messed-up without it https://github.com/Azure/cadl-ranch/pull/625/files#diff-e96e1cba09706c8a0e00bde73f721986543b99e3a0db21b3bbb98559db31185a

This is not required. But the summary seems to be

weidongxu-microsoft (commented):
@pshao25 for double check.

packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
```diff
@@ -214,3 +216,41 @@ Expected response body:
...
+    "id": 1,
+    "name": "Madge",
+    "etag": "11bdc430-65e8-45ad-81d9-8ffa60d55b59"
+  },
+  {
+    "id": 2,
+    "name": "John",
+    "etag": "22bdc430-65e8-45ad-81d9-8ffa60d55b59"
+  }
+]
```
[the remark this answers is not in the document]
mcgallan: fixed

packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
mcgallan: Based on the generated results, it is necessary to retain @collectionAction. After removing it, the generated path directly calls the operation name, resulting in uppercase letters appearing in the path. Therefore, I added @collectionAction to prevent this from happening. @post added.

packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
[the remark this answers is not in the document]
weidongxu-microsoft: It's not fixed. Now is it not even a valid json. Please do a simple verify.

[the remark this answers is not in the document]
mcgallan: The corrections have been made, including fixing the case issue with "users".

packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
weidongxu-microsoft: Eh, who says uppercase letters cannot appear in the path?

packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
```diff
@@ -214,3 +216,44 @@ Expected response body:
...
+      "name": "Madge",
+      "etag": "11bdc430-65e8-45ad-81d9-8ffa60d55b59"
+    },
+    {
+      "id": 2,
+      "name": "John",
+      "etag": "22bdc430-65e8-45ad-81d9-8ffa60d55b59"
+    }
+  ]
+}
```
weidongxu-microsoft, line 248: you miss the ending ```

mcgallan: fixed

packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
mcgallan: In a previous PR, there was a comment suggesting that I change the uppercase letters in the path to lowercase. Additionally, I have rarely seen paths with uppercase letters in mockapi. So, which approach is more appropriate?

weidongxu-microsoft: https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#http-url-casing

I don't have particular preference in this case. It is fine for all-lower-case. Just want to know why the extra `@collectionAction`.

weidongxu-microsoft (approved):
Please wait another approval (from non-Java dev).

ArcturusZhang (approved):

weidongxu-microsoft:
@mcgallan 

Please format the PR again.

weidongxu-microsoft:
Format still fails

PR #627, opened 2024-07-09
Add Test in Versioning/Removed

mcgallan:
The purpose of this PR is to migrate [the tests](https://github.com/Azure/autorest.csharp/blob/feature/v3/test/TestProjects/LatestVersion-TypeSpec/LatestVersion-TypeSpec.tsp) from the test.project folder in autorest.csharp to the cadl branch, specifically referring to the tests in latestVersion.typeSpec as detailed in the .changeset.
related issue: https://github.com/Azure/autorest.csharp/issues/4850
https://github.com/Azure/autorest.csharp/issues/4853
https://github.com/Azure/autorest.csharp/issues/4856
The content of these three projects is identical and can be handled together.


packages/cadl-ranch-specs/http/versioning/removed/main.tsp
```diff
@@ -123,3 +133,16 @@ interface InterfaceV1 {
   @route("/v1")
   v1InInterface(@body body: ModelV1): ModelV1;
 }
+
+/** This operation should not be generated with old version */
+@scenario
+@scenarioDoc("""
+
```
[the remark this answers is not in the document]
pshao25: We are asking vendors to help us move some tests in autorest.csharp to cadl-ranch. I'll let them add the descriptions.

BTW They might not that familiar with our product. I'll make sure there won't any obvious problem before you start your review. So I think to save your time, you could wait till my approval.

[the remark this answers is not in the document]
mcgallan: Added

packages/cadl-ranch-specs/http/versioning/removed/main.tsp
```diff
@@ -123,3 +133,16 @@ interface InterfaceV1 {
...
+Expected request body:
+```json
+{ "id": "123", "type": "foo" }
+```
+""")
+@post
+@route("/v3")
+op v3(@body body: ModelV3): ModelV3;
```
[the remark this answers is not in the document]
mcgallan: The operation name is to maintain consistency with the operation name above. As for this operation, it tests the scenario where Added and Removed are used in combination in situations where there are three or more versions. Applicable to scenarios like the following: a property is added in the first version, removed in the second version, and then added back in the third version.

archerzz (approved):
LGTM @mcgallan Make sure we can generate 3 test projects from this case, e.g. `api-version` should be `v1`, `beta`, `v2` respectively.

packages/cadl-ranch-specs/http/versioning/removed/main.tsp
```diff
@@ -123,3 +138,19 @@ interface InterfaceV1 {
...
+@scenario
+@scenarioDoc("""
+
+Expected request body:
+```json
+v1/V2:
+{ "id": "123", "type": "foo" }
+ beta:
+{ "id": "123" }
+```
```
tadelesh, line 152: this is not a json string

packages/cadl-ranch-specs/http/versioning/removed/main.tsp
```diff
@@ -66,6 +73,14 @@ model ModelV2 {
   unionProp: UnionV2;
 }
 
+model ModelV3 {
```
iscai-msft, line 76: Shouldn't `type` be added in `v1`, since we expect `v1` and `v2` to send a value for `type`, but not `v3`? If not, I don't really see how this test is different from `removed_v2`, since they're both testing if it's included for one version and removed in the next

packages/cadl-ranch-specs/http/versioning/removed/main.tsp
mcgallan: fixed

packages/cadl-ranch-specs/http/versioning/removed/main.tsp
mcgallan: If we didn't specifically handle it with `@removed(Versions.V1)`, isn't the type property present by default in version V1? 
 This operation tests the combination of `@removed` and `@added`. According to @archerzz 's suggestion, this type property is added in the first version, removed in the second version, and added again in the third version, which still differs from V2.

iscai-msft: You've explicitly added `@added(v2)`, which makes it not present in `v1`

mcgallan: > is it kind of a test of ABA problem?

I think so. If using @added() causes a property to not exist in another version, I'm not sure how to design a reasonable test case for testing this.

iscai-msft: Removing the `@added` should fix the issue, since by default everything is added in the first api version, unless you explicitly specify that it was added in a later version

mcgallan: Removed.

iscai-msft (approved):
I think this looks good to me, once we've moved to more ABA testing. @tadelesh can you confirm the new test makes sense with the existing scenarios? Thanks so much!

packages/cadl-ranch-specs/http/versioning/removed/main.tsp
```diff
@@ -123,3 +144,38 @@ interface InterfaceV1 {
...
+@scenario
+@scenarioDoc("""
+  Expected request body:
+  ```json
+  { "id": "123", "enumProp": "enumMemberV1" }
+  ```
+  """)
+@post
+@route("/v3inversion2")
+op modelV3InVersion2(@body body: ModelV3): ModelV3;
```
tadelesh, line 181: we should not add three op to test three version, instead, we should add three tests in the mock api.

tadelesh (changes requested):
i'm little bit confused of what this test is trying to achieve?

packages/cadl-ranch-specs/http/versioning/removed/main.tsp
mcgallan: In actual operations, the removed mockapi seems to recognize these tests more based on the expected body rather than the path. Therefore, for the BetaVersion mockapi, it is difficult to merge it with the other two V1 and V2 versions of the mockapi. The generated code is based on the V2 version, so in actual generation, if we remove a property in the Beta version, it will also be absent in the V2 version. Therefore, we can only handle it again using @added.

mcgallan: @tadelesh Is the current version of mockapi a suitable solution?

packages/cadl-ranch-specs/http/versioning/removed/main.tsp
```diff
@@ -32,12 +34,17 @@ namespace Versioning.Removed;
...
-   * The version v1.
+   * The original version v1.
    */
   v1: "v1",
 
   /**
-   * The version v2.
+   * The beta version.
+   */
+  beta: "beta",
```
iscai-msft, line 44: in our tests we generally use `preview` instead of `beta`. The preview is also attached with a specific version. Would this be `v3Preview`?

tadelesh (approved):

tadelesh:
please refine the version name

packages/cadl-ranch-specs/http/versioning/removed/main.tsp
mcgallan: The beta version is essentially a transitional version between V1 and V2, so renaming it to v2preview is more appropriate.  
 Fixed

PR #628, opened 2024-07-11
specs, add test scenario for JSON example used in Azure

weidongxu-microsoft:
Java sample code snippet would look like this
```java
        ActionResponse response = azureExampleClient.basicAction("query", "header",
            new ActionRequest("text")
                .setModelProperty(
                    new Model().setInt32Property(1).setFloat32Property(1.5D).setEnumProperty(Enum.ENUM_VALUE1))
                .setArrayProperty(Arrays.asList("item"))
                .setRecordProperty(mapOf("record", "value")));
```

Pass in Java https://github.com/weidongxu-microsoft/autorest.java/blob/cadl-ranch-example-json/typespec-tests/src/test/java/com/_specs_/azure/example/basic/BasicTests.java

Test likely need to set a `examples-directory` in emitter option, when generating code for this case.

### DIFF
--- a/packages/cadl-ranch-specs/http/azure/example/basic/client.tsp
+++ b/packages/cadl-ranch-specs/http/azure/example/basic/client.tsp
@@ -1,0 +1,50 @@
+import "@azure-tools/typespec-client-generator-core";
+
+import "./main.tsp";
+
+using Azure.ClientGenerator.Core;
+
+@TypeSpec.Versioning.useDependency(_Specs_.Azure.Example.Basic.Versions.v2022_12_01_preview)
+namespace Client;
+
+@client({
+  name: "AzureExampleClient",
+  service: _Specs_.Azure.Example.Basic,
+})
+interface AzureExampleClient {
+  @scenario
+  @scenarioDoc("""
+  Expected request and response is same as the JSON example at examples/2022-12-01-preview/basic.json
+
+  When generate the code, one need to set the "examples-directory" option.
+
+  Expected query parameter: query-param=query&api-version=2022-12-01-preview
+  Expected header parameter: header-param=header
+
+  Expected input body:
+  ```json
+  {
+    "stringProperty": "text",
+    "modelProperty": {
+      "int32Property": 1,
+      "float32Property": 1.5,
+      "enumProperty": "EnumValue1"
+    },
+    "arrayProperty": [
+      "item"
+    ],
+    "recordProperty": {
+      "record": "value"
+    }
+  }
+  ```
+
+  Expected response body:
+  ```json
+  {
+    "stringProperty": "text"
+  }
+  ```
+  """)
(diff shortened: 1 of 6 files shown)


msyyc (approved):
Looks good to Python

XiaofeiCao (approved):

packages/cadl-ranch-specs/http/azure/example/basic/client.tsp
```diff
@@ -0,0 +1,50 @@
+import "@azure-tools/typespec-client-generator-core";
+
+import "./main.tsp";
+
+using Azure.ClientGenerator.Core;
+
+@TypeSpec.Versioning.useDependency(_Specs_.Azure.Example.Basic.Versions.v2022_12_01_preview)
+namespace Client;
+
+@client({
```
tadelesh, line 10: a little bit confused why we use client.tsp to add example?

weidongxu-microsoft: I'd like the case verify the correct match of ServiceOperation and JSON example via operationId = `ServiceOperationGroup_Basic`

Just a test to avoid picking up the wrong operationId like `AzureExampleClient_basicAction` from the explicit client.

tadelesh: oh, i see.

tadelesh (approved):

packages/cadl-ranch-specs/http/azure/example/basic/mockapi.ts
```diff
@@ -0,0 +1,31 @@
+import { passOnSuccess, mockapi, json } from "@azure-tools/cadl-ranch-api";
+import { ScenarioMockApi } from "@azure-tools/cadl-ranch-api";
+
+export const Scenarios: Record<string, ScenarioMockApi> = {};
+
+Scenarios.Client_AzureExampleClient_basicAction = passOnSuccess(
+  mockapi.post("/azure/example/basic/basic", (req) => {
```
qiaozha, line 7: in this case the endpoint at scenario service is "/azure/example/basic" and we have a `@route("/azure/example/basic/basic")` decorator, the expected url should be "/azure/example/basic/azure/example/basic/basic" ?

weidongxu-microsoft: It actually be this. Java had this test covered.

My guess of the reason is that when there is customization (and namespace) in client.tsp, some of the decorator of e.g. `@scenarioService` on main.tsp namespace does not work (or not work the same way).

weidongxu-microsoft: Emm, does it cause problem on RLC when no client.tsp is read?

PR #633, opened 2024-07-16
Add typespec/openapi3 to typespec-bump-deps

pshao25:
We have a [test ](https://github.com/Azure/autorest.csharp/blob/b222aaf9cc4ba420322ccd98f45425c2bbca3a49/test/UnbrandedProjects/Platform-OpenAI-TypeSpec/completions/models.tsp#L114)need `typespec/openapi3`.

### DIFF
--- a/.changeset/silly-kings-wink.md
+++ b/.changeset/silly-kings-wink.md
@@ -1,0 +1,5 @@
+---
+"@azure-tools/typespec-bump-deps": patch
+---
+
+Add `@typespec/openapi3` to typespec-bump-deps
--- a/packages/typespec-bump-deps/src/cli/typespec-bump-deps.ts
+++ b/packages/typespec-bump-deps/src/cli/typespec-bump-deps.ts
@@ -7,6 +7,7 @@ const knownPackages = [
   "@typespec/compiler",
   "@typespec/http",
   "@typespec/openapi",
+  "@typespec/openapi3",
   "@typespec/rest",
   "@typespec/versioning",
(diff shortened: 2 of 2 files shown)


weidongxu-microsoft:
Why you need openapi3 now?

pshao25:
> Why you need openapi3 now?

We have [test ](https://github.com/Azure/cadl-ranch/pull/633#issuecomment-2230169511)need this. Updated the description.

weidongxu-microsoft:
I will let other reviews to chime in as well.

I don't see why your test depends on this `@oneOf` for typespec. I believe neither tsp compiler nor TCGC would handle this `@oneOf` decorator.

pshao25:
> I will let other reviewers to chime in as well.
> 
> I don't see why your test depends on this `@oneOf` for typespec. I believe neither tsp compiler nor TCGC would handle this `@oneOf` decorator.

I'm not sure what you mean by "tsp compiler not handle this". Compiler of course need this package otherwise it throw error that "cannot find @oneOf".

weidongxu-microsoft:
> I'm not sure what you mean by "tsp compiler not handle this". Compiler of course need this package otherwise it throw error that "cannot find @OneOf".

What I mean is, does .NET code change, if `@oneOf` is removed from your test.

If no code change, why this `@oneOf` needed in your test?

pshao25:
Well I just want to keep the integrity of this test, since it is OpenAI. But yes I could remove it.

PR #641, opened 2024-07-18
fix some typos in plain date and plain time scenarios

ArcturusZhang:
The scalar type `plainDate` and `plainTime` are two words in typespec definition, there is no reason that we make a difference in cadl-ranch cases.

### DIFF
--- a/packages/cadl-ranch-specs/cadl-ranch-summary.md
+++ b/packages/cadl-ranch-specs/cadl-ranch-summary.md
@@ -6443,9 +6443,9 @@ Expected request body:
 {}
 ```
 
-### Type_Property_Optional_Plaindate_getAll
+### Type_Property_Optional_PlainDate_getAll
 
-- Endpoint: `get /type/property/optional/plaindate/all`
+- Endpoint: `get /type/property/optional/plainDate/all`
 
 Expected response body:
 
@@ -6453,9 +6453,9 @@ Expected response body:
 { "property": "2022-12-12" }
 ```
 
-### Type_Property_Optional_Plaindate_getDefault
+### Type_Property_Optional_PlainDate_getDefault
 
-- Endpoint: `get /type/property/optional/plaindate/default`
+- Endpoint: `get /type/property/optional/plainDate/default`
 
 Expected response body:
 
@@ -6463,9 +6463,9 @@ Expected response body:
 {}
 ```
 
-### Type_Property_Optional_Plaindate_putAll
+### Type_Property_Optional_PlainDate_putAll
 
-- Endpoint: `put /type/property/optional/plaindate/all`
+- Endpoint: `put /type/property/optional/plainDate/all`
 
 Expected request body:
 
@@ -6473,9 +6473,9 @@ Expected request body:
 { "property": "2022-12-12" }
 ```
 
-### Type_Property_Optional_Plaindate_putDefault
+### Type_Property_Optional_PlainDate_putDefault
 
-- Endpoint: `put /type/property/optional/plaindate/default`
+- Endpoint: `put /type/property/optional/plainDate/default`
 
 Expected request body:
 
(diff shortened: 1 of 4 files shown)


haolingdong-msft (approved):
LGTM.

packages/cadl-ranch-specs/http/type/property/optionality/main.tsp
```diff
@@ -112,19 +112,19 @@ model DurationProperty is ModelTemplate<duration>;
...
-@doc("Model with a plaindate property")
-model PlaindateProperty is ModelTemplate<plainDate>;
-@route("/plaindate")
+// Model with optional plainDate property
+@doc("Model with a plainDate property")
+model PlainDateProperty is ModelTemplate<plainDate>;
+@route("/plainDate")
 @operationGroup
-interface Plaindate extends OperationsTemplate<PlaindateProperty, "\"2022-12-12\""> {}
+interface PlainDate extends OperationsTemplate<PlainDateProperty, "\"2022-12-12\""> {}
```
weidongxu-microsoft, line 120: Em, that actually change the client name...

weidongxu-microsoft (approved):

packages/cadl-ranch-specs/http/type/property/optionality/main.tsp
ArcturusZhang: yeah and this is the purpose because plaindate is not actually a word.
I wonder why the cspell here does not report it when we introduce it

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -6443,79 +6443,79 @@ Expected request body:
...
 
 ```json
 { "property": "13:06:12" }
 ```
 
-### Type_Property_Optional_Plaintime_getDefault
+### Type_Property_Optional_PlainTime_getDefault
 
-- Endpoint: `get /type/property/optional/plaintime/default`
+- Endpoint: `get /type/property/optional/plainTime/default`
```
MaryGao, line 6498: just wondering if we have some convention for path, i notice some paths use `-` as break.

weidongxu-microsoft: both is acceptable for azure

https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#http-url-casing

packages/cadl-ranch-specs/http/type/property/optionality/main.tsp
qiaozha, line 120: PlainDate is not in the format of "2022-12-12", it should be "Mon Dec 12 2022", see https://typespec.io/docs/standard-library/built-in-data-types#plainDate

packages/cadl-ranch-specs/http/type/property/optionality/main.tsp
```diff
@@ -112,19 +112,19 @@ model DurationProperty is ModelTemplate<duration>;
...
 
 // Model with optional  property
-@doc("Model with a plaintime property")
+@doc("Model with a plainTime property")
 model PlainTimeProperty is ModelTemplate<plainTime>;
-@route("/plaintime")
+@route("/plainTime")
 @operationGroup
-interface Plaintime extends OperationsTemplate<PlainTimeProperty, "\"13:06:12\""> {}
+interface PlainTime extends OperationsTemplate<PlainTimeProperty, "\"13:06:12\""> {}
```
qiaozha, line 127: similar for plainTime type

packages/cadl-ranch-specs/http/type/property/optionality/main.tsp
qiaozha: @timotheeguerin Could you help confirm here? I notice that you have examples for plainDate here https://github.com/microsoft/typespec/blob/main/packages/compiler/lib/intrinsics.tsp#L107 and it says that Create a plain date from an ISO 8601 string, and according to https://ijmacd.github.io/rfc3339-iso8601/ 2022-12-12 is a valid plainDate string and April 10th is not ?

timotheeguerin: Yeah plainDate and plaintive should be in the iso format over json by default. Should update the encoding doc to include that

PR #643, opened 2024-07-18
Addition of initial XML tests

jhendrixMSFT:
Covers some basic scenarios. More will be added.


jhendrixMSFT:
More tests are needed. However, I want to ensure that things look ok before I add more.

packages/cadl-ranch-specs/http/payload/xml/main.tsp
```diff
@@ -0,0 +1,195 @@
...
+Expected response body:
+```xml
+<SimpleModel>
+  <Name>foo</Name>
+  <Age>123</Age>
+</SimpleModel>
+```
+""")
+@get
+op getSimpleModel(): {
```
timotheeguerin, line 48: Nit but I would create a structure where we have 

interface SimpleModel {
  get()
  put();
}

make it nicer in the dashboard, you can then also use template to easily define both as I assume the get and put should send/receive teh same data

timotheeguerin: there is a lot of other spec with this `GetOrSend` tempalte interface that is reused for example

packages/cadl-ranch-specs/http/payload/xml/main.tsp
```diff
@@ -0,0 +1,195 @@
...
+  @body input: SimpleModel,
+): void;
+
+@scenario
+@route("/modelWithSimpleArrays")
+@scenarioDoc("""
+Returns an instance of ModelWithSimpleArrays.
+Expected response body:
+```xml
+<ModelWithSimpleArrays>
```
timotheeguerin, line 77: By default properties are always wrapped so this should be 

```
<ModelWithSimpleArrays>
  <Colors>
     <string>red<string>
     <string>green<string>
    <string>blue<string>
  </Colors>
  <Counts>
     <string>1<string>
     <string>2<string>
  </Counts>
</ModelWithSimpleArrays>

```

packages/cadl-ranch-specs/http/payload/xml/main.tsp
```diff
@@ -0,0 +1,195 @@
...
+
+@scenario
+@route("/simpleModel")
+@scenarioDoc("""
+Sends an instance of SimpleModel.
+Expected request body:
+```xml
+<SimpleModel>
+  <Name>foo</Name>
+  <Age>123</Age>
```
timotheeguerin, line 61: I don't this we should Pascale case the properties automatically. The spec doesn't say that

packages/cadl-ranch-specs/http/payload/xml/main.tsp
```diff
@@ -0,0 +1,195 @@
...
+): void;
+
+@scenario
+@route("/modelWithArrayOfModel")
+@scenarioDoc("""
+Returns an instance of ModelWithArrayOfModel.
+Expected response body:
+```xml
+<ModelWithArrayOfModel>
+  <SimpleModel>
```
timotheeguerin, line 120: same here should always wrap, doc for example https://typespec.io/docs/next/libraries/xml/guide

packages/cadl-ranch-specs/http/payload/xml/main.tsp
jhendrixMSFT: I recall that we support both in autorest.testserver. However, I don't know if that was for some hypothetical scenario or if it actually exists.

FWIW, my plan was to also add tests with wrapped arrays. If we don't think we need the unwrapped variant, I can convert the existing ones to be wrapped.

jhendrixMSFT: Ok I see the default is wrapped. I will convert these and then add tests for the unwrapped variants.

packages/cadl-ranch-specs/http/payload/xml/mockapi.ts
```diff
@@ -0,0 +1,117 @@
...
+    <Name>bar</Name>
+    <Age>456</Age>
+  </SimpleModel>
+</ModelWithArrayOfModel>
+`;
+
+const modelWithOptionalField = `
+<ModelWithOptionalField>
+  <Item>widget</Item>
+  <Counts />
```
jhendrixMSFT, line 39: For optional fields, should we use an empty element (i.e. _not_ omit it)?

packages/cadl-ranch-specs/http/payload/xml/main.tsp
jhendrixMSFT: I can change this, but I presume we need to support both cases (storage pascal-cases their XML responses). If the default expectation is to assume camel-case for properties, I can change this and add separate tests for pascal case.

timotheeguerin: It's up to a service to judge what they want to do, either they use the pascal case name to start with or use the xml.name decorator/@encodedName to specify the right name over the wire

packages/cadl-ranch-specs/http/payload/xml/mockapi.ts
timotheeguerin: I think you wouldn't pass the field at all

packages/cadl-ranch-specs/http/payload/xml/main.tsp
jhendrixMSFT: fixed

packages/cadl-ranch-api/src/request-validations.ts
```diff
@@ -34,6 +36,38 @@ export const validateBodyEquals = (request: RequestExt, expectedBody: unknown |
...
 
+export const validateXmlBodyEquals = (request: RequestExt, expectedBody: string): void => {
+  if (isBodyEmpty(request.rawBody)) {
+    throw new ValidationError(BODY_EMPTY_ERROR_MESSAGE, expectedBody, request.rawBody);
+  }
+
+  expectedBody = `<?xml version='1.0' encoding='UTF-8'?>` + expectedBody;
+
+  let actualParsedBody = "";
+  parseString(request.rawBody!, (err: Error | null, result: any): void => {
```
timotheeguerin, line 47: I think this can be simplified to use the `parseStringPromise` version so you can do 

```
const actualParsedBody = await parseStringPromise(request.rawbody);
```

packages/cadl-ranch-api/src/request-validations.ts
```diff
@@ -34,6 +36,38 @@ export const validateBodyEquals = (request: RequestExt, expectedBody: unknown |
   }
 };
 
+export const validateXmlBodyEquals = (request: RequestExt, expectedBody: string): void => {
+  if (isBodyEmpty(request.rawBody)) {
```
timotheeguerin, line 40:
```suggestion
  if (request.rawBody === undefined || isBodyEmpty(request.rawBody)) {
```

that way you don't need the `!` below and it is accurate

pnpm-lock.yaml
timotheeguerin, line 1: I think there is a lot of unexpected change here, you are reverting packages version which explain some of the prettier diff

packages/cadl-ranch-api/src/request-validations.ts
jhendrixMSFT: I thought about that but wasn't sure if it was ok to change this (and callers) to async functions.

timotheeguerin: I think its fair, specially as the code you have technically wouldn't work if the callback is actually called async(not sure when it happens)

jhendrixMSFT: Yeah, changing it all to async causes the test server to return a 500 when the request body doesn't match the expected body.

pnpm-lock.yaml
jhendrixMSFT: I see, will clean up. I think there was some unexpected changes as I was on `v8.x` of pnpm.

packages/cadl-ranch-api/src/request-validations.ts
jhendrixMSFT: Hmm that's not true. I'm seeing 500 even with the sync version.

packages/cadl-ranch-api/src/request-validations.ts
jhendrixMSFT: fixed

pnpm-lock.yaml
jhendrixMSFT: fixed

packages/cadl-ranch-specs/http/payload/xml/main.tsp
jhendrixMSFT: fixed

packages/cadl-ranch-specs/http/payload/xml/main.tsp
jhendrixMSFT: fixed

packages/cadl-ranch-specs/http/payload/xml/mockapi.ts
jhendrixMSFT: Do you think omitting it is the correct default behavior? I bring this up because we had a bug in the Go SDK where tables would sometimes send an empty element (a timestamp IIRC) and we didn't properly handle that case. If an empty element is the uncommon case for "no value" then I can add a separate test for that.

packages/cadl-ranch-specs/http/payload/xml/main.tsp
jhendrixMSFT: Ok, so if I correctly understand the [guide](https://typespec.io/docs/next/libraries/xml/guide), the casing for properties should match the casing in the tsp. Is that what we're saying?

timotheeguerin: yeah

packages/cadl-ranch-specs/http/payload/xml/mockapi.ts
timotheeguerin: if I see this I kinda see it equivalent to sending that as json
```json
{item: "widget", counts: []}
```
instead of 
```json
{item:  "widget"}
```

packages/cadl-ranch-specs/http/payload/xml/main.tsp
jhendrixMSFT: fixed

packages/cadl-ranch-specs/http/payload/xml/mockapi.ts
jhendrixMSFT: Here, `counts` is an `int32`, not array, but I see your point (I'm going to rename this field to avoid confusion with other models).

This feels similar to JSON nulls. For GETs, it seems that omission of the `counts` field or an empty element are semantically equivalent. I've seen this in the wild, so we do need to ensure that generated code properly handles it.

But for PUTs etc, omission vs empty element is _not_ equivalent. If I omit the element, then it means "preserve the existing value". But if I send an empty element, then it means "erase the existing value". I don't know if this is a real scenario though.

jhendrixMSFT: Checked the blob storage [REST API docs](https://learn.microsoft.com/rest/api/storageservices/set-blob-service-properties?tabs=microsoft-entra-id) and the omission vs empty element on PUT is real.

I will change this test to cover the omission scenario and add separate test for empty element.

timotheeguerin: yeah exactly, for serializing I assume you wouldn't put it ever but for deserializing it I guess it should be equivalent as long as you have the schema that says what type it expect

timotheeguerin (approved):
I think this looks correct

lmazuel (approved):

.changeset/silent-doors-pull.md
```diff
@@ -0,0 +1,2 @@
+---
```
tadelesh, line 1: @timotheeguerin it seems ci not fail for this empty changelog, is it a bug?

timotheeguerin: yeah changesets requirement for passing the CI is having a changeset file doesn't matter about the content.

We can change to use chronus here too, i just didn't do it because there was teh talk about moving it directly into typespec repo but not sure where thats' going

PR #648, opened 2024-07-23
Add more tests to payload/xml

jhendrixMSFT:
Includes types that use TypeSpec.Xml decorators.

### DIFF
--- a/packages/cadl-ranch-specs/http/payload/xml/main.tsp
+++ b/packages/cadl-ranch-specs/http/payload/xml/main.tsp
@@ -38,11 +38,17 @@ model ModelWithOptionalField {
 model ModelWithAttributes {
   @attribute id1: int32;
   @attribute id2: string;
-  @name("Enabled") enabled: boolean;
+  enabled: boolean;
 }
 
 @doc("Contains fields of wrapped and unwrapped arrays of primitive types.")
 model ModelWithUnwrappedArray {
+  @unwrapped colors: string[];
+  counts: int32[];
+}
+
+@doc("Contains fields of wrapped and unwrapped arrays of primitive types that have different XML representations.")
+model ModelWithRenamedArrays {
   @name("Colors") @unwrapped colors: string[];
   @name("Counts") counts: int32[];
 }
@@ -68,6 +74,13 @@ model ModelWithText {
 @doc("Contains a dictionary of key value pairs.")
 model ModelWithDictionary {
   metadata: Record<string>;
+}
+
+@doc("Uses encodedName instead of Xml.Name which is functionally equivalent.")
+@encodedName("application/xml", "ModelWithEncodedNamesSrc")
+model ModelWithEncodedNames {
+  @encodedName("application/xml", "SimpleModelData") modelData: SimpleModel;
+  @encodedName("application/xml", "PossibleColors") colors: string[];
 }
 
 @doc("Template for XML operations")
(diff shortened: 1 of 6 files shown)


packages/cadl-ranch-specs/http/payload/xml/main.tsp
```diff
@@ -32,6 +34,42 @@ model ModelWithOptionalField {
   value?: int32;
 }
 
+@doc("Contains fields that are XML attributes.")
+model ModelWithAttributes {
+  @attribute id1: int32;
+  @attribute id2: string;
+  @name("Enabled") enabled: boolean;
```
timotheeguerin, line 41: i think you have casing discrepancies between the docs on model

timotheeguerin, line 41: could we keep trhe renaming of props as a separate dedicated test.

Should also add one that is done with `@encodedName("application/xml", "abc")` which should be equivalent to Xml.name

jhendrixMSFT: done

packages/cadl-ranch-specs/http/payload/xml/main.tsp
```diff
@@ -32,6 +34,55 @@ model ModelWithOptionalField {
...
+  counts: int32[];
+}
+
+@doc("Contains fields of wrapped and unwrapped arrays of primitive types that have different XML representations.")
+model ModelWithRenamedArrays {
+  @name("Colors") @unwrapped colors: string[];
+  @name("Counts") counts: int32[];
+}
+
+@doc("Contains fields of the same type that have different XML representation.")
```
tadelesh, line 56: a little confused about "same type that have different XML representation"

tadelesh (approved):

packages/cadl-ranch-specs/http/payload/xml/main.tsp
```diff
@@ -130,3 +181,127 @@ interface ModelWithOptionalFieldValue
...
+  extends XmlOperations<
+      ModelWithUnwrappedArray,
+      """
+  <ModelWithUnwrappedArray>
+    <colors>red</colors>
+    <colors>green</colors>
+    <colors>blue</colors>
+    <counts>
+      <int32>1</int32>
+      <int32>2</int32>
```
tadelesh, line 209: is this the convention to use type name as tag name?

jhendrixMSFT: Yes, see https://typespec.io/docs/next/libraries/xml/guide

packages/cadl-ranch-specs/http/payload/xml/main.tsp
jhendrixMSFT: This type has two fields that are both of type `SimpleModel` but have different XML names, `InputData` and `OutputData`.

It's the same as [this](https://github.com/Azure/azure-rest-api-specs/blob/a58614a53ecd06c6e6638b58c407d5bd8e61c009/specification/storage/data-plane/Microsoft.BlobStorage/stable/2021-12-02/blob.json#L10870-L10880) scenario in storage.

packages/cadl-ranch-specs/http/payload/xml/main.tsp
tadelesh: i see. thanks for the info.

PR #650, opened 2024-07-29
Add test for templated type

melina5656:
This PR includes some tests from the [https://github.com/Azure/autorest.csharp/blob/feature/v3/test/TestProjects/Models-TypeSpec/Models-TypeSpec.tsp](url). I have moved some tests to http/type/model/templated.


packages/cadl-ranch-specs/http/type/model/generic/main.tsp
```diff
@@ -0,0 +1,50 @@
...
+    [
+      1234
+    ],
+    "value": 1234,
+    "field": "<field>",
+}
+```
+""")
+@route("/genericType")
+@get
```
pshao25, line 49: get operation should not have body. 
for this case, it's better to have a put with both request body and response body.

packages/cadl-ranch-specs/http/type/model/generic/main.tsp
```diff
@@ -0,0 +1,50 @@
...
+import "@azure-tools/typespec-client-generator-core";
+
+using TypeSpec.Http;
+using Azure.ClientGenerator.Core;
+
+@doc("Illustrates the model generic cases.")
+@scenarioService("/type/model/generic")
+namespace Type.Model.Generic;
+
+@doc("Generic")
```
pshao25, line 12: Please enrich the document.

packages/cadl-ranch-specs/http/type/model/generic/main.tsp
```diff
@@ -0,0 +1,50 @@
...
+model Generic {
+  @doc("A field to facet by, where the field is attributed as 'facetable'")
+  field: string;
+}
+
+model NumericType<T extends numeric> extends Generic {
+  @doc("The facet ranges to produce. The values must be listed in ascending order to get the expected results. For example, values=10,20 produces three buckets: one for base rate 0 up to but not including 10, one for 10 up to but not including 20, and one for 20 and higher.")
```
pshao25, line 19: Change this document. The test has nothing to do with facet.

packages/cadl-ranch-specs/http/type/model/generic/main.tsp
```diff
@@ -0,0 +1,50 @@
...
+
+using TypeSpec.Http;
+using Azure.ClientGenerator.Core;
+
+@doc("Illustrates the model generic cases.")
+@scenarioService("/type/model/generic")
+namespace Type.Model.Generic;
+
+@doc("Generic")
+model Generic {
```
pshao25, line 13: I think one layer of inheritance is enough. We don't need this model.

packages/cadl-ranch-specs/http/type/model/generic/main.tsp
```diff
@@ -0,0 +1,41 @@
...
+  [
+    1234
+  ],
+  "value": 1234
+}
+```
+""")
+@route("/genericType")
+@put
+op genericType(@body input: Int32Type): void;
```
timotheeguerin, line 41: What is special about those generic types? Is it the template? To note TypeSpec use template(Get fully rendered as different instances) not generic

pshao25: It is the `extends` makes it special. The background is we have some requests [here ](https://github.com/Azure/autorest.csharp/issues/3655)and [here](https://github.com/Azure/autorest.csharp/issues/3263#issuecomment-1786503820). The ideal solution is we could generate a real generic type for `NumericType<T extends numeric>` in C# something like `class NumericType<T> where T is number`. However, architects decided we don't want to generate generic types for now. They decided we only support instantiated type like `Int32Type` in this case. That is why we added `Int32Type` here.

> To note TypeSpec use template(Get fully rendered as different instances) not generic

Will you suggest a different terminology for an instantiated model extending templated model?

packages/cadl-ranch-specs/http/type/model/generic/main.tsp
```diff
@@ -0,0 +1,41 @@
+import "@typespec/http";
+import "@azure-tools/cadl-ranch-expect";
+import "@azure-tools/typespec-client-generator-core";
+
+using TypeSpec.Http;
+using Azure.ClientGenerator.Core;
+
+@doc("Illustrates the model generic cases.")
```
pshao25, line 8:
```suggestion
@doc("Illustrates the model generic cases. There is a base generic type and an instantiated type extending from it.")
```

packages/cadl-ranch-specs/http/type/model/generic/main.tsp
```diff
@@ -0,0 +1,41 @@
...
+namespace Type.Model.Generic;
+
+model NumericType<T extends numeric> {
+  @doc("This model is used to perform facet processing on fields based on a specified range of values.")
+  values: T[];
+
+  value: T;
+}
+
+@doc("Facets an int32 field by the specified value ranges.")
```
pshao25, line 19: Use pattern 
```
/**
*/
```
instead of `@doc`.

packages/cadl-ranch-specs/http/type/model/generic/main.tsp
```diff
@@ -0,0 +1,52 @@
...
+  /**
+   * An array of numeric values.
+   */
+  values: T[];
+
+  value: T;
+}
+
+/**
+ * The parameters for the generic type operation.
```
pshao25, line 27:
```suggestion
 * An instantiated type representing int32 type.
```

packages/cadl-ranch-specs/http/type/model/generic/main.tsp
```diff
@@ -0,0 +1,52 @@
...
+ * Illustrates the model generic cases. There is a base generic type and an instantiated type extending from it.
+ */
+@scenarioService("/type/model/generic")
+namespace Type.Model.Generic;
+
+/**
+ * A template model for handling numeric types.
+ */
+model NumericType<T extends numeric> {
+  /**
```
pshao25, line 18: I think we could remove this comment to keep consistence.

packages/cadl-ranch-specs/http/type/model/generic/main.tsp
```diff
@@ -0,0 +1,52 @@
...
+  [
+    1234
+  ],
+  "value": 1234
+}
+```
+""")
+@route("/genericType")
+@put
+op genericType(@body input: Int32Type): void;
```
pshao25, line 52: put operation should have return type. we could use Int32Type here.

packages/cadl-ranch-specs/http/type/model/generic/main.tsp
```diff
@@ -0,0 +1,49 @@
...
+  [
+    1234
+  ],
+  "value": 1234
+}
+```
+""")
+@route("/genericType")
+@put
+op genericType(@body input: Int32Type): Int32Type;
```
weidongxu-microsoft, line 49: Document the response too.

packages/cadl-ranch-specs/http/type/model/generic/main.tsp
```diff
@@ -0,0 +1,49 @@
...
+using TypeSpec.Http;
+using Azure.ClientGenerator.Core;
+
+/**
+ * Illustrates the model generic cases. There is a base generic type and an instantiated type extending from it.
+ */
+@scenarioService("/type/model/generic")
+namespace Type.Model.Generic;
+
+model NumericType<T extends numeric> {
```
weidongxu-microsoft, line 14: I assume a template should have a `@friendlyName`? @tadelesh

packages/cadl-ranch-specs/http/type/model/generic/main.tsp
weidongxu-microsoft, line 1: I am not sure whether `model/generic` is appropriate here.

Maybe `model/template`?

pshao25 (approved):

packages/cadl-ranch-specs/http/type/model/templated/main.tsp
```diff
@@ -0,0 +1,62 @@
...
+using TypeSpec.Http;
+using Azure.ClientGenerator.Core;
+
+/**
+ * Illustrates the model templated cases. There is a base templated type and an instantiated type extending from it.
+ */
+@scenarioService("/type/model/templated")
+namespace Type.Model.Templated;
+
+@friendlyName("{name}Template", T)
```
tadelesh, line 14:
```suggestion
@friendlyName("{name}Type", T)
```

packages/cadl-ranch-specs/http/type/model/templated/main.tsp
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * An instantiated type representing int32 type.
+ */
+model Int32Type extends NumericType<int32> {
+  /**
+   * The Kind of the Int32Type.
+   */
+  kind: "Int32Values";
+}
```
tadelesh, line 32: it is really weird that this `Int32Type` extends from `NumericType` and no other properties added. could you try to change your model templated test to a more reasonable scenario.

weidongxu-microsoft: I guess `kind` is "added" in this model, though it is a constant.

Agree it is not very meaningful model. Do we have any source of such usage in service (and maybe adjust this a bit from that source)?

pshao25: Please see the story here https://github.com/Azure/cadl-ranch/pull/650#discussion_r1699342839

pshao25: Do you guys have suggestion on testing such case?

weidongxu-microsoft: I don't have any particular case. The thought is that `NumericValuesFacet<T extends numeric>` is not meaningful, if only `model Int32Type` extends it.

A more meaningful case is that there be another e.g. `model FloatType extends NumericValuesFacet<Float>`.

pshao25: I would prefer add another float. There are already a lot of "a direct use of xxx"  cases.

melina5656: A float type has been added

tadelesh (approved):

packages/cadl-ranch-specs/http/type/model/templated/main.tsp
```diff
@@ -0,0 +1,72 @@
...
+    [
+      1234
+    ],
+    "value": 1234
+  }
+  ```
+  """)
+@route("/templatedType")
+@put
+op templatedType(@body input: Int32Type): Int32Type;
```
iscai-msft, line 72: do we expect to add more templated types to the test? If so, we can make this an operation group and this specific one `int32`. If not, we can keep it as is

packages/cadl-ranch-specs/http/type/model/templated/main.tsp
```diff
@@ -0,0 +1,72 @@
...
+  /**
+   * The Kind of the Int32Type.
+   */
+  kind: "Int32Values";
+}
+
+/**
+ * An instantiated type representing float type.
+ */
+model FloatType extends NumericType<float> {
```
ArcturusZhang, line 37: we should use float32 instead of float

packages/cadl-ranch-specs/http/type/model/templated/main.tsp
```diff
@@ -0,0 +1,72 @@
...
+   */
+  values: T[];
+
+  value: T;
+}
+
+/**
+ * An instantiated type representing int32 type.
+ */
+model Int32Type extends NumericType<int32> {
```
ArcturusZhang, line 27: so this is one scenario that we extends a generic type.
I think we need to add some more operation takes `NumericType<T>` directly.

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -5725,6 +5725,81 @@ Expected input body:
 }
 ```
 
+### Type_Model_Templated_TemplatedTypes_float32TemplatedType
```
iscai-msft, line 5728: 1. Do we need a subnamespace for `TemplatedTypes`?
2. It can just be `Type_Model_Templated_float32` (or `Type_Model_Templated_TemplatedTypes_float32` if we still want `TemplatedTypes` operation group

packages/cadl-ranch-specs/http/type/model/templated/mockapi.ts
```diff
@@ -0,0 +1,49 @@
...
+      body: json(body),
+    };
+  }),
+);
+
+Scenarios.Type_Model_Templated_TemplatedTypes_float32TemplatedType = passOnSuccess(
+  mockapi.put("/type/model/templated/float32ValuesType", (req) => {
+    const body = {
+      kind: "Float32Values",
+      values: [1234],
```
iscai-msft, line 25: shouldn't use an int, you can use `12.34`

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -5725,6 +5725,81 @@ Expected input body:
...
+
+```json
+{
+  "kind": "Int32Values",
+  "values": [1234],
+  "value": 1234
+}
+```
+
+### Type_Model_Templated_numericTemplatedType
```
iscai-msft, line 5779: @melina5656 last nit, but otherwise looks great. Can you just rename them to `Type_Model_Templated_numeric` etc? We don't need `TemplatedType` again

melina5656: Hi @iscai-msft, If I simplify the name of the operation to `numeric`, `float32`, `int32`, it will lead to the problem of `the constraint of type parameter' T 'referencing' T 'itself`. Therefore, I would like to ask if it is possible to change the name of the operation to `numericType`, `float32Type`, `int32Type`.

iscai-msft: ah that's a good point, I didn't think about that. Yeah go ahead with `Templated_numericType`

melina5656: Hi @iscai-msft , I have uploaded the latest changes. If you feel that the code meets the requirements, please leave a comment in this conversation and I will merge this PR after receiving the message.

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -6104,6 +6104,81 @@ Expected input body:
...
+```json
+{
+  "kind": "Int32Values",
+  "values":
+  [
+    1234
+  ],
+  "value": 1234
+}
+````
```
weidongxu-microsoft, line 6146: Check the red part. The markdown is wrong in your scenarioDoc.

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -6104,6 +6104,81 @@ Expected input body:
...
+Expected input body:
+
+````json
+{
+  "kind": "Float32Values",
+  "values":
+  [
+    12.34
+  ],
+  "value": 12.34
```
weidongxu-microsoft, line 6120: Use an exact float value in test, e.g. `0.5`, `1.25`, etc.

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -6104,6 +6104,81 @@ Expected input body:
...
+}
+Expected response body:
+{
+  "values":
+  [
+    1234
+  ],
+  "value": 1234
+}
+
```
weidongxu-microsoft, line 6181: here too, wrong markdown

weidongxu-microsoft (approved):

PR #653, opened 2024-07-31
Mix api version0722

mcgallan:
related issue: https://github.com/Azure/autorest.csharp/issues/4852
related code: https://github.com/Azure/autorest.csharp/blob/feature/v3/test/TestProjects/MixAPIVersion-TypeSpec/MixAPIVersion-TypeSpec.tsp


.changeset/many-forks-enjoy.md
```diff
@@ -0,0 +1,5 @@
+---
+"@azure-tools/cadl-ranch-specs": major
```
timotheeguerin, line 2: patch

mcgallan: fixed

packages/cadl-ranch-specs/http/server/versions/versioned/main.tsp
```diff
@@ -64,3 +64,9 @@ op withPathApiVersion(@path apiVersion: string): OkResponse;
 @route("/with-query-old-api-version")
 @head
 op withQueryOldApiVersion(@query("api-version") apiVersion: string): OkResponse;
+
+@scenario
+@scenarioDoc("A simple operation with path api-version, that do NOT use the default but '2022-07-09'.Expected url: '/with-old-path-api-version?api-version=2022-07-09'.")
+@route("/with-query-unknown-api-version")
+@head
+op withQueryUnknownApiVersion(@path apiVersion: "2022-07-09"): OkResponse;
```
chunyu3, line 72: Is apiVersion a query parameter? but you define it as a path paramerter.

mcgallan: fixed

packages/cadl-ranch-specs/http/server/versions/versioned/main.tsp
```diff
@@ -64,3 +64,9 @@ op withPathApiVersion(@path apiVersion: string): OkResponse;
 @route("/with-query-old-api-version")
 @head
 op withQueryOldApiVersion(@query("api-version") apiVersion: string): OkResponse;
+
+@scenario
+@scenarioDoc("A simple operation with path api-version, that do NOT use the default but '2022-07-09'.Expected url: '/with-old-path-api-version?api-version=2022-07-09'.")
+@route("/with-query-unknown-api-version")
+@head
+op withQueryUnknownApiVersion(@query("api-version") apiVersion: "2022-07-09"): OkResponse;
```
tadelesh, line 72: why could we have a constant apiversion?

tadelesh: @chunyu3 is it a valid case to set a constant api version that not exist in the service api version enum?

chunyu3: hello @tadelesh it is possible. When there existing client which defined in a set of api-version, then they defined another client which will re-define another set of api-version, and then they combine those two client into one service (one sdk), in that case the previous will not use the service apiversion.
The more common case is that the previous client is use path api-version parameter (v1), and then the service use query api-version

Maybe we can update this to:
op withPathUnknownApiVersion(@path apiVersion: "v1"): OkResponse

iscai-msft: I think @tadelesh is more concerned with the constant nature of the api version. I think this would be better to test overriding. It doesn't have to be a path api version overriding an api version, but I think testing that a method-level setting overrides the client setting is a good test to add if we don't have it

chunyu3: This test belongs to `method-level setting overriding client setting` scenario somehow, but not exact the same. `apiversion` is a special parameter which will automatically bump to client-level parameter, only when it is a constant, we will not use the client-level paramter to create the HttpRequest, but use the constant value directly.

tadelesh: @chunyu3 , do you have any real case that could help me to understand the scenario?

PR #654, opened 2024-07-31
Add second C#

lmazuel:


packages/cadl-ranch-dashboard/src/apis.ts
```diff
@@ -12,6 +12,7 @@ export type GeneratorNames =
   | "typescript/rlc"
   | "typescript/modular"
   | "csharp"
+  | "typespec/http-client-csharp"
```
timotheeguerin, line 15:
```suggestion
  | "@typespec/http-client-csharp"
```

can we just use the full name, that way won't need to update that again later

timotheeguerin: i thikn also need to update a list below too

lmazuel: Gotcha. Feels consistent now with PR like https://github.com/Azure/cadl-ranch/pull/586/files

timotheeguerin (approved):

weidongxu-microsoft (approved):

PR #655, opened 2024-08-02
specs, fix azure example route

weidongxu-microsoft:

### DIFF
--- a/packages/cadl-ranch-specs/http/azure/example/basic/client.tsp
+++ b/packages/cadl-ranch-specs/http/azure/example/basic/client.tsp
@@ -2,9 +2,11 @@ import "@azure-tools/typespec-client-generator-core";
 
 import "./main.tsp";
 
+using TypeSpec.Http;
 using Azure.ClientGenerator.Core;
 
 @TypeSpec.Versioning.useDependency(_Specs_.Azure.Example.Basic.Versions.v2022_12_01_preview)
+@route("/azure/example/basic")
 namespace Client;
 
 @client({
--- a/.changeset/green-wasps-sin.md
+++ b/.changeset/green-wasps-sin.md
@@ -1,0 +1,5 @@
+---
+"@azure-tools/cadl-ranch-specs": patch
+---
+
+Bug fix on azure example test route
--- a/packages/cadl-ranch-specs/http/azure/example/basic/main.tsp
+++ b/packages/cadl-ranch-specs/http/azure/example/basic/main.tsp
@@ -50,7 +50,7 @@ model ActionResponse is ActionRequest;
 
 interface ServiceOperationGroup {
   #suppress "@azure-tools/cadl-ranch-expect/missing-scenario" "scenario defined in client.tsp"
-  @route("/azure/example/basic/basic")
+  @route("/basic")
   @post
   basic(
     ...ApiVersionParameter,


packages/cadl-ranch-specs/http/azure/example/basic/client.tsp
```diff
@@ -2,9 +2,11 @@ import "@azure-tools/typespec-client-generator-core";
 
 import "./main.tsp";
 
+using TypeSpec.Http;
 using Azure.ClientGenerator.Core;
 
 @TypeSpec.Versioning.useDependency(_Specs_.Azure.Example.Basic.Versions.v2022_12_01_preview)
+@route("/azure/example/basic")
 namespace Client;
```
weidongxu-microsoft, line 10: `@scenarioService` add this route to namespace in main.tsp
https://github.com/azure/cadl-ranch/blob/main/packages/cadl-ranch-expect/src/decorators.ts#L242

Therefore, I need to add the same `@route` to this namespace in client.tsp as well.

@timotheeguerin Would you let me know if this is the correct understanding and the correct fix to make the route same (via main.tsp and via client.tsp)?

timotheeguerin: yeah that would be the way but is it necessary to have a client.tsp here or can't you do `@@client` on the service namespace. Leave client structure testing to another test

weidongxu-microsoft: Got it. Thanks Timothee.

The client.tsp is here intended to verify the correctness of `operationId` (as interface/operation in client.tsp can provide a false `operationId: AzureExampleClient_BasicAction`. The intention is that example loader find the correct one in main.tsp as `operationId: ServiceOperationGroup_Basic`.

qiaozha (approved):

packages/cadl-ranch-specs/http/azure/example/basic/client.tsp
qiaozha: I thought in this case `_Specs_.Azure.Example.Basic`, but it was not, which is contradict with my understanding of global namespace.

tadelesh (approved):

packages/cadl-ranch-specs/http/azure/example/basic/client.tsp
timotheeguerin: @qiaozha no sure i get the problem here can you share more details?

qiaozha: Here we have specified the endpoint, which should be a host template right ? https://github.com/Azure/cadl-ranch/pull/655/files#diff-ddd397f87276d566b5395a2c63595c897126244510eede7bc2675652bb616af1L13 
and operation route path should be `/azure/example/basic/basic` and the final url should be endpoint host template + operation route path, we get the `/azure/example/basic/azure/example/basic/basic`.
However, we got `/azure/example/basic/azure/example/basic/basic` directly in the operation route path and the host template is empty if we read from main.tsp, but if we read from client.tsp, the operation route path is  `/azure/example/basic/basic`, and still no host template.

timotheeguerin: `@serviceScenario` just calls `@route` with that value. When you do `op is` it doesn't pull in values from containers so this change just copies over the info that is not carried over

PR #656, opened 2024-08-06
specs, add test for encode integer as string

weidongxu-microsoft:
Fix https://github.com/Azure/cadl-ranch/issues/634

Not include case of "query" and "header", because they are always string on wire.

"body property" tests on
1. `safeint` (int64) as required property
2. `uint32` as optional property


packages/cadl-ranch-specs/http/encode/string/main.tsp
```diff
@@ -0,0 +1,40 @@
...
+    "value": "1"
+  }
+  ```
+  """)
+  @post
+  op encodeAsString(@body body: IntegerAsStringProperty): IntegerAsStringProperty;
+
+  model IntegerAsStringProperty {
+    @encode(string)
+    value: safeint;
```
archerzz, line 38: Shall we add a series of test cases like `integer`, `int8`... etc?

packages/cadl-ranch-specs/http/encode/string/main.tsp
```diff
@@ -0,0 +1,40 @@
...
+  {
+    "value": "1"
+  }
+  ```
+  """)
+  @post
+  op encodeAsString(@body body: IntegerAsStringProperty): IntegerAsStringProperty;
+
+  model IntegerAsStringProperty {
+    @encode(string)
```
weidongxu-microsoft, line 37: Current TCGC does not handle the ModelProperty path. `@encode(string) value: safeint;` does not work.

It has to be
```
  @encode(string)
  scalar safeintEncodedAsString extends safeint;

  model IntegerAsStringProperty {
    value: safeintEncodedAsString;
  }
```

image

Need a fix.

packages/cadl-ranch-specs/http/encode/string/main.tsp
weidongxu-microsoft: Personally I'd avoid adding type that would be same client type for language.

Let me know what matters in your and I can add them.

weidongxu-microsoft: I've added another `int32` (optional property).

haolingdong-msft (approved):

packages/cadl-ranch-specs/http/encode/string/main.tsp
```diff
@@ -0,0 +1,65 @@
...
+
+@operationGroup
+@route("/property")
+namespace Property {
+  @route("/safeint")
+  @scenario
+  @scenarioDoc("""
+    Test operation with request and response model contains integer property with string encode.
+    Expected request body:
+    ```json
```
msyyc, line 21: One thing to confirm: for now we only support for int, not including float/decimal?

weidongxu-microsoft: I am not exactly sure which is the final decision (long discussion and I wasn't in the meeting that reached decision).

I assume Laurent want it simple as int to string https://github.com/Azure/typespec-azure/issues/1179#issuecomment-2233697638
Me too...

tadelesh (commented):
shall we change the folder to numeric? the previous tests for encode is named by original type name.

msyyc (approved):

weidongxu-microsoft:
> shall we change the folder to numeric? the previous tests for encode is named by original type name.

Make sense. Moved to `encode/numeric` folder (and namespace).

archerzz (approved):

tadelesh (approved):

packages/cadl-ranch-specs/http/encode/string/main.tsp
weidongxu-microsoft: From current lint, it appears that other numeric like `float32` etc. also won't trigger a error (`boolean` or `url` do).

PR #668, opened 2024-08-08
Upgrade cadl-ranch to 0.59.0

timotheeguerin:


packages/cadl-ranch-specs/http/authentication/api-key/main.tsp
```diff
@@ -17,16 +17,16 @@ op valid(): NoContentResponse;
 
 @scenario
 @scenarioDoc("""
-Expect error code 403 and error body:
-  ```json
-  {
-    "error": {
-      "code": "InvalidApiKey",
-      "message": "API key is invalid"
+  Expect error code 403 and error body:
```
timotheeguerin, line 20: Those change are actually from an older version of the formatter which hasn't been upgraded in a while(5 version)

chrisradek (approved):

iscai-msft (approved):

PR #673, opened 2024-08-08
Add new spec for URI template

timotheeguerin:

There is a lot of new scenario covering all the cases. The goal is not necessarily that all emitter should support all the cases immediately but this will then help keeping track of uri template support across emitter

resolve: https://github.com/Azure/cadl-ranch/issues/666
resolve: https://github.com/Azure/cadl-ranch/issues/665


packages/cadl-ranch-specs/http/routes/main.tsp
```diff
@@ -0,0 +1,476 @@
+import "@typespec/http";
+import "@azure-tools/cadl-ranch-expect";
+
+using TypeSpec.Http;
+
+/**
+ * Define scenario in building the http route/uri
+ */
+@scenarioService("/routes")
+namespace Routes;
```
timotheeguerin, line 10: open to a different name here too

packages/cadl-ranch-specs/http/routes/main.tsp
```diff
@@ -0,0 +1,476 @@
...
+      Defines a path parameter that shouldn't encode reserved characters. It should however still encode the other url characters.
+        Param value: "foo/bar baz"
+        Expected path: "/routes/path/reserved-expansion/annotation/foo/bar%20baz"
+      """)
+    @route("annotation")
+    op annotation(@path(#{ allowReserved: true }) param: string): void;
+  }
+
+  @route("simple")
+  namespace Simple {
```
timotheeguerin, line 82: should maybe update all those namespace to be `XExpansion`

packages/cadl-ranch-specs/http/routes/main.tsp
```diff
@@ -0,0 +1,476 @@
...
+        Expected path: "/routes/path/reserved-expansion/template/foo/bar%20baz"
+      """)
+    @route("template/{+param}")
+    op template(param: string): void;
+
+    @scenario
+    @scenarioDoc("""
+      Defines a path parameter that shouldn't encode reserved characters. It should however still encode the other url characters.
+        Param value: "foo/bar baz"
+        Expected path: "/routes/path/reserved-expansion/annotation/foo/bar%20baz"
```
weidongxu-microsoft, line 75: What if user set param value `foo/bar%20baz`? Does it result in same path?

Reading the [RFC](https://datatracker.ietf.org/doc/html/rfc6570#section-3.2.3) but feeling unsure about such case.

timotheeguerin: If the param is already encoded it shouldn't reencode I can add scenarios for that

weidongxu-microsoft: No need a new test. Just want to be sure of "what char still need encode on Reserved Expansion case".

packages/cadl-ranch-specs/http/routes/main.tsp
```diff
@@ -0,0 +1,476 @@
...
+      op array(param: string[]): void;
+
+      @scenario
+      @scenarioDoc("""
+        Test simple expansion with explode: true when passed a record value.
+            Param value: {a: 1, b: 2}
+            Expected path: /routes/simple/explode/recorda=1,b=2
+        """)
+      @route("record{param*}")
+      op record(param: string[]): void;
```
weidongxu-microsoft, line 140: How's param value `{a: 1, b: 2}` (looks like a dict) matches the op `op record(param: string[])` (`param` an array)?

weidongxu-microsoft (commented):
LGTM except one question at https://github.com/Azure/cadl-ranch/pull/673#discussion_r1714619140

I didn't check the detailed mockapi impl. I guess the first dev impl these tests in emitter would take the runtime check.

packages/cadl-ranch-specs/http/routes/main.tsp
MaryGao: @timotheeguerin Could you help me understand the differences between the [x-ms-skip-url-encoding](https://github.com/Azure/autorest/blob/main/docs/extensions/readme.md#x-ms-skip-url-encoding) and [allowReserved](https://datatracker.ietf.org/doc/html/rfc6570#section-3.2.3) in uri template?

- x-ms-skip-url-encoding: Setting this as true would prevent the automatic encoding behavior for ALL characters
- allowReserved: Setting this as true would have RESERVED characters :/?#[]@!$&'()*+,;= in parameter values to be allowed to be sent as they are

My understanding is their scope for encoding or not is different. The former is for whole parameter value but latter is only working for reserved characters. If this is it, I would question on how to translate swagger extension to correct typespec.

```
path: template/{+param}

param value: foo/bar%20baz

x-ms-skip-url-encoding: template/foo/bar%20baz
allowReserved:  template/foo/bar%2520baz ==> pls note the % would also be pct-encoded as %25
```

Here is [one implementation](https://www.typescriptlang.org/play/?#code/GYVwdgxgLglg9mABAUwB4AcCGYAmBVAJxgBVkBbdAG0ymQApaLraAuRAZyiLAHMAaRADphAN0yUQydm2wBPANoBdAJRtO3HogDeAKET7ElZFERiJyAJK40iALyIADAG49BgsZAEkjKjWSD3Xwh6AHoAHS06eQBqAGJBEKcAfgAyRSTlOjokljCAdwAfAFJ5BwBaAE4AQTKAMUwy4EUtACYAXwL5QTCygD8AQgAqAHI6ZScWAAEU22iAEj4QpNiw+TDFFWjMnJY6MJxogrDB5WUksLaQ-kQ6MhoIAAsBOHRkAho4AgExIkwAIyMAjIcBwMGAMDeyjsAD5tK4DAYwTczJIrDgbND7CipIIjLwoA8oboESS3B4vIhhsMXKT9G14SSGQijCZsXZTOJJOx5Ni0WhotFFDTaUzEcBkZzkHZbPZwOjwWBkDgiaLSe4oJ4kFThaT6arEEjbiCwRCCCraSTDcDQeC3tL7MMRuaLbTDVUCO9ZIIYOx3Z66NjTnCXSGOeZ2djBAArOAwMB0F5vD4Ee2UkLDRBJNMZtjDPjDcb6i1tFCUdhSw1QWSvODitkyh1wP5R5DQAvB0MW+uIADyzdbUEEAGtkLJ2AHJcpBHd0HQhzDEAADOZaIdtWwr7HyIeKNqLqcxuN0YYpAs6zuIemdkvIMtS4kX-TdyNQOAAZS4cZ4Y0E7BAf3UL86AcAQsAIcsrCgI0bVNAQAEYHFOc9i31PULXYPIYCgR4bkTd5XzNDsQwgTBy0paJhhYItaXVTUUEgEFkAAJSkN4RCVCdzELTsSLI4ZYko6i1XJJBF1iFdkAY9EWPLAh2JwTjJGUPdkNpXipWGQRBMffRaIpRdBAkqTkDwJiLAAYTgCgEEkqDAxUoTEHU7MqJ0xA9NEkIjIgRiZLYjj7MXVTSWc6ltJ0jylycTdMF+AFkHXbzGNMiyrPQGywDsycHJ40iNKScLH0ixckhiuKjESrRJJ89EUss6zFUyxTkGUoLHNC09XIikSlxSMqYH+CqNyq4y6rSjKsq4nLQ3lTAQEoKAuqKnrquSsz6vSxrJqU4LLwZNpCz1UBIFgBB6Jq5jWLkgLJTUT9eFUDh7s0B9dJW0b1vGrbmqnQJqGCOgQnkfpRmUQZFCuAQIAXRcihXCBBEeWLLPRKooMQwRXw-DQ6DggA2KdXzwdBXgIcy8rGPdfuQIJQiKFpakh7NqdpwGigAZiqJnKILAIaf+umWnM7n80OnRxZCEJED866cEQABRDBsHYeAwB0HywHYOAjFxOBvzQLBcEIEhyF8Wg6AAIgeKAoHQFhJbQTAmH8HyyBCLRoh+S50SgTAYDLC2BAt4A4DgEIHlvSg4EQPJPkoHALaQxAdEl3sQFt9O2Gt237ZCR3ncRqyQhDsO-likIfb9stxY1rWdaj-WlaNohSCYPxLezu2HdQJ2qBdouPa9ivjCr9hA8QYPQ5CMuCHphwy4AL0T8Zk9Tnt0-QTPEE73P877wu3ZL8PI7gemAFYWgcWOCHj4fff99gdCAA) for uri template in JS and it do re-encode % twice. But I am not sure this is fully correct.

timotheeguerin: The question is more why do you actually want to complexly skip encoding. Most of those cases are because they include / and it still make sense to encode uri elements.

If you look at JS functions it’s the difference between encodeUri(encode elements that are not valid in uri) and encodeUriComponent(encode the reserved chars)

packages/cadl-ranch-specs/http/routes/main.tsp
timotheeguerin: more like should it be `SimpleExpansion` `ReservedExpansion` `PathExpansion` (the terms used in the Rfc)

packages/cadl-ranch-specs/http/routes/main.tsp
```diff
@@ -0,0 +1,476 @@
...
+            Param value: {a: 1, b: 2}
+            Expected path: /routes/label/explode/record.a=1.b=2
+        """)
+      @route("record{.param*}")
+      op record(param: string[]): void;
+    }
+  }
+
+  @route("matrix")
+  namespace Matrix {
```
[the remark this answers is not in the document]
timotheeguerin: it isn't because there is extra nesting and every scenario is just so slightly different that that main template would still need 7+ params to pass all the different input to output. Could make 2 reusable interface(explode and non explode) but it also feels like it might be more readable with everything explicit anyway here

packages/cadl-ranch-specs/http/routes/main.tsp
```diff
@@ -0,0 +1,476 @@
...
+        Test simple expansion with explode: true when passed a record value.
+            Param value: {a: 1, b: 2}
+            Expected path: /routes/simple/explode/recorda=1,b=2
+        """)
+      @route("record{param*}")
+      op record(param: Record<int32>): void;
+    }
+  }
+
+  @route("path")
```
MaryGao, line 144: @lmazuel just confirm the explode for path parameter supporting is not in this semaster's scope: https://github.com/Azure/typespec-azure/issues/1318

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -3572,6 +3572,366 @@ With the above two calls, we test the following configurations from this service
...
+
+Test label expansion with explode: true when passed an array value.
+Param value: ["a","b"]
+Expected path: /routes/label/explode/array.a.b
+
+### Routes_PathParameters_Label_Explode_primitive
+
+- Endpoint: `get /routes/path/label/explode/primitive{.param*}`
+
+Test label expansion with explode: true when passed a primitive value.
```
iscai-msft, line 3617: can we explode primitives? I thought explode had more to do with arrays and dicts

packages/cadl-ranch-specs/http/routes/main.tsp
iscai-msft: there's also overlap here with the parameters grouping of tests: I think it would be better to add these uri tests to existing `query`, `path` etc sections, what are people's thoughts

lmazuel: While technically I could write a full template with path and query at the same time which would make a generic "route" better, I feel you're right that we should probably split by path/query. I feel like the full template will be rare if happening at all, so I'm ok to keep it as a special case of "path"

packages/cadl-ranch-specs/http/routes/main.tsp
lmazuel: Let's case worst case of a customer passing an evil parameter (because if the parameter is well formed, it's doesn't really matter). If you skip all encoding, you likely will get a connection error (invalid url), or a 404 because the url doesn't make sense. If you escape special characters only, the url is always valid, but you will likely get a 404. In pragmatic terms, the different is not that big.
So I agree with Tim, that we could consider those equivalent, I dare you to find any cases where the user parameter would would have worked fine with no encoding, but would break if we encode some characters :)

packages/cadl-ranch-specs/http/routes/main.tsp
lmazuel: Yes, path explode is out of scope

packages/cadl-ranch-specs/cadl-ranch-summary.md
lmazuel: The url template spec accept primitive types with explode, though it's ignoring explode since there is nothing to explode. So it's fair to have a test that check we don't "explode" (pun intended)

packages/cadl-ranch-specs/http/routes/main.tsp
weidongxu-microsoft: > ```
> allowReserved:  template/foo/bar%2520baz ==> pls note the % would also be pct-encoded as %25
> ```

Another online tester (using [rezigned/uri-template.js](https://github.com/rezigned/uri-template.js)) also apply this logic
image

packages/cadl-ranch-specs/http/routes/main.tsp
timotheeguerin: well I think it is good to have all of uri template here, uri template kinda abstract away the fact that you have query and path parameter at this point, you just have params to build the uri.

I think it is nicer to have an overall view of what uri template is supported by emitters than having to fetch in every single category. I am however down to cleanup the other ones as support for uri template is added

iscai-msft: that's fair, thanks @timotheeguerin

iscai-msft (approved):

PR #674, opened 2024-08-08
Better warning for duplicate mockapi

timotheeguerin:


packages/cadl-ranch/src/scenarios-resolver.ts
```diff
@@ -173,13 +173,26 @@ export async function loadScenarioMockApis(scenariosPath: string): Promise<Recor
...
   for (const file of files) {
     for (const [key, scenario] of Object.entries(file.scenarios)) {
-      if (key in result) {
-        logger.warn(`Scenario ${key} is being defined twice.`);
+      if (duplicateTracker[key]) {
+        duplicateTracker[key].push(file.path);
+      } else {
+        duplicateTracker[key] = [file.path];
       }
+      duplicateTracker[key] = duplicateTracker[key] || [];
```
weidongxu-microsoft, line 184: Not sure if I understand this line. I thought `duplicateTracker[key]` is already updated on the lines above (both branching sets the `duplicateTracker[key]`?)

Not expert though. Feel free to ignore.

weidongxu-microsoft (approved):

packages/cadl-ranch/src/scenarios-resolver.ts
timotheeguerin: Yeah no a mistake

timotheeguerin, line 184:
```suggestion
      duplicateTracker[key] = duplicateTracker[key];
```

PR #681, opened 2024-08-12
Facet Test

mcgallan:
To migrate the [Models-TypeSpec](https://github.com/Azure/autorest.csharp/blob/68e44c384b4809bb4c1e989571ba19f081c3ff48/test/TestProjects/Models-TypeSpec/Models-TypeSpec.tsp#L533) from the autorest test project to cadl ranch, and specifically to handle the migration of the facet operation.

### DIFF
--- a/packages/cadl-ranch-specs/http/azure/core/facet/mockapi.ts
+++ b/packages/cadl-ranch-specs/http/azure/core/facet/mockapi.ts
@@ -1,4 +1,4 @@
-import { passOnSuccess, mockapi, json, ValidationError } from "@azure-tools/cadl-ranch-api";
+import { passOnSuccess, mockapi, json } from "@azure-tools/cadl-ranch-api";
 import { ScenarioMockApi } from "@azure-tools/cadl-ranch-api";
 
 export const Scenarios: Record<string, ScenarioMockApi> = {};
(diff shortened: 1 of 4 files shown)


timotheeguerin (commented):
what feasture of typespec is this testing

mcgallan:
> what feasture of typespec is this testing

This PR is to add tests for the facet operation.

weidongxu-microsoft (changes requested):
What is facet?

Why it need to be under azure/core? What Azure-specific feature it provides? Non-Azure cannot use this?

packages/cadl-ranch-specs/http/azure/core/facet/mockapi.ts
```diff
@@ -0,0 +1,23 @@
...
+  value: 15,
+  field: "price",
+};
+
+const responseBody = {
+  field_name: "price",
+  results: [priceResult],
+};
+
+Scenarios.Azure_Core_Facet_getInt32Facets = passOnSuccess(
```
iscai-msft, line 18: it's unclear what testing a facet is about testing. What are you trying to test in `azure.core`?

mcgallan: Regarding the facet related operations in Models-TypeSpec, I had some misunderstandings. There are actually no new features. I will close this PR

PR #694, opened 2024-08-14
Add test for pagination

melina5656:
This PR includes tests from the [https://github.com/Azure/autorest.csharp/blob/feature/v3/test/TestProjects/Pagination-TypeSpec/Pagination-TypeSpec.tsp](url). I have moved tests to http/azure/core/page.


packages/cadl-ranch-specs/http/azure/core/page/main.tsp
```diff
@@ -200,3 +201,411 @@ interface TwoModelsAsPageItem {
   @route("/second-item")
   listSecondItem is global.Azure.Core.Foundations.Operation<{}, global.Azure.Core.Page<SecondItem>>;
 }
+
+@resource("transactions")
+model LedgerEntry {
```
ArcturusZhang, line 206: could we generalize the name of the models here so that it does not look like being borrowed from a service?

ArcturusZhang: I think the `@resource` decorator is the place that makes this case unique from others?
If so let us keep this.
If not, we should not add irrelevant stuff on cadl ranch cases.

packages/cadl-ranch-specs/http/azure/core/page/main.tsp
```diff
@@ -200,3 +201,411 @@ interface TwoModelsAsPageItem {
...
+  /**
+   * The list of Pools.
+   */
+  @global.Azure.Core.items
+  value?: BatchPool[];
+
+  /**
+   * The URL to get the next set of results.
+   */
+  #suppress "@azure-tools/typespec-azure-core/casing-style" "The names of Property types must use camelCase"
```
ArcturusZhang, line 445: why we are having this linter suppression here? I did not see a violation here. `nextLink` here follows the rule

packages/cadl-ranch-specs/http/azure/core/page/main.tsp
```diff
@@ -200,3 +201,411 @@ interface TwoModelsAsPageItem {
...
+  @global.Azure.Core.nextLink
+  nextLink?: string;
+}
```
ArcturusZhang, line 448: is there any unique place in this pagination result? If no, we do not need to have this case.

packages/cadl-ranch-specs/http/azure/core/page/main.tsp
```diff
@@ -200,3 +201,411 @@ interface TwoModelsAsPageItem {
...
+  """)
+@route("/custom-resource-list")
+op ListCustomResource<
+  TResource extends {},
+  TCustom extends global.Azure.Core.Foundations.CustomizationFields = {}
+> is global.Azure.Core.Foundations.ResourceList<
+  TResource,
+  global.Azure.Core.Foundations.CustomParameters<TCustom>,
+  global.Azure.Core.Page<TResource>
+>;
```
ArcturusZhang, line 599: if this template is only used in one place, we do not need to have it. It reduces the readability.
Please inline it or we could try to make abstraction to all the new operations into one template - I see all the new operations are trying to list a resource, maybe we could do it.

packages/cadl-ranch-specs/http/azure/core/page/main.tsp
```diff
@@ -200,3 +201,411 @@ interface TwoModelsAsPageItem {
...
+    ]
+  }
+  """)
+@route("/ledger-entries")
+@put
+op ListPaginationLedgerEntries is global.Azure.Core.Foundations.ResourceList<
+  LedgerEntry,
+  ListPaginationLedgerEntryInput,
+  PagedLedgerEntries
+>;
```
weidongxu-microsoft, line 291: Why we have PUT for pagination in Azure?

weidongxu-microsoft (changes requested):

packages/cadl-ranch-specs/http/azure/core/page/mockapi.ts
```diff
@@ -50,3 +50,75 @@ Scenarios.Azure_Core_Page_listWithCustomPageModel = passOnSuccess(
...
+Scenarios.Azure_Core_Page_UseFoundationsResourceList = passOnSuccess(
+  mockapi.get("/azure/core/page/foundations-resource-list/custom-resource-list/transactions", (req) => {
+    const responseBody = {
+      transactionId: "1",
+      collectionId: "1",
+      contents: "contents",
+    };
+    return { status: 200, body: json(responseBody) };
+  }),
+);
```
weidongxu-microsoft, line 124: I don't see any element in response. Why is this pagination?

packages/cadl-ranch-specs/http/azure/core/page/main.tsp
pshao25: It is under the `azure` folder so I think it is more appropriate to keep the `@resource`.

packages/cadl-ranch-specs/http/azure/core/page/main.tsp
```diff
@@ -200,3 +201,392 @@ interface TwoModelsAsPageItem {
...
+@scenario
+@scenarioDoc("""
+  List the dimension values for the given metric dimension name.
+  """)
+interface AdditionalParameter {
+  /**
+   * List the dimension values for the given metric dimension name.
+   */
+  @route("/metric-dimensions")
+  ListMetricDimensionValues is global.Azure.Core.ResourceList<
```
iscai-msft, line 376: can these names be more clear? it's important that the coverage scenario clearly tells what this is testing. `MetricDimensionValues` doesn't have much meaning. For each of these operations, can you name them so we can explicitly see what we're testing that we didn't test before?

packages/cadl-ranch-specs/http/azure/core/page/main.tsp
```diff
@@ -200,3 +201,392 @@ interface TwoModelsAsPageItem {
...
+   */
+  @maxLength(128)
+  text: string;
+}
+
+@scenario
+@scenarioDoc("""
+  This scenario is to demonstrate an interface with two operations, each returning a list of different resource types.
+  """)
+interface TwoResourcesAsListItems {
```
pshao25, line 540: We already have this test.

pshao25: The whole interface can be removed.

packages/cadl-ranch-specs/http/azure/core/page/main.tsp
```diff
@@ -200,3 +201,392 @@ interface TwoModelsAsPageItem {
...
+  List the dimension values for the given metric dimension name.
+  """)
+interface AdditionalParameter {
+  /**
+   * List the dimension values for the given metric dimension name.
+   */
+  @route("/metric-dimensions")
+  ListMetricDimensionValues is global.Azure.Core.ResourceList<
+    DimensionValueListItem,
+    global.Azure.Core.Traits.QueryParametersTrait<MetricDimensions>
```
pshao25, line 378: You could call it `listWithTraits`. And change the model to a more generic name.

packages/cadl-ranch-specs/http/azure/core/page/main.tsp
```diff
@@ -200,3 +201,392 @@ interface TwoModelsAsPageItem {
...
+@scenario
+@scenarioDoc("""
+  Lists all of the Pools in the specified Account.
+  """)
+interface Pools {
+  /**
+   * Lists all of the Pools in the specified Account.
+   */
+  @route("/pools")
+  listPools is global.Azure.Core.RpcOperation<
```
pshao25, line 481: Is that correct syntax?

packages/cadl-ranch-specs/http/azure/core/page/main.tsp
```diff
@@ -200,3 +201,93 @@ interface TwoModelsAsPageItem {
   @route("/second-item")
   listSecondItem is global.Azure.Core.Foundations.Operation<{}, global.Azure.Core.Page<SecondItem>>;
 }
+
+/**
+ * The interval
+ */
+union TimeInterval {
```
iscai-msft, line 208: i don't think we need most of these models anymore

.changeset/spicy-trees-know.md
```diff
@@ -0,0 +1,9 @@
+---
+"@azure-tools/cadl-ranch-specs": patch
+---
+
+Add Scenarios.Azure_Core_Page_ListPaginationLedgerEntries test scenarios and corresponding Mock API implementations for azure/core/page.
```
iscai-msft, line 5: can you update the changeset to the test you're adding?

packages/cadl-ranch-specs/http/azure/core/page/main.tsp
```diff
@@ -200,3 +201,93 @@ interface TwoModelsAsPageItem {
...
+@scenario
+@scenarioDoc("""
+  List the item values.
+  """)
+interface AdditionalParameter {
+  /**
+   * List the item values.
+   */
+  @route("/items-traits")
+  listWithTraits is global.Azure.Core.ResourceList<
```
iscai-msft, line 289: I think we already have a very similar test in `azure/core/basic` https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/azure/core/basic/main.tsp#L178. Can you move that test instead into `azure/core/page`? Thank you!

packages/cadl-ranch-specs/http/azure/core/basic/main.tsp
```diff
@@ -145,45 +145,6 @@ op createOrReplace is ResourceOperations.ResourceCreateOrReplace<User>;
...
-  """)
-op list is ResourceOperations.ResourceList<
-  User,
-  ListQueryParametersTrait<global.Azure.Core.StandardListQueryParameters &
-    global.Azure.Core.OrderByQueryParameter &
-    global.Azure.Core.FilterQueryParameter &
-    global.Azure.Core.SelectQueryParameter &
-    global.Azure.Core.ExpandQueryParameter>
->;
-
```
weidongxu-microsoft, line 186: please put it back.

List is a basic op in azure.

packages/cadl-ranch-specs/http/azure/core/page/main.tsp
weidongxu-microsoft: I'd against moving the test. List is one of the basic CRUD of any Azure resource. We already had it in basic for months. There is no point moving this test, just for the sake of moving...

PR #699, opened 2024-08-14
Add cadl-ranch-api to changeset for header validation fix

JoshLove-msft:


weidongxu-microsoft (approved):
Not sure if we want to remove https://github.com/Azure/cadl-ranch/blob/main/.changeset/wet-apes-laugh.md

tadelesh:
> Not sure if we want to remove https://github.com/Azure/cadl-ranch/blob/main/.changeset/wet-apes-laugh.md

yes. i left a comment for @JoshLove-msft in teams.

Issue #689, opened 2024-08-13
ARM, test case for `ExtensionResource`


XiaofeiCao:
doc: https://azure.github.io/typespec-azure/docs/howtos/ARM/resource-type#extension-resource

example: https://github.com/Azure/typespec-azure/blob/main/packages/samples/specs/resource-manager/resource-types/extension/main.tsp#L29-L31

add `extension.tsp` under https://github.com/Azure/cadl-ranch/tree/main/packages/cadl-ranch-specs/http/azure/resource-manager/models/resources

PR #701, opened 2024-08-15
Tests ARM, test case for ExtensionResource

v-hongli1:
- https://github.com/Azure/cadl-ranch/issues/689


packages/cadl-ranch-specs/http/azure/resource-manager/models/resources/mockapi.ts
```diff
@@ -42,6 +42,162 @@ const validNestedResource = {
...
+    createdBy: "AzureSDK",
+    createdByType: "User",
+    createdAt: new Date(),
+    lastModifiedBy: "AzureSDK",
+    lastModifiedAt: new Date(),
+    lastModifiedByType: "User",
+  },
+};
+
+const validExtensionResources = [
```
XiaofeiCao, line 63: Why there are two `validExtensionResources`?

packages/cadl-ranch-specs/http/azure/resource-manager/models/resources/mockapi.ts
```diff
@@ -42,6 +42,162 @@ const validNestedResource = {
   },
 };
 
+const validExtensionResource = {
+  id: `/subscriptions/${SUBSCRIPTION_ID_EXPECTED}/resourceGroups/${RESOURCE_GROUP_EXPECTED}/providers/Azure.ResourceManager.Models.Resources/extensionTrackedResources/extension`,
+  name: "roleAssign",
```
XiaofeiCao, line 47: `name` property value(`roleAssign`) and the one in the path(`extension`) don't match. Maybe just `extension`.

packages/cadl-ranch-specs/http/azure/resource-manager/models/resources/extension.tsp
```diff
@@ -0,0 +1,174 @@
...
+using TypeSpec.Http;
+using TypeSpec.Rest;
+using Azure.Core;
+using Azure.ResourceManager;
+using TypeSpec.OpenAPI;
+
+namespace Azure.ResourceManager.Models.Resources;
+
+@resource("extensionResources")
+model RoleAssignment is ExtensionResource<RoleAssignmentProperties> {
```
XiaofeiCao, line 15: Better to be consistent here, e.g. `@resource("roleAssignments")`

v-hongli1: Fixed in the new version.

packages/cadl-ranch-specs/http/azure/resource-manager/models/resources/mockapi.ts
```diff
@@ -61,7 +199,7 @@ Scenarios.Azure_ResourceManager_Models_Resources_TopLevelTrackedResources_action
         urgent: true,
       });
       return {
-        status: 204,
+        status: 200,
```
XiaofeiCao, line 202: Reason?

v-hongli1: This return status code has been rollbacked.

packages/cadl-ranch-specs/http/azure/resource-manager/models/resources/mockapi.ts
```diff
@@ -42,6 +42,144 @@ const validNestedResource = {
   },
 };
 
+const validExtensionResource = {
+  id: `/subscriptions/${SUBSCRIPTION_ID_EXPECTED}/resourceGroups/${RESOURCE_GROUP_EXPECTED}/providers/Azure.ResourceManager.Models.Resources/roleAssignments/role`,
+  name: "role",
+  type: "extensionResource",
```
XiaofeiCao, line 48: type is wrong

check https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/azure/resource-manager/models/resources/mockapi.ts#L11

v-hongli1: Fixed in the new version.

packages/cadl-ranch-specs/http/azure/resource-manager/models/resources/extension.tsp
```diff
@@ -0,0 +1,174 @@
...
+        "createdByType": "User",
+        "createdAt": <any date>,
+        "lastModifiedBy": "AzureSDK",
+        "lastModifiedAt": <any date>,
+        "lastModifiedByType": "User",
+      }
+    }
+    ```
+    """)
+  get is ArmResourceRead<RoleAssignment>;
```
iscai-msft, line 58: does the fact that it's a `RoleAssignment` object matter? With the scenario names we're really trying to convey what exactly we're testing. Right now, we're clearly testing `RoleAssignment`, which isn't a model from mgmt core, so I just want to make sure it's necessary that the scenario name includes `RoleAssignment`

XiaofeiCao: It's defined by us and just an `ExtensionResource`: https://github.com/Azure/cadl-ranch/pull/701/files#diff-caceefb500557a9f283bd4399e6ecc73278b9008e9b10b129c62403ea81a76cfR15

Agree that the name should be more generic than `RoleAssignment`. @v-hongli1 Please rename it to something else, e.g. `CustomExtensionResource`.

v-hongli1: Fixed in the new version.

packages/cadl-ranch-specs/http/azure/resource-manager/models/resources/extension.tsp
```diff
@@ -0,0 +1,179 @@
...
+  provisioningState?: ProvisioningState;
+}
+
+@armResourceOperations
+interface CustomExtensionResources {
+  @scenario
+  @scenarioDoc("""
+    Resource GET operation.
+    Expected path: /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/test-rg/providers/Azure.ResourceManager.Models.Resources/customExtensionResources/extension
+    Expected query parameter: api-version=2023-12-01-preview
```
tadelesh, line 40: is it an extension resource?

v-hongli1: @tadelesh
I think this is an  extension resource, do you have any better ideas?

XiaofeiCao: Extension resource is an extension of another resource: https://learn.microsoft.com/sr-latn-rs/azure/azure-resource-manager/management/extension-resource-types

@v-hongli1 See if we could modify the path to be the extension of the tracked resource, e.g.
`/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/test-rg/providers/Azure.ResourceManager.Models.Resources/topLevelTrackedResources/top/customExtensionResources/extension`

packages/cadl-ranch-specs/http/azure/resource-manager/models/resources/extension.tsp
```diff
@@ -11,7 +11,7 @@ using TypeSpec.OpenAPI;
 
 namespace Azure.ResourceManager.Models.Resources;
 
-@resource("customExtensionResources")
+@parentResource(TopLevelTrackedResource)
```
XiaofeiCao, line 14: We don't need @parentResource, extension resource can have more than 1 parent type.
Just change the mockapi.ts to accept top-level resource as its parent(and in your test for generated code in autorest.java, pass in the top-level resource id), and it should be fine.

v-hongli1: @XiaofeiCao 
If you don't use `@parentResource`, then there must be a parameter in Autorest.java that accepts the top-level resource id, but I haven't been able to find a `Resource` or `Parameter` that I can use.

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -907,6 +907,156 @@ Expected response body:
...
+- Endpoint: `delete https://management.azure.com`
+
+Resource DELETE operation.
+Expected path: /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/test-rg/providers/Azure.ResourceManager.Models.Resources/topLevelTrackedResources/top/customExtensionResources/extension
+Expected query parameter: api-version=2023-12-01-preview
+
+````
+Expected response status code: 204
+
+### Azure_ResourceManager_Models_Resources_CustomExtensionResources_get
```
iscai-msft, line 960: I think we should correspond the scenario names 1:1 with the exact ARM models names, so `Azure_ResourceManager_Models_Resources_Extensions_get` / `Azure_ResourceManager_Models_Resources_ExtensionsResources_get`

XiaofeiCao: Makes sense. We should call it `ExtensionResources`.

packages/cadl-ranch-specs/http/azure/resource-manager/models/resources/extension.tsp
```diff
@@ -0,0 +1,174 @@
...
+  @doc("The description of the resource.")
+  description?: string;
+
+  /** The status of the last operation. */
+  @visibility("read")
+  provisioningState?: ProvisioningState;
+}
+
+@armResourceOperations
+interface CustomExtensionResources {
```
XiaofeiCao, line 30: Change to `ExtensionResources`.

v-hongli1: Fixed in the new version https://github.com/Azure/cadl-ranch/pull/701/commits/3399f26b6ea8081e0519ef28c9521392e0769724

packages/cadl-ranch-specs/http/azure/resource-manager/models/resources/mockapi.ts
```diff
@@ -155,6 +173,141 @@ Scenarios.Azure_ResourceManager_Models_Resources_SingletonTrackedResources_listB
   ),
 ]);
 
+Scenarios.Azure_ResourceManager_Models_Resources_ExtensionsResources_get = passOnSuccess([
+  mockapi.get(
+    "/subscriptions/:subscriptionId/resourceGroups/:resourceGroup/topLevelTrackedResources/:topLevelResourceName/providers/Azure.ResourceManager.Models.Resources/extensionsResources/:extensionName",
```
XiaofeiCao, line 178: Will this path work?

v-hongli1: Yes, the paths of  `ExtensionResource` in `mockapi.tsp` has been verified and works fine.

XiaofeiCao: Thanks, this means our code doesn't have encoding issue with `resourceUri`, is that right?

Another thing is, where did you get the value to pass into the `resourceUri`? Seems it lacks namespace `Azure.ResourceManager.Models.Resources` for topLevelTrackedResource...

XiaofeiCao: @v-hongli1 check the top level resource id: https://github.com/Azure/cadl-ranch/blob/4022f01c7cebed9df15ba36d591fb5784a6b23ff/packages/cadl-ranch-specs/http/azure/resource-manager/models/resources/toplevel.tsp#L48

The one you passed in is wrong.

v-hongli1: 1. First, resourceUri still has encoding issues
2. Secondly, the top level resource id is also passed through resourceUri.
image

XiaofeiCao: Got it.

For 1, we should wait for the fix before merging this PR.

For 2, The value is not correct. It should be: https://github.com/Azure/cadl-ranch/blob/4022f01c7cebed9df15ba36d591fb5784a6b23ff/packages/cadl-ranch-specs/http/azure/resource-manager/models/resources/toplevel.tsp#L48

v-hongli1: @XiaofeiCao 
Use `@parentResource` or `@resource("topLevelTrackedResources/top/extensionsResources")`?

XiaofeiCao: No. I mean the one in your mockapi.ts(same as the one you passed to `.withExistingResourceUri()` in your `testExtensionResource()`) is wrong.

Use `/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/test-rg/providers/Azure.ResourceManager.Models.Resources/topLevelTrackedResources/top`.

v-hongli1: Fixed by commit https://github.com/Azure/cadl-ranch/pull/701/commits/7f74cde43b5bb7830af8575e11e3d5b9743c57c6

packages/cadl-ranch-specs/http/azure/resource-manager/models/resources/extension.tsp
```diff
@@ -0,0 +1,173 @@
...
+  @visibility("read")
+  provisioningState?: ProvisioningState;
+}
+
+@armResourceOperations
+interface ExtensionsResources {
+  @scenario
+  @scenarioDoc("""
+    Resource GET operation.
+    Expected path: /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/test-rg/providers/Azure.ResourceManager.Models.Resources/topLevelTrackedResources/top/providers/Azure.ResourceManager.Models.Resources/extensionsResources/extension
```
ArcturusZhang, line 33: is it possible that we could construct a path like
```
/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/test-rg/providers/Azure.ResourceManager.Models.Resources/extensionsResources/extension
```
via the extension resource feature?

ArcturusZhang: as far as I know, the `ExtensionResource` model is like a "scope" resource which generates the resource like:
```
/{resourceUri}/providers/{namespace}/foos/{name}
```
and since its resourceUri could be any valid existing resource, it could be a subscription or resource group, we should add some verification for these scenario as well

XiaofeiCao: With existing feature, we could, at least for Java. Though it's not a valid extension resource.

packages/cadl-ranch-specs/http/azure/resource-manager/models/resources/mockapi.ts
XiaofeiCao: @v-hongli1 , after I offline discussed with chenjie, we should make mockapi return double "//".

packages/cadl-ranch-specs/http/azure/resource-manager/models/resources/extension.tsp
XiaofeiCao: > it could be a subscription or resource group, we should add some verification for these scenario as well

Yes, you are right! We could add case for tenant, subscription, resource group and normal resources.

XiaofeiCao: @v-hongli1 Consider adding scenarios for another two scopes: subscription and resource groups.

You probably need:

1. Refer below https://github.com/Azure/cadl-ranch/blob/c5f897d13996f5de2a078ddf8f127570bb59fead/packages/cadl-ranch-specs/http/azure/client-generator-core/access/mockapi.ts#L35-L54

for adding more than 1 mockapis to one scenario.

2. In extension.tsp, update each `@scenarioDoc` to indicate that three expected scopes are needed to pass this scenario test.

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -907,6 +907,156 @@ Expected response body:
 }
 ```
 
+### Azure_ResourceManager_Models_Resources_ExtensionsResources_createOrUpdate
+
```
iscai-msft, line 911: nit: do we need `ExtensionsResources`, or is just `Extensions` ok bc it's in the `Resources` subspace

XiaofeiCao: `Extension` is good. 

Glad you brought it up. In fact, I revisited our existing ARM cases:
Screenshot 2024-08-29 at 12 25 21

Some of them contain duplicate like this, and some not appropriate, e.g. `ManagedIdentityTrackedResources`..
I've made a refactor PR to decouple SDK generated interface name with the scenarioName(it shouldn't affect existing generated SDKs):
https://github.com/Azure/cadl-ranch/pull/714

iscai-msft: Sorry I'm confused why you would need to use `clientName`. Can you clarify what is failing / needing extra tests because that shouldn't be happening. [Here's](https://github.com/Azure/cadl-ranch/pull/150/files) an example of a renaming pr, if that helps

v-hongli1: Fixed by commit https://github.com/Azure/cadl-ranch/pull/701/commits/8ca02c564ae744600ca1d7dbfba100606086d852

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -907,6 +907,156 @@ Expected response body:
...
+}
+```
+
+Expected response body:
+
+```json
+{
+  "id": "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/test-rg/providers/Azure.ResourceManager.Models.Resources/topLevelTrackedResources/top/providers/Azure.ResourceManager.Models.Resources/extensionsResources/extension",
+  "name": "extension",
+  "type": "extension",
```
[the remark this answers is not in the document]
v-hongli1: Fixed by commit https://github.com/Azure/cadl-ranch/commit/8ca02c564ae744600ca1d7dbfba100606086d852

packages/cadl-ranch-specs/http/azure/resource-manager/models/resources/extension.tsp
v-hongli1: Fixed by commit https://github.com/Azure/cadl-ranch/commit/8ca02c564ae744600ca1d7dbfba100606086d852

packages/cadl-ranch-specs/cadl-ranch-summary.md
XiaofeiCao: Thanks, it's mainly because that I don't want to break existing generated SDKs from current cadl-ranch. We already have several languages generated test for them. 
Wonder if there's better way to achieve that rather than `@clientName`..

packages/cadl-ranch-specs/http/azure/resource-manager/models/resources/extension.tsp
```diff
@@ -26,19 +28,20 @@ model ExtensionsResourceProperties {
 }
 
 @armResourceOperations
-interface ExtensionsResources {
+@clientName("ExtensionsResources")
```
XiaofeiCao, line 31: This is new test, guess we don't need `@clientName`?

v-hongli1: Rollback by [commit#7331206](https://github.com/Azure/cadl-ranch/pull/701/commits/7331206a2fa5cd52817a57f888095b9ea497821b) and [commit#69a4d5b](https://github.com/Azure/cadl-ranch/pull/701/commits/69a4d5b2fec7eaeccfe4dd706865675d501283c9)

packages/cadl-ranch-specs/http/azure/resource-manager/models/resources/extension.tsp
```diff
@@ -0,0 +1,173 @@
...
+  @visibility("read")
+  provisioningState?: ProvisioningState;
+}
+
+@armResourceOperations
+interface ExtensionsResources {
+  @scenario
+  @scenarioDoc("""
+    Resource GET operation.
+    Expected path: /<resourceUri>/providers/Azure.ResourceManager.Models.Resources/extensionsResources/extension
```
tadelesh, line 33: you need to give the full test path that emitter could know how to create this resource

packages/cadl-ranch-specs/cadl-ranch-summary.md
iscai-msft: In cases like these, we should go ahead and break. To do so, make sure everyone is aware of this issue, and that the PR has a minor version bump. Writing a mitigation guide (can just be a few lines: change x import to x import) is also very welcome.

packages/cadl-ranch-specs/http/azure/resource-manager/models/resources/extension.tsp
v-hongli1: Fixed by commit [9103aac](https://github.com/Azure/cadl-ranch/pull/701/commits/9103aac3c596fbcec47a70c86aec3c1c60dd5628)

packages/cadl-ranch-specs/cadl-ranch-summary.md
XiaofeiCao: @iscai-msft May I know the concern of `@clientName`? There's one more potential use case: https://github.com/Azure/cadl-ranch/pull/714#discussion_r1740296189

iscai-msft: somehow I totally missed this comment, my bad. The concern is that we're using `@clientName` to get around breaking cadl-ranch, and we're ok breaking cadl-ranch as long as everyone is aware. It adds technical debt and also makes the tests not as atomic: we aren't explicitly testing `@clientName` here so we shouldn't be using it

PR #704, opened 2024-08-21
Fix test case "client-operation-group" for python

msyyc:

### DIFF
--- a/packages/cadl-ranch-specs/http/client/structure/client-operation-group/client.tsp
+++ b/packages/cadl-ranch-specs/http/client/structure/client-operation-group/client.tsp
@@ -54,14 +54,14 @@ namespace Client.Structure.ClientOperationGroup {
     name: "SubNamespace.SecondClient",
     service: Client.Structure.Service,
   },
-  "csharp,java,python"
+  "csharp,java"
 )
 @client(
   {
     name: "SecondClient",
     service: Client.Structure.Service,
   },
-  "javascript,go"
+  "javascript,go,python"
 )
 namespace Client.Structure.AnotherClientOperationGroup {
   op five is Client.Structure.Service.Bar.five;
--- a/.changeset/hungry-cherries-fix.md
+++ b/.changeset/hungry-cherries-fix.md
@@ -1,0 +1,5 @@
+---
+"@azure-tools/cadl-ranch-specs": patch
+---
+
+Fix test case "client-operation-group" for python


packages/cadl-ranch-specs/http/client/structure/client-operation-group/client.tsp
```diff
@@ -54,14 +54,14 @@ namespace Client.Structure.ClientOperationGroup {
...
-  "csharp,java,python"
+  "csharp,java"
 )
 @client(
   {
     name: "SecondClient",
     service: Client.Structure.Service,
   },
-  "javascript,go"
+  "javascript,go,python"
```
msyyc, line 64: Python SDK doesn't permit client name has namespace prefix.

msyyc: @tadelesh @iscai-msft please help review

weidongxu-microsoft (approved):

packages/cadl-ranch-specs/http/client/structure/client-operation-group/client.tsp
tadelesh: > Python SDK package name, namespace and package folder are one-one mapping. If add sub-namespace, package folder shall also have related change which makes the folder complicated to understand for users.

do you have a link to the generated code?

msyyc: Without this PR, python emitter will crash for this case.

tadelesh: approve this change, but i believe we need to discuss this scenario in python.

tadelesh (approved):

packages/cadl-ranch-specs/http/client/structure/client-operation-group/client.tsp
iscai-msft: this is related to the overarching namespace changes we're looking into, see @srnagar 's proposal [here](https://gist.github.com/srnagar/a695faa1a4d8a5dbf44af826f05916f0). Python should start following this structure, which is something that all languages will start working on

msyyc: I see and create https://github.com/Azure/autorest.python/issues/2795 to track

PR #706, opened 2024-08-22
Remove inapplicable logic for dashboard

pshao25:
As titled.


pshao25:
> @pshao25 can you clarify why you're removing this code? Is it because we won't get report docs with the name `${generatorName}/reports/${index.version}.json` anymore?

Correct.

Those days when we only have one mode "azure", the reports store there. But now we have two modes, "azure" and "non-azure", the reports are stored in `${generatorName}/reports/${index.version}/${generatorMode}.json`.

We currently have some problem in the dashboard by the current logic "if old path exists then retrieve from old path, else retrieve from new path" because we somehow pushed an old version by accident. Since we don't need the old path, it is safe to remove. What's more I think we could also remove all the data in `${generatorName}/reports/${index.version}.json` unless we want an archive.

tadelesh (approved):

PR #707, opened 2024-08-22
Type-Model-Visibility Test

mcgallan:
The previous version of VisibilityModel had many properties with different visibilities in the same model, making it difficult to conduct precise tests in autorest.csharp. Therefore, in this PR, I have split these properties to make the structure clearer and facilitate easier testing.


packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -5596,7 +5596,7 @@ Expected response body:
 
 - Endpoint: `head /type/model/visibility`
 
-Generate abd send put model with write/create properties.
+Generate abd send put model with query properties.
```
iscai-msft, line 5599: based off of the url, it seems more that we are testing `head` calls, not the query properties

packages/cadl-ranch-specs/http/type/model/visibility/main.tsp
```diff
@@ -59,14 +82,14 @@ model ReadOnlyModel {
   ```
   """)
 @get
-op getModel(@body input: VisibilityModel): {
+op getModel(@body input: QueryModel): {
```
iscai-msft, line 85: if we want to test query parameters with readonly, I think we should be more clear and add another test for it, that way the scenario name is accurate

packages/cadl-ranch-specs/cadl-ranch-summary.md
mcgallan: This is an operation left over from two years ago. The visibility folder mainly tests the impact of @visibility on properties in the model and should not include tests for head, I will remove this operation to ensure clear functional separation.

packages/cadl-ranch-specs/http/type/model/visibility/main.tsp
mcgallan: Operation GetModelWithQuery() Added.

packages/cadl-ranch-specs/http/type/model/visibility/main.tsp
```diff
@@ -59,23 +82,29 @@ model ReadOnlyModel {
...
-    queryProp: 123,
+    readProp: "abc",
   }
   ```
   """)
-@head
-op headModel(@body input: VisibilityModel): OkResponse;
+@get
+@route("/getwithquery")
+op getModelWithQuery(@query queryProp: int32): {
```
iscai-msft, line 104: I don't think there's a good reason to remove `headModel`, we can just add `getModelWithQuery` as a separate test

packages/cadl-ranch-specs/cadl-ranch-summary.md
iscai-msft: I'm assuming the old test still serves it's purpose bc we are testing the affect of `head` on visibility. Additionally, removing it will be a breaking change which will have to be communicated to all emitter authors. I would recommend just adding your test as a separate test

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -5562,7 +5562,7 @@ Send a GET request which return the following body {requiredProp: "example-value
 
 - Endpoint: `delete /type/model/visibility`
 
-Generate abd send put model with write/create properties.
+Generate abd send put model with delete properties.
```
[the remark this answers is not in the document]
mcgallan: fixed

packages/cadl-ranch-specs/http/type/model/visibility/main.tsp
mcgallan: I have expanded the types of visibility properties in headmodel and added them back. Perhaps this will be more appropriate.

packages/cadl-ranch-specs/http/type/model/visibility/main.tsp
```diff
@@ -7,25 +7,67 @@ using TypeSpec.Http;
...
+
+  @doc("Required int32[], illustrating a update property.")
+  @visibility("update")
+  updateProp: int32[];
+}
+
+@doc("Input model with visibility properties.")
+model HeadVisibilityModel {
+  @doc("Required int32, illustrating a query property.")
+  @visibility("query")
```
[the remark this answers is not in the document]
mcgallan: @iscai-msft  This model is the input model for the operation `headModel`. However, because the name `VisibilityModel` was already taken by another model, it was named `HeadVisibilityModel`. Perhaps naming it `HeadInputModel` or `HeadModel` would be more appropriate. I apologize for any confusion caused by the name.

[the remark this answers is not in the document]
iscai-msft: I feel the point of this test is to have this one model that has all of these different visibilities, and examining how it behaves in ops with different methods. I would suggest adding the new visibility testing onto the existing model, and just adding properties for the visibilities you want to test, which I think is only the `@visibility("query")`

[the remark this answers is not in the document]
mcgallan: @iscai-msft The original purpose of the test code might be as you mentioned, but there have been some problems in the actual generation of autorest.csharp. Due to the various visibility parameters included in the VisibilityModel, the public constructor requires passing parameters for all properties, including those not needed in the mock API's expected body. 
image
This results in a mismatch between the provided body and the expected body. Therefore, a better approach is needed to handle this issue. Now, it is unclear whether modifying the parameters to add optional attributes, allowing us to choose which visibility parameters to assign, or splitting the VisibilityModel for testing is the better solution.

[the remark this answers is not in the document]
iscai-msft: we want to use this test though to determine what we will generate under these circumstances. it's up to csharp to decide how they want to generate params with visibility. It doesn't mean you have to follow 1-to-1 the typespec, but there should be logic here determining what csharp will do when they see a visibility "query" etc.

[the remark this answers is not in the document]
mcgallan: It appears that the main reason is that C# currently does not seem to support visibility-related features. Although the @visibility("read") related tests I previously attempted to add were able to run successfully, the actual results indicate that C# currently does not support properties with different visibilities exhibiting corresponding behaviors in different method operations. These tests will need to wait until C# supports visibility before they can be added.

PR #708, opened 2024-08-22
Etag new header `if-modified-since` and `if-unmodified-since`

msyyc:
with reference of real service https://github.com/Azure/azure-rest-api-specs/blob/2396cc5957cc3a286a0647cee07d9df286b11e54/specification/batch/Azure.Batch/common.tsp#L246-L264

### DIFF
--- a/packages/cadl-ranch-specs/http/special-headers/conditional-request/main.tsp
+++ b/packages/cadl-ranch-specs/http/special-headers/conditional-request/main.tsp
@@ -53,9 +53,9 @@ op postIfNoneMatch(
   Expected header parameters:
   - if-modified-since=Fri, 26 Aug 2022 14:38:00 GMT
   """)
-@post
+@head
 @route("/if-modified-since")
-op postIfModifiedSince(
+op headIfModifiedSince(
   @doc("""
     A timestamp indicating the last modified time of the resource known to the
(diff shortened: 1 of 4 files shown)


packages/cadl-ranch-specs/http/special-headers/conditional-request/main.tsp
```diff
@@ -43,3 +43,47 @@ op postIfNoneMatch(
...
+  Check when only If-Modified-Since in header is defined.
+  """)
+@scenarioDoc("""
+  Check when only If-Modified-Since in header is defined.
+  Expected header parameters:
+  - if-modified-since=Fri, 26 Aug 2022 14:38:00 GMT
+  """)
+@post
+@route("/if-modified-since")
+op postIfModifiedSince(
```
XiaofeiCao, line 58: Seems `if-modified-since` header can only appear in `GET` or `HEAD`:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since

msyyc: Updated to use `HEAD`

XiaofeiCao (approved):

weidongxu-microsoft (approved):

tadelesh (approved):

PR #711, opened 2024-08-23
refactor(spec): add uint8 for int encoding and templatize

archerzz:
- .net doesn't support uint32 yet, so use uint8
- use a value which will be negative in signed int

### DIFF
--- a/packages/cadl-ranch-specs/cadl-ranch-summary.md
+++ b/packages/cadl-ranch-specs/cadl-ranch-summary.md
@@ -2413,7 +2413,7 @@ Expected response body:
 }
 ```
 
-### Encode_Numeric_Property_uint8AsStringOptional
+### Encode_Numeric_Property_uint8AsString
 
 - Endpoint: `post /encode/numeric/property/uint8`
 
--- a/packages/cadl-ranch-specs/cadl-ranch-summary.md
+++ b/packages/cadl-ranch-specs/cadl-ranch-summary.md
@@ -2246,16 +2246,16 @@ Expected response body:
 }
 ```
 
-### Encode_Numeric_Property_uint32AsStringOptional
+### Encode_Numeric_Property_uint8AsStringOptional
 
-- Endpoint: `post /encode/numeric/property/uint32`
+- Endpoint: `post /encode/numeric/property/uint8`
 
-Test operation with request and response model contains property of uint32 type with string encode.
+Test operation with request and response model contains property of uint8 type with string encode.
 Expected request body:
 
 ```json
 {
-  "value": "1"
+  "value": "255"
 }
 ```
 
@@ -2263,7 +2263,7 @@ Expected response body:
 
 ```json
 {
-  "value": "1"
+  "value": "255"
 }
 ```
 
--- a/packages/cadl-ranch-specs/cadl-ranch-summary.md
+++ b/packages/cadl-ranch-specs/cadl-ranch-summary.md
@@ -2246,6 +2246,27 @@ Expected response body:
 }
 ```
 
+### Encode_Numeric_Property_uint32AsStringOptional
+
+- Endpoint: `post /encode/numeric/property/uint32`
+
+Test operation with request and response model contains property of uint32 type with string encode.
+Expected request body:
+
+```json
+{
+  "value": "1"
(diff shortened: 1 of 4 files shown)


packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -2246,24 +2246,24 @@ Expected response body:
...
+- Endpoint: `post /encode/numeric/property/uint8`
 
-Test operation with request and response model contains property of uint32 type with string encode.
+Test operation with request and response model contains property of uint8 type with string encode.
 Expected request body:
 
 ```json
 {
-  "value": "1"
+  "value": "255"
```
iscai-msft, line 2258: why do we want to change it to `uint8`? Would it be better to add a test for `uint8` instead

archerzz: Thanks. I've added back `uint32` test case.

packages/cadl-ranch-specs/http/encode/numeric/main.tsp
```diff
@@ -62,4 +62,29 @@ namespace Property {
...
+    ```
+    Expected response body:
+    ```json
+    {
+      "value": "255"
+    }
+    ```
+    """)
+  @post
+  op uint8AsStringOptional(@body body: Uint8AsStringProperty): Uint8AsStringProperty;
```
iscai-msft, line 84: do you want to add one for `uint8AsString`? Not sure if it's needed, it's just that corresponds more directly with what the tests are for `uint32`. They can also be templatized this way, but you can do that in a separate pr

archerzz: the reason I add `uint8` is because .net generator doesn't support uint32 yet, so I need an `uint8` case

weidongxu-microsoft: I think Issabella is only about the "optional" part.

I've added uint32 as optional property, because Java type is different on required/optional for primitive type. You may not need this "optional" for uint8

archerzz: OK. I've templatize the test cases.

packages/cadl-ranch-specs/http/encode/numeric/main.tsp
```diff
@@ -13,53 +13,60 @@ namespace Encode.Numeric;
...
+
+  alias SendUint8AsString = SendIntAsString<uint8, "255", Uint8AsStringProperty>;
+
+  @route("/uint8")
+  op uint8AsStringOptional is SendUint8AsString.sendIntAsString;
+
+  model Uint8AsStringProperty {
+    @encode(string)
+    value: uint8;
+  }
```
weidongxu-microsoft, line 44: This value seems not "optional"

archerzz: ah, removed the `Optional` suffix.

packages/cadl-ranch-specs/http/encode/numeric/main.tsp
```diff
@@ -13,53 +13,60 @@ namespace Encode.Numeric;
...
-    Expected response body:
-    ```json
-    {
-      "value": "1"
-    }
-    ```
-    """)
-  @post
-  op uint32AsStringOptional(@body body: Uint32AsStringProperty): Uint32AsStringProperty;
+  op uint32AsStringOptional is SendUint32AsString.sendIntAsString;
```
iscai-msft, line 29: nit: we shouldn't need the alias, you should just be able to directly do `op uint32AsStringOptions is SendIntAsString<uint32, "1", Uint32AsStringProperty>`

iscai-msft (approved):

weidongxu-microsoft (approved):

packages/cadl-ranch-specs/http/encode/numeric/main.tsp
archerzz: Looks like the compiler won't allow it.
image

I'll merge this first. We can create a separate PR to refactor if needed.

PR #712, opened 2024-08-23
Optional-Query-Parameter

mcgallan:
This PR is intended to handle [the test case](https://github.com/Azure/autorest.csharp/blob/3455ff05e5e3c6a95c33d298c479206440c1f94c/test/TestProjects/Parameters-TypeSpec/Parameters-TypeSpec.tsp#L36) that need to be migrated from autorest.csharp to cadl ranch.

### DIFF
--- a/packages/cadl-ranch-specs/http/resiliency/srv-driven/main.tsp
+++ b/packages/cadl-ranch-specs/http/resiliency/srv-driven/main.tsp
@@ -128,6 +128,37 @@ interface AddOptionalParam {
     - A client generated from the second service spec can call the second deployment of a service with api version v1
     - A client generated from the second service spec can call the second deployment of a service with api version v2 with the updated changes
     
+    Tests that we can grow up an operation from accepting one optional parameter to accepting an optional parameter and a required parameter.
+    """)
+  @route("/from-reverse-required")
+  @doc("Operation that grew up from accepting one optional parameter to accepting an optional parameter and a required parameter.")
+  @get
+  fromReverseRequired(
+    @doc("I am an optional parameter")
+    @query
+    parameter?: string,
+
+    @added(Versions.v2)
+    @doc("I'm a new input required parameter")
+    @query
+    `new-parameter`: string,
+  ): NoContentResponse;
+
+  @scenario
+  @scenarioDoc("""
+    Need the following two calls:
+    - Pass in `serviceDeploymentVersion="v2"` and `apiVersion="v1"` with query parameter `parameter="optional"`.
+    - Pass in `serviceDeploymentVersion="v2"` and `apiVersion="v2"` with query parameter `parameter="optional"` and query parameter `new-parameter="new"`.
+    
+    There are three concepts that should be clarified:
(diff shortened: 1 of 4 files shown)


packages/cadl-ranch-specs/http/resiliency/srv-driven/main.tsp
```diff
@@ -113,6 +113,37 @@ interface AddOptionalParam {
...
+    3. Api version: The initial deployment of the service only supports api version 'v1'. The new deployment of the service supports api versions 'v1' and 'v2'.
+    
+    With the above two calls, we test the following configurations from this service spec:
+    - A client generated from the second service spec can call the second deployment of a service with api version v1
+    - A client generated from the second service spec can call the second deployment of a service with api version v2 with the updated changes
+    
+    Tests that we can grow up an operation from accepting one optional parameter to accepting an optional parameter and a required parameter.
+    """)
+  @route("/from-reverse-required")
+  @doc("Operation that grew up from accepting one optional parameter to accepting an optional parameter and a required parameter.")
```
iscai-msft, line 134: isn't this clearly breaking? I'm not sure what we're supposed to do in method signatures with the introduction of a required query parameter

mcgallan: This newly added operation is mainly to test the following scenario: an optional parameter is placed before a required parameter, but in the generated SDK code, the required parameter is placed before the optional parameter. This serves as a special case for testing the fromOneRequired operation. This part of the test case is reflected in AutoRest: https://github.com/Azure/autorest.csharp/blob/3455ff05e5e3c6a95c33d298c479206440c1f94c/test/TestProjects/Parameters-TypeSpec/Parameters-TypeSpec.tsp#L36

iscai-msft: @tadelesh @m-nash do you have thoughts? I know Python won't be able to handle it, but that doesn't mean it shouldn't be in the testserver. I'm just not convinced this is something that sdks should be tolerant of

weidongxu-microsoft: @mcgallan 

This is clearly not a "fromOneRequired" case. There is no "required" parameter in v1, isn't it?

And I also think adding a new *required* parameter in v2 is a breaking change.

mcgallan: > @mcgallan
> 
> This is clearly not a "fromOneRequired" case, which adds an _optional_ parameter in v2. There is no "required" parameter in v1, isn't it?
> 
> And I also think adding a new _required_ parameter in v2 is a breaking change.
> 
> Also, your test case in csharp repo does not have any versioning on parameter at all.

I think placing this part of the test in the service-driven file doesn't seem appropriate. A more suitable approach would be to add an optional-query file in the parameters folder to test the order of optional query parameters and their combination with required query parameters in the generated code. As for the test for adding required query parameters in subsequent versions, if it doesn't make sense, I will close this PR.

PR #703, opened 2024-08-16
Test ARM, test case for location-based resource

v-hongli1:
- https://github.com/Azure/cadl-ranch/issues/686

# Cadl Ranch Contribution Checklist:

- [ ] I have written a [scenario spec](../docs/writing-scenario-spec.md)
- [ ] I have **meaningful** `@scenario` names. Someone can look at the list of scenarios and understand what I'm covering.
- [ ] I have written a [mock API](../docs/writing-mock-apis.md)
- [ ] I have used `@scenarioDoc`s for extra scenario description and to tell people **how to pass** my mock api check.

PR #714, opened 2024-08-29
fix ARM scenario names

XiaofeiCao:
Related comment: https://github.com/Azure/cadl-ranch/pull/701#discussion_r1735033126

Shouldn't affect existing generated SDK code. Verified for Java, codegen PR: https://github.com/microsoft/typespec/pull/4717
May affect existing coverage data in [dashboard](https://azure.github.io/cadl-ranch/), though should be auto-updated after each SDK's generator CI runs.

```
Breaking changes:

1. Renamed namespace `Azure.ResourceManager.Models.Resources` to `Azure.ResourceManager.Resources`.
2. Renamed namespace `Azure.ResourceManager.Models.CommonTypes.ManagedIdentity` to `Azure.ResourceManager.CommonProperties`.
3. Renamed folder `common-types/managed-identity` to `common-properties`.
4. Renamed `ManagedIdentityTrackedResources` interface in `common-properties` to `ManagedIdentity`.
```


packages/cadl-ranch-specs/http/azure/resource-manager/models/common-types/managed-identity/main.tsp
```diff
@@ -47,7 +48,8 @@ model ManagedIdentityTrackedResourceProperties {
 }
 
 @armResourceOperations
-interface ManagedIdentityTrackedResources {
+@clientName("ManagedIdentityTrackedResources")
+interface CrudOperations {
```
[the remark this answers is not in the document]
XiaofeiCao: Yes, it could be misleading.

How about `Azure_ResourceManager_Models_CommonTypes_ManagedIdentity_AsResourceProperty_createWithSystemAssigned`?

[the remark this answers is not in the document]
weidongxu-microsoft: Maybe just `ResourceProperty`? Unsure.

Any opinion from reviewers?

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -791,7 +791,7 @@ Expected response body:
 }
 ```
 
-### Azure_ResourceManager_Models_CommonTypes_ManagedIdentity_ManagedIdentityTrackedResources_createWithSystemAssigned
+### Azure_ResourceManager_Models_CommonTypes_ManagedIdentity_CrudOperations_createWithSystemAssigned
```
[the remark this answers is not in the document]
XiaofeiCao: Original intent for the extra interface `CrudOperations` was to mitigate breaking changes to client code, as we previously have `ManagedIdentityTrackedResources`, which is definitely inappropriate...

Though a further look, as weidong mentioned, `create` directly after it may lead to confusion, even if we remove the interface, e.g. 

`Azure_ResourceManager_Models_CommonTypes_ManagedIdentity_createWithSystemAssigned`

Looks like it's for CRUD operations on ManagedIdentity. Though actually, it's rather for creating another resource with ManagedIdentity as its property.

Thus, we proposed:
`Azure_ResourceManager_Models_CommonTypes_ManagedIdentity_AsResourceProperty_createWithSystemAssigned`
or 
`Azure_ResourceManager_Models_CommonTypes_ManagedIdentity_ResourceProperty_createWithSystemAssigned`.

Or we could remove the managed identity test case, and embed it into one of the resources test... Downside is that we may mix too many test cases inresources...

XiaofeiCao:
@iscai-msft I've renamed the namespace according to https://github.com/Azure/cadl-ranch/pull/703#discussion_r1744043677

I removed the extra `s`, e.g. `Resources`->`Resource`, guess it's better to be shown in dashboard.

Let me know your thoughts.

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -791,12 +791,12 @@ Expected response body:
 }
 ```
 
-### Azure_ResourceManager_Models_CommonTypes_ManagedIdentity_ManagedIdentityTrackedResources_createWithSystemAssigned
+### Azure_ResourceManager_CommonType_ManagedIdentity_Property_createWithSystemAssigned
```
iscai-msft, line 794: I think `CommonTypes` is better grammatically

iscai-msft: I also don't think we need a `_Property_` part of this name. I like just `Azure_Resource_Mangager_CommonTypes_ManagedIdentity_createWithSystemAssigned`

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -946,28 +946,28 @@ Expected response body:
 }
 ```
 
-### Azure_ResourceManager_Models_Resources_NestedProxyResources_delete
+### Azure_ResourceManager_Resource_Nested_delete
```
iscai-msft, line 949: is the fact that it's a proxy resource part of the test? If so, we can definitely include `NestedProxyResource`. I do think resources should still be plural for grammatic sense. So either `Azure_ResourceManager_Resources_NestedProxyResource_delete` (note the first Resources is plural, but the second one is singular. It's because `Resources` is a grouping of all of the resources, while proxy resource is a singular resource in the list of `Resources`). Or it could be `Azure_ResourceManager_Resources_Nested_delete` if proxy doesn't matter to the test

iscai-msft (commented):
thank you for all of your work @XiaofeiCao, I think this definitely is almost there. @timotheeguerin if you could take a look with your ARM knowledge, that would be great. Thank you so much!

packages/cadl-ranch-specs/cadl-ranch-summary.md
weidongxu-microsoft: `@armResourceOperations` have some autoRoute (and tag) related code within. And it is used by almost all ARM typespec spec.
https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-resource-manager/src/resource.ts#L255-L256

Therefore, I'd wish we keep the operation group and the arm decorator, unless we are sure we don't want it in test.

XiaofeiCao: Thanks! I also prefer to have the interface. Adding `@autoRoute` is uncommon in current ARM practice.
Another option is that we could have the scenario name be `Azure_Resource_Mangager_CommonTypes_ManagedIdentity_ResourceProperty`, and have all the `create`, `get` and `update` grouped into the `ResourceProperty` interface.

Wonder if there is a way to define the full path of the scenario. 
e.g. `@scenario(fullName="Azure_Resource_Mangager_CommonTypes_ManagedIdentity_createWithSystemAssigned")`
That way, we could decouple scenario structure from the actual client structure.

iscai-msft:
@XiaofeiCao is this PR ready to re-review?

XiaofeiCao:
> @XiaofeiCao is this PR ready to re-review?

Yes, please! Would like a re-review for this PR, since it's required for future ARM test cases.

I added breaking change note in changeset, let me know if it looks good.

packages/cadl-ranch-specs/cadl-ranch-summary.md
```diff
@@ -833,19 +833,19 @@ Expected response body:
 }
 ```
 
-### Azure_ResourceManager_Models_CommonTypes_ManagedIdentity_ManagedIdentityTrackedResources_get
+### Azure_ResourceManager_CommonTypes_ManagedIdentity_Property_get
```
iscai-msft, line 836: would we just be able to do `Azure_ResourceManager_CommonTypes_ManagedIdentity_get` and remove `Property` for all of the `ManagedIdentity` sections?

iscai-msft (approved):
Thanks @XiaofeiCao this looks great! Can you make sure @timotheeguerin approves first, and that folks in shanghai are aware of the breaking change? After that is done, should be good to merge. Thanks again for all of your changes and all the work you've put into this, it looks great

packages/cadl-ranch-specs/cadl-ranch-summary.md
XiaofeiCao: Thanks, seems the answer is yes. I found a way to make the namespace `Azure.ResourceManager.CommonTypes` work. Seems I just need to change the `Versions` name.

Downside is the confusion of whether `ManagedIdentity` is resource or a property. 
From dashboard perspective, since we already have the separation of `Resources` and `CommonTypes`, it should be good.
From codegen side, we'll have generated code be like `ManagedIdentity.create()`... Though seems only to be implementation details.

XiaofeiCao:
> Thanks @XiaofeiCao this looks great! Can you make sure @timotheeguerin approves first, and that folks in shanghai are aware of the breaking change? After that is done, should be good to merge. Thanks again for all of your changes and all the work you've put into this, it looks great

Yes, shanghai folks are informed of this breaking change.

timotheeguerin (approved):

weidongxu-microsoft (approved):
Recommend that you tested this in Java (as it seems involve some mockapi change as well).

XiaofeiCao:
> Recommend that you tested this in Java (as it seems involve some mockapi change as well).

Yeah, regened and tested: https://github.com/microsoft/typespec/pull/4717

PR #718, opened 2024-09-02
Dashboard arm

XiaofeiCao:
1. Added Azure MPG report.
2. Removed MPG from Azure DPG and standard report.

New Bitmap image (3)

### DIFF
--- a/.changeset/warm-wasps-boil.md
+++ b/.changeset/warm-wasps-boil.md
@@ -1,0 +1,7 @@
+---
+"@azure-tools/cadl-ranch-coverage-sdk": minor
+"@azure-tools/cadl-ranch-dashboard": minor
+"@azure-tools/cadl-ranch": minor
+---
+
+Add dashboard for Azure Resource Manager.
--- a/packages/cadl-ranch-coverage-sdk/src/constants.ts
+++ b/packages/cadl-ranch-coverage-sdk/src/constants.ts
@@ -1,1 +1,14 @@
-export const GeneratorMode: string[] = ["azure", "standard"];
+export enum GeneratorMode {
+  azure = "azure",
+  standard = "standard",
+}
+export const TableConfigs: TableConfig[] = [
+  { mode: GeneratorMode.azure, scenarioFilter: (name) => !name.startsWith("Azure_ResourceManager_") },
+  { mode: GeneratorMode.azure, scenarioFilter: (name) => name.startsWith("Azure_ResourceManager_") },
+  { mode: GeneratorMode.standard, scenarioFilter: (name) => !name.startsWith("Azure_ResourceManager_") },
+];
+
+export interface TableConfig {
+  mode: GeneratorMode;
+  scenarioFilter?: (scenarioName: string) => boolean;
+}
--- a/packages/cadl-ranch-coverage-sdk/src/constants.ts
+++ b/packages/cadl-ranch-coverage-sdk/src/constants.ts
@@ -2,21 +2,3 @@ export enum GeneratorMode {
   azure = "azure",
   standard = "standard",
 }
-export const TableConfigs: TableConfig[] = [
-  {
-    mode: GeneratorMode.azure,
-    scenarioFilter: (name) => !name.startsWith("Azure_ResourceManager_"),
-    name: "Azure DPG",
-  },
-  { mode: GeneratorMode.azure, scenarioFilter: (name) => name.startsWith("Azure_ResourceManager_"), name: "Azure MPG" },
-  {
-    mode: GeneratorMode.standard,
-    scenarioFilter: (name) => !name.startsWith("Azure_ResourceManager_"),
-  },
-];
-
-export interface TableConfig {
-  mode: GeneratorMode;
-  scenarioFilter?: (scenarioName: string) => boolean;
-  name?: string;
-}
--- a/packages/cadl-ranch-coverage-sdk/src/constants.ts
+++ b/packages/cadl-ranch-coverage-sdk/src/constants.ts
@@ -3,12 +3,20 @@ export enum GeneratorMode {
   standard = "standard",
 }
 export const TableConfigs: TableConfig[] = [
-  { mode: GeneratorMode.azure, scenarioFilter: (name) => !name.startsWith("Azure_ResourceManager_") },
-  { mode: GeneratorMode.azure, scenarioFilter: (name) => name.startsWith("Azure_ResourceManager_") },
-  { mode: GeneratorMode.standard, scenarioFilter: (name) => !name.startsWith("Azure_ResourceManager_") },
+  {
+    mode: GeneratorMode.azure,
+    scenarioFilter: (name) => !name.startsWith("Azure_ResourceManager_"),
+    name: "Azure DPG",
+  },
+  { mode: GeneratorMode.azure, scenarioFilter: (name) => name.startsWith("Azure_ResourceManager_"), name: "Azure MPG" },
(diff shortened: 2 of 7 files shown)


packages/cadl-ranch-dashboard/src/components/dashboard-table.tsx
```diff
@@ -286,5 +292,19 @@ function createTree(manifest: ScenarioManifest): ManifestTreeNode {
 
     current.scenario = scenario;
   }
-  return root;
+
+  return cutTillMultipleChildren(root);
```
XiaofeiCao, line 296: Remove unnecessary parent nodes which only contain 1 child.

before:
- Azure:
  - ResourceManger
    - Models
      - CommonTypes
      - Resources

Now: 
- CommonTypes
- Resources

packages/cadl-ranch-coverage-sdk/src/constants.ts
```diff
@@ -1 +1,22 @@
-export const GeneratorMode: string[] = ["azure", "standard"];
+export enum GeneratorMode {
+  azure = "azure",
+  standard = "standard",
+}
+export const TableConfigs: TableConfig[] = [
```
timotheeguerin, line 5: this feels like this should not be part of the SDK no?

XiaofeiCao: Right, it should be part of the dashboard. Fixed.

packages/cadl-ranch-dashboard/src/components/dashboard-table.tsx
XiaofeiCao: Or we could expand root nodes with 1 child by default, to better reflect the actual structure.

timotheeguerin: I think ideally we would only collapse into a single node(like vscode does) but I understand the need here. Maybe what you want instead is to say you want to only show a subgraph(Azure.AzureResourceManager) instead of filtering

.changeset/warm-wasps-boil.md
timotheeguerin, line 1: Making a comment here so we can have a thread.

So I had plans to change this UI to support better the new emitter names by going the other way(removing the separate tables and instead having some dynamic filtering)

Not necessarily opposed to doing this change(at least for now) but would like to understand the emitter structure you see here.

- Standard scenarios
- Azure Core Scenarios
- Arm Scenarios
  The way I saw it was as follow:

- We have 3 category of emitters

  - `@typespec/http-client-$lang`
  - `@azure-tools/typespec-$lang` with mode: dataplane
  - `@azure-tools/typespec-$lang` with mode: arm

- And we have categories of scenarios
  - Standard scenarios
  - Azure Core Scenarios
  - Arm Scenarios

| Scenarios              | `@typespec/http-client-$lang` | `@azure-tools/typespec-$lang` with mode: dataplane | `@azure-tools/typespec-$lang` with mode: arm |
| ---------------------- | ----------------------------- | -------------------------------------------------- | -------------------------------------------- |
| Standard scenarios     | ✅                            | ✅                                                 | ✅                                           |
| Azure common scenarios | ❌                            | ✅                                                 | ✅                                           |
| Azure Core Scenarios   | ❌                            | ✅                                                 | ❌                                           |
| Arm Scenarios          | ❌                            | ❌                                                 | ✅                                           |


is this more or less accurate?

packages/cadl-ranch-dashboard/src/components/dashboard-table.tsx
XiaofeiCao: Not exactly, but I think I get your point. 
Do you mean instead of having 3 separate graphs, we could have one graph be the subgraph of a total group of scenarios, projected by each mode? I do believe it's better.

.changeset/warm-wasps-boil.md
XiaofeiCao: Thanks for the chart, mostly looks accurate to me. 
One thing to confirm is the definition of `Standard scenarios` and `Azure common scenarios`. Does `Standard` mean non-branded? I believe `@azure-tools/typespec-$lang` with mode `arm` doesn't have non-branded scenarios since it's for Azure only.

timotheeguerin: yeah I mean I don't know how you want to do it with the data plane vs arm emitters.

Do you expect arm then to only select some of the standard test, none of them and define its own parallel structure? I would kinda expect at least a lot of the generic type test should be applicable for arm as well no?

srnagar: @timotheeguerin the branded libraries will be named `@azure-typespec/http-client-[lang]`. Regarding arm, we are currently thinking we will use the same emitter with arm mode but we might even consider using a separate package `@azure-typespec/http-client-[lang]-mgmt`. This is not finalized yet and we will update this thread if we decide to go with a separate emitter for arm.

XiaofeiCao: @timotheeguerin Could you give an example of the generic type test(is it part of `Azure common scenarios`?) that should be covered for both dpg and arm?

Personally I feel like `Arm Scenarios` for "`@azure-tools/typespec-$lang` with mode: arm" is good enough.

timotheeguerin: I assume things like using armResourceIdentifier/uuid or other types provided in azure core but would also be used in arm spec?

XiaofeiCao: Not sure we should include `armResourceIdentifier` for dpg test case... As the name implies, it's for arm.

For uuid case, we already have that in some ARM templates, e.g. [`ManagedServiceIdentity`](https://github.com/Azure/typespec-azure/blob/c76a9c190ec5753627e56365dce8dd0ca53486f3/packages/typespec-azure-resource-manager/lib/common-types/managed-identity.tsp#L16)

Could you share more about how you classify these four categories? I'm also wondering why `Azure Core Scenarios` is excluded for ARM in the chart..

timotheeguerin: Well `armResourceIdentifier` was explicitly added(moved) to `Azure.Core` because there was data plane use cases.

For the question on how I classify categories that's I guess where I am wondering how you all see testing the arm scenario work. I would kinda expect that in an arm library you can still send for example an int32, string, boolean, query parameters in certain ways(Basically a subset of the standard scenarios). And in the same style some scenarios supported by azure core as well.
So the question is do you want the arm scenario testing to retest those scenarios again or it should just have a parallel test suite and in that case then yeah having an Azure common is probably useless.

XiaofeiCao: Got it. 
I think original plan is to have a parallel test suite for ARM, not retest shared scenarios between DPG and ARM. My arguments are:
1. Most language's arm emitter shares the same logic with dpg for most of types, e.g. int32, string, etc. The extra effort to retest is huge, though I'm not sure about the value.
2. In Swagger, we follow this parallel test(at least for Java) and haven't met issues so far.
3. It's hard to figure out which one to fit in the common scenario. Some of them may only serve dpg(e.g. client hierarchy, json-merge-patch), some only ARM. 

Just my pov. Welcome feedback.

XiaofeiCao:
@timotheeguerin 
Would you help merge this PR if no objection? 
There's an ongoing effort to [refactor existing ARM tests](https://github.com/Azure/cadl-ranch/pull/714) as well as adding new ones. Kinda don't want it to affect overall cadl-ranch dashboard coverage, thanks!

PR #726, opened 2024-09-12
cadl-ranch-specs, case for access/usage on namespace

weidongxu-microsoft:

### DIFF
--- a/packages/cadl-ranch-specs/http/azure/client-generator-core/access-namespace/main.tsp
+++ b/packages/cadl-ranch-specs/http/azure/client-generator-core/access-namespace/main.tsp
@@ -1,0 +1,36 @@
+import "@typespec/http";
+import "@azure-tools/cadl-ranch-expect";
+import "@azure-tools/typespec-client-generator-core";
+
+using TypeSpec.Http;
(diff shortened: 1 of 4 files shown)


weidongxu-microsoft:
Current main.tsp has a client that only have internal API.
I am not sure if all language can call it in test.
Let me know if better idea.

packages/cadl-ranch-specs/http/azure/client-generator-core/access-namespace/main.tsp
```diff
@@ -0,0 +1,50 @@
...
+  
+  Expected response body:
+  ```json
+  {
+    "name": <any string>
+  }
+  ```
+  """)
+@access(Access.internal)
+namespace InternalOperation {
```
iscai-msft, line 33: should this test be added to `client-generator-core/access` as a new separate namespace?

weidongxu-microsoft: Do you mean put it into `client-generator-core/access/namespace/main.tsp` ? Or reverse it as `client-generator-core/namespace/access/main.tsp` ?

PR #727, opened 2024-09-12
cadl-ranch-specs, add basic test case of clientInitialization

weidongxu-microsoft:


chunyu3 (commented):
We may need add more complex cases:

- more complex client hierarchy, e.g. more than 3 layer clients
- co-operate with @client / @operationGroup decorator
- with paramMaps to map a parameter to the property of the Client initialization Model
- a senario that not all operations has the bumped parameter

weidongxu-microsoft:
> We may need add more complex cases:
> 
> * more complex client hierarchy, e.g. more than 3 layer clients
> * co-operate with @client / @operationGroup decorator
> * with paramMaps to map a parameter to the property of the Client initialization Model
> * a senario that not all operations has the bumped parameter

Wondering whether we need to have a subfolder for different scenarios. And this be a "basic" folder.

As the characteristic of the "client hierarchy", some of them cannot co-exist in same main.tsp. E.g. client by namespace/interface cannot be in same file as client by `@client`/`@operationGroup`.

tadelesh:
@weidongxu-microsoft please hold this. i'm going to redesign all the client related thing in tcgc to consolidate all things related to clients, it may cause changes for decorator.

weidongxu-microsoft:
> @weidongxu-microsoft please hold this. i'm going to redesign all the client related thing in tcgc to consolidate all things related to clients, it may cause changes for decorator.

Sure.

Note that Face API already uses this decorator in their specs (and Java emitter also uses the `.initialization` in TCGC).

packages/cadl-ranch-specs/http/client/initialization/basic/main.tsp
```diff
@@ -0,0 +1,31 @@
...
+
+model SubClientOptions {
+  name: string;
+}
+
+@clientInitialization(SubClientOptions)
+interface SubClient {
+  @post
+  @route("/sub-client/{name}:action")
+  action(@path name: string): void;
```
XiaofeiCao, line 30: I'm not sure, maybe add another parameter that will be taken in method signature, as a contrast?

weidongxu-microsoft: Sure, I will first update to our local test first, as this one likely going to take time.
https://github.com/microsoft/typespec/pull/5027

PR #728, opened 2024-09-14
SingletonResource Fix

mcgallan:
Remove the unnecessary ListByResourceGroup operation for SingletonResource and correct the name of the get operation.


weidongxu-microsoft (commented):
Why is list unnecessary?

real service https://github.com/Azure/azure-rest-api-specs/blob/main/specification/networkanalytics/NetworkAnalytics.Management/main.tsp#L523-L524

mcgallan:
> Why is list unnecessary?
> 
> real service https://github.com/Azure/azure-rest-api-specs/blob/main/specification/networkanalytics/NetworkAnalytics.Management/main.tsp#L523-L524

Regarding this operation, I have inquired in the dotnet SDK channel, according to Dapeng's opinion, based on the current design, our dotnet SDK will not generate a list method for singleton resources, this part can omit the list operation. Moreover, the more critical issue is that in autorest.csharp, we actually do not generate the related code for this list. So I believe the list operation is not necessary at the moment.

tadelesh:
i think list should be added for testing. all other languages will generate list operation for singleton resource.

PR #729, opened 2024-09-22
[@azure-tools/cadl-ranch-specs] - Add Scenarios for Server Tests and Testing Framework

sarangan12:
This PR adds the server-test scenarios along with the framework to test those scenarios.

In order to test the new server-test scenarios, there are 3 modes enabled:

1. Specify the name of one scenario in the command line.
2. Specify the list of scenarios to be tested in a file and provide this file as input in the command line
3. If the above modes are not specified, then the framework will execute all the scenarios. 

**Mode 1 - One Scenario in Command Line**

Screenshot 2024-09-23 095806

**Mode 2 - Multiple Scenarios in file** 

Screenshot 2024-09-23 100006

**Mode 3 - All Scenarios**
Screenshot 2024-09-23 100310

Screenshot 2024-09-23 100614

This PR is equivalent of the PR https://github.com/microsoft/typespec/pull/4516. Except for some cosmetic changes in imports, etc, the code changes are same. I am doing the changes in both to ensure both are in sync.


sarangan12:
@timotheeguerin The original plan is to make the changes here and then port this package completely to the typespec repository. Do you agree? Do you want the changes only in Typespec repo i.e. migrate the cadl-ranch-specs repository with the new code changes in this PR? Please let me know.

timotheeguerin:
sure as long as the change you make here are completely back compatible

sarangan12:
> sure as long as the change you make here are completely back compatible

They are backward compatible. None of the existing scenarios will be affected

packages/cadl-ranch/src/actions/server-test.ts
```diff
@@ -0,0 +1,386 @@
...
+      counter++;
+    }
+    return testCasesContent;
+  }
+
+  private getFileContents(): string {
+    return `${this.getImportContents()}\n${this.getGlobalDeclarationContents()}\n${this.getDescribeContents()}`;
+  }
+
+  public async generateFile() {
```
timotheeguerin, line 279: This feels very weird to resort to that, why not directly run things, this is going to make debugging impossible. 
I don't feel like there is much point running vitest as well on top or any other test runner. We won't use the language for the test runner

timotheeguerin: this just feels like a lot of pain to maintain regardless, now if we need to add some more complex scenario you don't need to just make the call yourself you need to reserialize that code to a file(the whole request creation that is currently hardcoded in that file).
What is the benefit of using a test runner here? I think it would be much better to just have the cli and if having vitest IDE support is important then you call to vitest api directly without shelkling a new process and writing new files.

sarangan12:
This PR is no longer required. Closing it

Issue #733, opened 2024-09-30
bug in templated case


weidongxu-microsoft:
There is no `kind` property in `NumericType<int32>`

https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/type/model/templated/main.tsp#L44-L71

```
com.azure.core.exception.HttpResponseException: Status code 400, "{"message":"Body provided doesn't match expected body","expected":{"kind":"Int32Values","values":[1234],"value":1234},"actual":{"values":[1234],"value":1234}}"
```

melina5656:
After discussion, we believe that this test case is not suitable to be added. we have removed it. So close this issue.

PR #735, opened 2024-09-30
Fix templated issue

melina5656:
Fixed some issues that exist in `http\type\model\templated`.


melina5656:
Fix issue: #733

weidongxu-microsoft (approved):

Issue #734, opened 2024-09-30
bug in versioning/removed/mockapi.ts case


weidongxu-microsoft:
Request should be correct to `versioning/removed/api-version:v2preview/v3` with `enumMemberV2Preview`. But rejected by mockapi

```
[ForkJoinPool-1-worker-1] INFO com.versioning.removed.implementation.RemovedClientImpl$RemovedClientService.modelV3Sync - {"az.sdk.message":"HTTP request","method":"POST","url":"http://localhost:3000/versioning/removed/api-version:v2preview/v3","tryCount":1,"Date":"Mon, 30 Sep 2024 02:10:20 GMT","Content-Type":"application/json","x-ms-client-request-id":"5d87308e-055b-45fa-83a4-e375c4f8d42d","Accept":"application/json","User-Agent":"azsdk-java-${project.artifactId}/${project.version} (21.0.3; Windows 11; 10.0)","content-length":45,"body":"{\"id\":\"123\",\"enumProp\":\"enumMemberV2Preview\"}"}
[ForkJoinPool-1-worker-1] INFO com.versioning.removed.implementation.RemovedClientImpl$RemovedClientService.modelV3Sync - {"az.sdk.message":"HTTP response","statusCode":400,"url":"http://localhost:3000/versioning/removed/api-version:v2preview/v3","durationMs":6,"content-length":160,"Date":"Mon, 30 Sep 2024 02:10:20 GMT","Connection":"keep-alive","ETag":"W/\"a0-7pCSm5UxU05anpeENBbPPWfTNCY\"","Content-Type":"application/json; charset=utf-8","redactedHeaders":"Keep-Alive,X-Powered-By","content-length":160,"body":"{\"message\":\"Body provided doesn't match expected body\",\"expected\":{\"id\":\"123\",\"enumProp\":\"enumMemberV1\"},\"actual\":{\"id\":\"123\",\"enumProp\":\"enumMemberV2Preview\"}}"}

com.azure.core.exception.HttpResponseException: Status code 400, "{"message":"Body provided doesn't match expected body","expected":{"id":"123","enumProp":"enumMemberV1"},"actual":{"id":"123","enumProp":"enumMemberV2Preview"}}"
```

weidongxu-microsoft:
Commented by Crystal, the expected request body should only include `id=123`, no `enumProp` as it is removed in v2preview.

PR #736, opened 2024-09-30
Fix Issue In current Versioning\Removed

mcgallan:
Fixed issue that exist in http\versioning\removed.
Related Issue Link: https://github.com/Azure/cadl-ranch/issues/734


packages/cadl-ranch-specs/http/versioning/removed/main.tsp
```diff
@@ -188,3 +188,25 @@ interface InterfaceV1 {
...
+  ```json
+  { "id": "123"}
+  ```
+  
+  """)
+@post
+@route("/preview")
+@added(Versions.v2preview)
+@removed(Versions.v2)
+op preview(@body body: ModelV3): ModelV3;
```
weidongxu-microsoft, line 212: I have no idea why it needs to be broken to 2 different operations.

weidongxu-microsoft: Please figure out WHY the test facing same error.

weidongxu-microsoft: I will leave you and .NET dev to figure out what is the expected test first...

mcgallan: > Please figure out WHY the test facing same error.
> 
> Making a separate operation just make this test case pointless. You have bug, go fix the bug. Not writing a different test.



> Please figure out WHY the test facing same error.
> 
> Making a separate operation just make this test case pointless. You have bug, go fix the bug. Writing a different test is not the solution.

The previous mockapi was intended to validate the request body and return the expected response body correctly based on different paths. However, in the generated code, the mockapi does not correctly implement this functionality. For different paths of the same operation, the mockapi only validates the request body based on the first path's mockapi and does not implement different validations for different paths. Regardless of whether there are differences in the paths, it always uses the Expect Body of the first path to validate the information passed to this operation. This is the superficial reason for the occurrence of this bug.

mcgallan:
@tadelesh Regarding the mockapi added in Version\Removed tests, it does not actually achieve the expected effect. We want the mockapi to perform different validations and return different response bodies for the same operation based on different paths. However, in the code generated by autorest.csharp, the mockapi actually only validates based on the Expect Body of the first path and only returns the response body corresponding to the first path, without handling different paths differently. This has caused a bug. Now I want to know if there is a better solution for this situation. If the mockapi cannot perform corresponding validations and provide return values based on different paths, is there another way to achieve our expected effect?

tadelesh:
> @tadelesh Regarding the mockapi added in Version\Removed tests, it does not actually achieve the expected effect. We want the mockapi to perform different validations and return different response bodies for the same operation based on different paths. However, in the code generated by autorest.csharp, the mockapi actually only validates based on the Expect Body of the first path and only returns the response body corresponding to the first path, without handling different paths differently. This has caused a bug. Now I want to know if there is a better solution for this situation. If the mockapi cannot perform corresponding validations and provide return values based on different paths, is there another way to achieve our expected effect?

there are other test cases which also has multiple paths: https://github.com/Azure/cadl-ranch/blob/fbd34bdf33ecba8c65275e33533084e2b00f8eec/packages/cadl-ranch-specs/http/resiliency/srv-driven/mockapi.ts#L64. could you double check if it is a corner case or wrong usage?

packages/cadl-ranch-specs/http/versioning/removed/main.tsp
mcgallan: @weidongxu-microsoft This bug has now been fixed. After investigation, it was found that the issue was caused by using `:` to denote dynamic path parameters in the path. The correct approach is to use `[:]` to connect api-version and the dynamic version number. Using just `: `led to the problem exhibited by this bug: the mockapi could not correctly recognize multiple paths for the same operation and take different actions based on the different paths.

mcgallan:
> > @tadelesh Regarding the mockapi added in Version\Removed tests, it does not actually achieve the expected effect. We want the mockapi to perform different validations and return different response bodies for the same operation based on different paths. However, in the code generated by autorest.csharp, the mockapi actually only validates based on the Expect Body of the first path and only returns the response body corresponding to the first path, without handling different paths differently. This has caused a bug. Now I want to know if there is a better solution for this situation. If the mockapi cannot perform corresponding validations and provide return values based on different paths, is there another way to achieve our expected effect?
> 
> there are other test cases which also has multiple paths:
> 
> https://github.com/Azure/cadl-ranch/blob/fbd34bdf33ecba8c65275e33533084e2b00f8eec/packages/cadl-ranch-specs/http/resiliency/srv-driven/mockapi.ts#L64
> 
> . could you double check if it is a corner case or wrong usage?

This issue has been resolved. It was caused by using : instead of [:] to denote dynamic parameters in multiple paths. As for the resiliency part, since C# currently does not seem to support this feature, the successful execution of the test case does not indicate much. However, the way the mockapi is written is correct.

tadelesh (approved):

mcgallan:
@weidongxu-microsoft I have upgraded the PR code according to the new mockapi requirements and have also tested and verified it locally against three versions of the generated code. Please help review.

archerzz (approved):

PR #741, opened 2024-10-04
[cadl-ranch-*] Backport the server testing changes from typespec repository to cadl-ranch repository

sarangan12:
As part of `cadl-ranch` migration, several packages have been migrated from `cadl-ranch` repository to `typespec` repository. Now, the changes made to the `typespec` repository has to be backported to the `cadl-ranch` repository. This PR accomplishes that.  

This PR consists of the following changes:

1. Changes to the `cadl-ranch-api` & `cadl-ranch` packages have been migrated from the corresponding packages in the `typespec` repository. Except for 3 design changes, there is no bigger changes. These 3 changes are indicated in the PR Comments. 
2. Changes to the `cadl-ranch-specs` repository have been migrated from the corresponding package in the `typespec` repository. In addition to the endpoints in the `typespec` repository, the following folders have been migrated to new model:
- client folder
- azure folder
- resiliency folder
- payload/pageable folder

This change is done for 2 reasons. One, We do not want anyone to add new features in these folders using the old model. Second, I am planning to port these folders to the new package `azure-http-specs` for which I will create the PR shortly after this is merged. 

I have already executed e2e scenarios and all validations are successful.

**Note** I have not made any major/minor version bumps for any of the 3 packages involved in this PR

Please review and approve the PR. Thanks


packages/cadl-ranch-specs/http/azure/resource-manager/models/resources/mockapi.ts
```diff
@@ -17,9 +17,9 @@ const validTopLevelResource = {
   systemData: {
     createdBy: "AzureSDK",
     createdByType: "User",
-    createdAt: new Date(),
+    createdAt: "2024-10-04T00:56:07.442Z",
```
sarangan12, line 20: Design Change 1: The `new Date()` is changed to a fixed date. This enables the validation to be successful. Else, the server and client time will always be different and will cause error.

packages/cadl-ranch/src/actions/server-test.ts
```diff
@@ -0,0 +1,172 @@
...
+        if (this.mockApiDefinition.response.body.rawContent !== responseData.replace(this.serverBasePath, "")) {
+          logger.error(`Response data mismatch for ${this.name} endpoint`);
+          logger.error(`Expected: ${this.mockApiDefinition.response.body} - Actual: ${response.data}`);
+          throw new Error(`Response data mismatch for ${this.name} endpoint`);
+        }
+      }
+    }
+    if (this.mockApiDefinition.response.headers) {
+      for (const key in this.mockApiDefinition.response.headers) {
+        if (this.mockApiDefinition.response.headers[key] !== response.headers[key].replace(this.serverBasePath, "")) {
```
sarangan12, line 108: Originally, in the `typespec` package, the line is `if (this.mockApiDefinition.response.headers[key] !== response.headers[key])`. With this PR, I have changed it to `if (this.mockApiDefinition.response.headers[key] !== response.headers[key].replace(this.serverBasePath, ""))`. Because, the paging/LRO endpoints have nextLink in headers which have the baseURL. I am replacing it with spaces so, it can be compared and the validation will be successful.

packages/cadl-ranch/src/actions/helper.ts
```diff
@@ -0,0 +1,104 @@
...
+    }
+    request.options.requestBody = formData;
+    request.options.config = {
+      ...request.options.config,
+      headers: formData.getHeaders(),
+    };
+  }
+}
+
+function checkAndUpdateEndpoint(request: ServiceRequest) {
```
sarangan12, line 38: Design change 3: This method has been added newly in this PR. This is added to accomodate the scenarios where the URLs are registered as `..../users/:id` and the client call has to happen  as `..../users/`'.

XiaofeiCao (approved):
LGTM for azure/resource-manager.

So in the future, we should only update 4 azure scenarios(azure/client/resiliency/pageable) in this repo, and for others, we should be writing in microsoft/typespec?

timotheeguerin:
> LGTM for azure/resource-manager.
> 
> So in the future, we should only update 4 azure scenarios(azure/client/resiliency/pageable) in this repo, and for others, we should be writing in microsoft/typespec?

for a short period of time if you need to update teh typespec scenarios you should update it in both place.

For azure we are also moving them to the typespec-azure repo so you will need to update it there when that is done too.

When that is done we'll try to quickly rollout on having everyone use the new typespec and typespec-azure version instead so we can stop maintaining this repo

timotheeguerin (approved):
Lets merge this

iscai-msft (approved):

PR #755, opened 2024-10-16
[@azure-tools/cadl-ranch-specs] - Handle PassOnCode, Remove Handler & Remove commonapi.ts file

sarangan12:
This PR consists of of the following changes:

1. One of the was mislabled as `passOnSuccess` instead of `passOnCode`. In this PR, it is changed back to `passOnCode`. 
2. I am backporting the changes from the PRs: https://github.com/microsoft/typespec/pull/4725 & https://github.com/Azure/typespec-azure/pull/1691 regarding removing the handler code. This would stop anyone from making changes to handler code in this repository and thus avoid the need of porting it back to `typespec` repository.
3. With the recent fix in PR https://github.com/Azure/cadl-ranch/pull/748, the `test:e2e` script broke. Now, in the `main` branch - `cadl-ranch-specs` repository, if you run the `pnpm test:e2e` script, it will error out. This is happening because in one of the scenario, the way array's are expected and the   way axios is sending it. But, with the `handler` code removed this is automatically fixed.
4. One of the files, `commonapi.ts` under `authetication` folder is no longer required. That file is removed.

Please review and approve the PR. Thanks


timotheeguerin (approved):
Looks good but lets keep this PR open one day as a FYI for shanghai folks just in case it will break

packages/cadl-ranch-specs/http/azure/client-generator-core/access/mockapi.ts
```diff
@@ -58,15 +49,6 @@ Scenarios.Azure_ClientGenerator_Core_Access_RelativeModelInOperation = passOnSuc
       status: 200,
       body: json({ name: "Madge", inner: { name: "Madge" } }),
     },
-    handler: (req: MockRequest) => {
-      if (!("name" in req.query)) {
-        throw new ValidationError("Should submit name query", "any string", undefined);
-      }
```
weidongxu-microsoft, line 64: test failure here is due to difference of `request.params` and the validation.

validation here just check the existence of `name` in query.

packages/cadl-ranch-specs/http/azure/resource-manager/models/common-types/managed-identity/mockapi.ts
```diff
@@ -78,26 +78,6 @@ Scenarios.Azure_ResourceManager_Models_CommonTypes_ManagedIdentity_ManagedIdenti
...
-        "Unexpected managed identity resource name",
-        "identity",
-        req.params.managedIdentityResourceName,
-      );
-    }
-    return {
-      status: 200,
-      body: json(validSystemAssignedManagedIdentityResource),
-    };
-  },
```
weidongxu-microsoft, line 100: We probably still need the `handler`. The request does not need to be exactly same payload of `request`. The check is it MUST contains the correct `identity` property.

packages/cadl-ranch-specs/http/azure/core/basic/mockapi.ts
```diff
@@ -18,16 +18,6 @@ Scenarios.Azure_Core_Basic_createOrUpdate = passOnSuccess({
     body: { name: "Madge" },
   },
   response: { status: 200, body: json(validUser) },
-  handler: (req: MockRequest) => {
-    if (req.params.id !== "1") {
-      throw new ValidationError("Expected path param id=1", "1", req.params.id);
-    }
```
weidongxu-microsoft, line 24: Does `request.params.id = 1` in request works in validation for path parameter?

I got error `{"message":"Expected query param id=1 but got undefined","expected":"1"}` when running the test (on all the cases in this module that checks `id=1` in path parameter.

packages/cadl-ranch-specs/http/azure/core/traits/mockapi.ts
```diff
@@ -33,35 +33,6 @@ Scenarios.Azure_Core_Traits_smokeTest = passOnSuccess({
       "x-ms-client-request-id": "86aede1f-96fa-4e7f-b1e1-bf8a947cb804",
     },
   },
-  handler: (req: MockRequest) => {
-    if (!("x-ms-client-request-id" in req.headers)) {
-      throw new ValidationError("Should submit header x-ms-client-request-id", "any uuid", undefined);
-    }
```
weidongxu-microsoft, line 39: Same, here the validation is only GUID. (the header is automatically generated by SDK, hence "any uuid")

packages/cadl-ranch-specs/http/payload/json-merge-patch/mockapi.ts
```diff
@@ -79,27 +72,6 @@ Scenarios.Payload_JsonMergePatch_updateResource = passOnSuccess({
...
-      body: json({
-        name: "Madge",
-        map: {
-          key: {
-            name: "InnerMadge",
-          },
-        },
-      }),
-    };
-  },
```
weidongxu-microsoft, line 102: We may need to keep the handler on json-merge-patch (or at least modify the `expectedUpdateBody`.

The expected request may more accurately described in scenarioDoc
https://github.com/Azure/cadl-ranch/blob/aef9ccc66392cc6567c3f7621d99e6217f8ee2b0/packages/cadl-ranch-specs/http/payload/json-merge-patch/main.tsp#L107-L120
One difference is the `name: "InnerMadge",` part. Since json-merge-patch did not update this field, it can be omitted in request.

packages/cadl-ranch-specs/http/azure/client-generator-core/access/mockapi.ts
weidongxu-microsoft: This can be updated to emitter test.

XiaofeiCao:
@sarangan12 I merged a PR that contains conflicts with your PR:
https://github.com/Azure/cadl-ranch/pull/714

It reorganized ARM scenarios, mostly renames. Let me know if you run into any issues.

packages/cadl-ranch-specs/http/payload/xml/mockapi.ts
```diff
@@ -134,12 +134,6 @@ function createServerTests(uri: string, data?: any) {
         status: 200,
         body: xml(data),
       },
-      handler: (req: MockRequest) => {
```
tadelesh, line 137: for xml, current request comparison is format sensitive, which cause error:
```
  '<ModelWithEncodedNamesSrc>\n' +
  '  <SimpleModelData>\n' +
  '    <name>foo</name>\n' +
  '    <age>123</age>\n' +
  '  </SimpleModelData>\n' +
  '  <PossibleColors>\n' +
  '    <string>red</string>\n' +
  '    <string>green</string>\n' +
  '    <string>blue</string>\n' +
  '  </PossibleColors>\n' +
  '</ModelWithEncodedNamesSrc>\n'
 Actual:
"<?xml version='1.0' encoding='utf-8'?>\n" +
  '<ModelWithEncodedNamesSrc><SimpleModelData><name>foo</name><age>123</age></SimpleModelData><PossibleColors><string>red</string><string>green</string><string>blue</string></PossibleColors></ModelWithEncodedNamesSrc>'
```

packages/cadl-ranch-specs/http/azure/core/basic/mockapi.ts
sarangan12: This was due to the issue in the generated handler (`packages/cadl-ranch/src/app/app.ts`). Now, I have added a fix to the code and confirmed that it is working. Please refer the snapshot below:

image

I have updated this PR with the required code.

Note to self: Take that code change to `typespec` repository

packages/cadl-ranch-specs/http/azure/core/traits/mockapi.ts
sarangan12: Added the handler back.

packages/cadl-ranch-specs/http/azure/resource-manager/models/common-types/managed-identity/mockapi.ts
sarangan12: Can you please elaborate this comment? There are 3 Scenarios in this file. All three scenarios, the handlers are almost identical with or without my change. What comparison is causing issue in the Java tests?

packages/cadl-ranch-specs/http/payload/json-merge-patch/mockapi.ts
sarangan12: Omitted `name: "InnerMadge"` from json-merge-patch.

packages/cadl-ranch-specs/http/encode/bytes/mockapi.ts
```diff
@@ -68,13 +68,6 @@ function createPropertyServerTests(uri: string, data: any, value: any) {
     response: {
       status: 200,
     },
-    handler: (req: MockRequest) => {
```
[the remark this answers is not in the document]
sarangan12: Added it.

packages/cadl-ranch-specs/http/payload/xml/mockapi.ts
sarangan12: @tadelesh This handler removal will cause issue. But, I do not think it is because of the case sensitivity. The previous handler was using `data` for comparison while the generated handler uses the `xml(data)` for comparison which causes the issue. I have added a fix for that and confirmed this is working fine now.

packages/cadl-ranch-specs/http/azure/resource-manager/models/common-types/managed-identity/mockapi.ts
sarangan12: I have modified the code to check for the presence `identity` instead of validating the entire body. Resolving this for now.

weidongxu-microsoft (approved):
LGTM

packages/cadl-ranch-specs/http/payload/xml/mockapi.ts
```diff
@@ -154,13 +148,6 @@ function createServerTests(uri: string, data?: any) {
       response: {
         status: 204,
       },
-      handler: (req: MockRequest) => {
-        req.expect.containsHeader("content-type", "application/xml");
-        req.expect.xmlBodyEquals(data);
```
tadelesh, line 159: still get error for xml response comparison. just as my last comment said, the difference seems from the `<xml>` header and format.
```
warn: Request validation failed: Body provided doesn't match expected body:
 Expected:
'\n' +
  '<ModelWithDictionary>\n' +
  '  <metadata>\n' +
  '    <Color>blue</Color>\n' +
  '    <Count>123</Count>\n' +
  '    <Enabled>false</Enabled>\n' +
  '  </metadata>\n' +
  '</ModelWithDictionary>\n'
 Actual:
"<?xml version='1.0' encoding='utf-8'?>\n" +
  '<ModelWithDictionary><metadata><Color>blue</Color><Count>123</Count><Enabled>false</Enabled></metadata></ModelWithDictionary>'
info: PUT /payload/xml/modelWithDictionary 400 3.334 ms - 403
```

sarangan12: I have put the handler back. This should resolve the issue.

tadelesh (approved):
It's better to update some of tests scenario docs as this [comment](https://github.com/Azure/cadl-ranch/pull/755?notification_referrer_id=NT_kwDOABpX5rMxMjkwMzQzNzQ2MzoxNzI2NDM4&notifications_query=reason%3Aparticipating#discussion_r1804333923) mentioned. Previous, many tests allow request param as any value, and return that value as response. After this PR, they are all fixed.

packages/cadl-ranch-specs/http/azure/client-generator-core/access/mockapi.ts
```diff
@@ -16,15 +16,6 @@ function createMockApiDefinitions(route: string): MockApiDefinition {
       status: 200,
       body: json({ name: "sample" }),
     },
-    handler: (req: MockRequest) => {
```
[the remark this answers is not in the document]
sarangan12: Done

Issue #686, opened 2024-08-13
ARM, test case for location-based resource


XiaofeiCao:
doc: https://azure.github.io/typespec-azure/docs/howtos/ARM/resource-type#location-based-resource

example: https://github.com/Azure/typespec-azure/blob/main/packages/samples/specs/resource-manager/resource-types/location/main.tsp#L29-L32

under resource-manager/models/resources, add `location.tsp`.

PR #762, opened 2024-10-28
Reopen Test ARM, test case for location-based resource

v-hongli1:
- https://github.com/Azure/cadl-ranch/issues/686


packages/cadl-ranch-specs/http/azure/resource-manager/resources/location.tsp
```diff
@@ -0,0 +1,172 @@
...
+using Azure.Core;
+using Azure.ResourceManager;
+using TypeSpec.OpenAPI;
+
+namespace Azure.ResourceManager.Resources;
+
+@locationResource
+model LocationResource is ProxyResource<LocationResourceProperties> {
+  ...ResourceNameParameter<LocationResource>;
+}
```
XiaofeiCao, line 17: I think we should test `ArmLocationResource` instead of `@locationResource`.

As Mark mentioned:
image
Difference between these two are generated paths:
- For `@locationResource`, path be `/subscriptions/{subscriptionId}/locations/{location}/providers/Microsoft.xxxxx`
- For `ArmLocationResource`, path be `/subscriptions/{subscriptionId}/providers/Microsoft.xxxxx/locations/{location}`

I've only seen path 2 in real service. Also, I'm not seeing any real use case of `@locationResource` in specs repo.
@tadelesh for comment as well.

v-hongli1: The `ArmLocationResource` can only specify one type of resource from 3 kinds `Tenant, Subscription, ResourceGroup`, so in the fixed version, `ResourceGroup` is specified as the `BaseType` of `ArmLocationResource`.

XiaofeiCao: Use [`ResourceGroupLocationResource`](https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-resource-manager/lib/parameters.tsp#L64)?

packages/cadl-ranch-specs/http/azure/resource-manager/resources/location.tsp
```diff
@@ -0,0 +1,173 @@
...
+using TypeSpec.Http;
+using TypeSpec.Rest;
+using Azure.Core;
+using Azure.ResourceManager;
+using TypeSpec.OpenAPI;
+
+namespace Azure.ResourceManager.Resources;
+
+@resource("locationResources")
+@parentResource(ArmLocationResource<"ResourceGroup">)
```
XiaofeiCao, line 15: Use `ResourceGroupLocationResource`.

lirenhe:
I don't think we need to add more test cases into the repo of cadl-ranch now.  As this is an Azure case, could you move it to typespec-azure repo?

weidongxu-microsoft:
it should be already in typespec-azure https://github.com/Azure/typespec-azure/pull/1932

@v-hongli1 Can you close those PR that we no longer need?

PR #769, opened 2024-11-11
fix(mockapi): dotnet compatibility failure

archerzz:
dotnet uses different but compatible datetime format, so the comparison should use `coercedBodyEquals`

### DIFF
--- a/packages/cadl-ranch-specs/http/authentication/api-key/mockapi.ts
+++ b/packages/cadl-ranch-specs/http/authentication/api-key/mockapi.ts
@@ -1,4 +1,4 @@
-import { json, passOnCode, passOnSuccess, MockRequest } from "@azure-tools/cadl-ranch-api";
+import { json, passOnCode, passOnSuccess } from "@azure-tools/cadl-ranch-api";
 import { ScenarioMockApi } from "@azure-tools/cadl-ranch-api";
 
 export const Scenarios: Record<string, ScenarioMockApi> = {};
@@ -31,14 +31,6 @@ Scenarios.Authentication_ApiKey_invalid = passOnCode(403, {
     body: json({
       error: "invalid-api-key",
     }),
-  },
-  handler: (req: MockRequest) => {
-    return {
-      status: 403,
-      body: json({
-        error: "invalid-api-key",
-      }),
-    };
   },
   kind: "MockApiDefinition",
 });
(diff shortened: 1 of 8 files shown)


packages/cadl-ranch-specs/http/authentication/api-key/mockapi.ts
```diff
@@ -32,5 +32,13 @@ Scenarios.Authentication_ApiKey_invalid = passOnCode(403, {
       error: "invalid-api-key",
     }),
   },
+  handler: (req: MockRequest) => {
```
archerzz, line 35: This is just a revert of https://github.com/Azure/cadl-ranch/commit/caa22906ae9f9dc3031384b7a6b90857a52fbd15#diff-7e6ef54edf1535ebe6cbfdf2a7ea48d31f33f797ce5135ec7cb55a616f4f7f1bL39

However, it seems that we don't need this handler. But if I didn't add this handler there will be error saying that the returned status code is `400`:
image

packages/cadl-ranch-specs/http/authentication/http/custom/mockapi.ts
```diff
@@ -32,5 +32,13 @@ Scenarios.Authentication_Http_Custom_invalid = passOnCode(403, {
       error: "invalid-api-key",
     }),
   },
+  handler: (req: MockRequest) => {
```
archerzz, line 35: same as above

packages/cadl-ranch-specs/http/type/dictionary/mockapi.ts
```diff
@@ -23,6 +23,12 @@ function createServerTests(uri: string, data: any) {
       response: {
         status: 204,
       },
+      handler: (req) => {
+        req.expect.coercedBodyEquals(data);
```
archerzz, line 27: all the remainings are due to dotnet uses different but compatible datetime format. So, we need to use `coercedBodyEquals` instaed of exact match.

packages/cadl-ranch-specs/http/authentication/api-key/mockapi.ts
weidongxu-microsoft: For me, the response is 403

```
[ForkJoinPool-1-worker-1] INFO com.authentication.apikey.implementation.ApiKeyClientImpl$ApiKeyClientService.invalidSync - {"az.sdk.message":"HTTP response","statusCode":403,"url":"http://localhost:3000/authentication/api-key/invalid","durationMs":4,"content-length":27,"Date":"Mon, 11 Nov 2024 07:08:30 GMT","Connection":"keep-alive","ETag":"W/\"1b-VNTCj8CGjwiSZNVCTNW/yOEvenc\"","Content-Type":"application/json; charset=utf-8","redactedHeaders":"Keep-Alive,X-Powered-By","content-length":27,"body":"{\"error\":\"invalid-api-key\"}"}
```

Not sure what happens to .NET

archerzz: I found the root cause. The dotnet test case is wrong in the input. We don't need this change.

Issue #758, opened 2024-10-23
Namespace refactoring Scenario tests


lmazuel:

weidongxu-microsoft:
PR at initial review to get feedback from all languages.
https://github.com/Azure/cadl-ranch/pull/775

PR #775, opened 2024-11-20
specs, clientNamespace

weidongxu-microsoft:
fix https://github.com/Azure/cadl-ranch/issues/758

This scenario mainly test the case when `@clientNamespace` applied to TypeSpec Namespace/Interface/Model/Enum.

So far it does not test corner cases, like client/subclient in different namespace, name conflict for model in different namespace, etc.

### DIFF
--- a/packages/cadl-ranch-specs/http/azure/client-generator-core/client-namespace/client.tsp
+++ b/packages/cadl-ranch-specs/http/azure/client-generator-core/client-namespace/client.tsp
@@ -26,7 +26,7 @@ namespace Client;
 interface ClientNamespaceFirstClient {
   get is First.get;
 }
-@@clientNamespace(FirstModel, "azure.clientgenerator.core.clientnamespace.firstt", "java");
+@@clientNamespace(FirstModel, "azure.clientgenerator.core.clientnamespace.first", "java");
 
 @client({
   name: "ClientNamespaceSecondClient",
--- a/packages/cadl-ranch-specs/http/azure/client-generator-core/client-namespace/client.tsp
+++ b/packages/cadl-ranch-specs/http/azure/client-generator-core/client-namespace/client.tsp
@@ -1,0 +1,40 @@
+import "./main.tsp";
+import "@azure-tools/typespec-client-generator-core";
+
+using TypeSpec.Http;
+using global.Azure.ClientGenerator.Core;
+using _Specs_.Azure.ClientGenerator.Core.ClientNamespace;
+
+@scenario
+@scenarioDoc("""
+  Expected client namespace for clients:
+  - ClientNamespaceFirstClient: Azure.ClientGenerator.Core.ClientNamespace
+  - ClientNamespaceSecondClient: Azure.ClientGenerator.Core.ClientNamespace.Second
+  
+  Expected client namespace for models:
+  - FirstClientResult: Azure.ClientGenerator.Core.ClientNamespace.First
+  - SecondClientResult: Azure.ClientGenerator.Core.ClientNamespace.Second
+  - SecondClientEnumType: Azure.ClientGenerator.Core.ClientNamespace.Second.Sub
+  """)
+namespace Client;
+
+@client({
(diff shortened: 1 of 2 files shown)


packages/cadl-ranch-specs/http/azure/client-generator-core/client-namespace/client.tsp
```diff
@@ -0,0 +1,40 @@
...
+  - SecondClientResult: Azure.ClientGenerator.Core.ClientNamespace.Second
+  - SecondClientEnumType: Azure.ClientGenerator.Core.ClientNamespace.Second.Sub
+  """)
+namespace Client;
+
+@client({
+  name: "ClientNamespaceFirstClient",
+  service: _Specs_.Azure.ClientGenerator.Core.ClientNamespace,
+})
+@clientNamespace("azure.clientgenerator.core.clientnamespace", "java")
```
weidongxu-microsoft, line 25: I assume for other languages:

csharp: `Azure.ClientGenerator.Core.ClientNamespace`
python: `azure.client_generator.core.client_namespace`

Let me know what the pattern for Go/JS. I will fill in here before PR public for review.

packages/cadl-ranch-specs/http/azure/client-generator-core/client-namespace/client.tsp
```diff
@@ -0,0 +1,40 @@
...
+interface ClientNamespaceFirstClient {
+  get is First.get;
+}
+@@clientNamespace(FirstModel, "azure.clientgenerator.core.clientnamespace.firstt", "java");
```
msyyc, line 29:
```suggestion
@@clientNamespace(FirstModel, "azure.clientgenerator.core.clientnamespace.first", "java");
```

packages/cadl-ranch-specs/http/azure/client-generator-core/client-namespace/main.tsp
```diff
@@ -0,0 +1,39 @@
...
+}
+
+namespace Second {
+  #suppress "@azure-tools/cadl-ranch-expect/missing-scenario" "scenario defined in client.tsp"
+  @route("/second")
+  @get
+  op get(): Model.SecondClientResult;
+
+  namespace Model {
+    model SecondClientResult {
```
msyyc, line 30: I think it is a little weird to put model into another namespace different with client on purpose. For python SDK, we will always put model into folder `models`, then current cadl-ranch will results in namespace `xxx.model.models` in python.

weidongxu-microsoft: Java is same logic, it would be in `model.models` in azure flavor.

However, there is nothing prohibit unbranded from doing this?

weidongxu-microsoft: PS: in client.tsp, I actually map this to same client namespace of the client.
On the other hand, I actually map the `SecondClientEnumType` to a sub namespace.

```ts
@@clientNamespace(Second.Model, "azure.clientgenerator.core.clientnamespace.second", "java");
@@clientNamespace(Second.Model.SecondClientEnumType, "azure.clientgenerator.core.clientnamespace.second.sub", "java");
```

weidongxu-microsoft: @msyyc 

The possible scenario of model/client in difference space (described by Srikanta) is that there be 2 catalogs of models
1. model common to multiple clients
2. model specific on a single client

If we put clients to different namespaces, there will certainly be for "common model", some of them would have different namespace from the client that uses them.

weidongxu-microsoft: And even without this test scenario, lots of existing tests defined model in a different namespace (e.g. https://github.com/microsoft/typespec/pull/5051#discussion_r1845911525)

